Issue #2, opened 2022-09-07
Initial parsing of constraints against the `clk` column


grjte:
To start the project very simply, we should define the constraints and then the grammar to parse the following constraints against Miden VM's clock cycle column using LALRPOP

- One boundary constraint that clk at the first step is 1.
- One transition constraint which says that clk' = clk + 1.

grjte:
Closed by #4

PR #4, opened 2022-09-08
Parsing simple constraints

tohrnii:
Addressing #2.


parser/src/token.rs
```diff
@@ -0,0 +1,86 @@
...
+            Self::Lpar => write!(f, "("),
+            Self::Rpar => write!(f, ")"),
+            Self::Lsqb => write!(f, "["),
+            Self::Rsqb => write!(f, "]"),
+            Self::Star => write!(f, "*"),
+            Self::Minus => write!(f, "-"),
+            Self::MinusEqual => write!(f, "-="),
+            Self::Plus => write!(f, "+"),
+            Self::PlusEqual => write!(f, "+="),
+            Self::Rarrow => write!(f, "->"),
```
bobbinth, line 83: This is just a single comment (I haven't really looked at other parts much): We should probably start with the smallest set of keywords we need for now, and then expand this in future PRs. So, things like `import`, `for`, `in` etc. should probably be removed from here for now.

tohrnii: I've removed all keywords that we don't use in the current test file.

grjte:
Thanks @tohrnii looks like a great start! I'll review in more depth soon. In the meantime, can you add some comments & a brief README?

parser/src/tests/input/testair.airdsl
```diff
@@ -0,0 +1,10 @@
+def TestAir
```
grjte, line 1: Maybe we should just use `.air` as the exension for now to keep it shorter and simpler? Let's also rename this to either `clock` or `system`, since we'll be adding more tests later. So maybe `Def SystemAir` and `system.air`? (or something like this)

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,144 @@
...
+}
+
+SourceSection: SourceSection = {
+    AirDef => SourceSection::AirDef(<>),
+    BoundaryConstraints => SourceSection::BoundaryConstraints(<>),
+    TraceCols => SourceSection::TraceCols(<>),
+    TransitionConstraints => SourceSection::TransitionConstraints(<>),
+}
+
+// FILE DEF
```
grjte, line 27: nit: I would probably call this "AIR DEF"

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,144 @@
...
+        TraceColsGrpType, TransitionConstraints
+    }, error::{Error, ParseError::InvalidInt}, lexer::Tok
+};
+use std::str::FromStr;
+use lalrpop_util::ParseError;
+
+grammar;
+
+// AST NODE
+// ------------------------------------------------------------------------------------------------
```
grjte, line 14: nit: I would use the "======" for top-level separators as we do in the Miden repo

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,144 @@
...
+
+BoundaryConstraints: BoundaryConstraints = {
+    "boundary_constraints" ":" <boundary_constraints: Constraint+> =>
+        BoundaryConstraints { boundary_constraints }
+}
+
+Boundary: Boundary = {
+    "first" => Boundary::First,
+    "last" => Boundary::Last
+}
```
grjte, line 67: nit: I would put the boundary constraints section before the transition constraints section. It's not a big deal, but it's just useful to be consistent across all places, and that seems to be the logical order we follow elsewhere in Miden.

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,144 @@
...
+        "first" => Tok::First,
+        "last" => Tok::Last,
+        ":" => Tok::Colon,
+        "," => Tok::Comma,
+        "=" => Tok::Equal,
+        "[" => Tok::Lsqb,
+        "]" => Tok::Rsqb,
+        "." => Tok::Dot,
+        "+" => Tok::Plus,
+    }
```
grjte, line 143: nit: I would order as suggested elsewhere, or at least aim to have our order here be consistent with the order in the lexer and the order above to the extent that it's possible

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,144 @@
...
+            error: Error::ParseError(InvalidInt(format!("Int too big : {}", <>).to_string()))
+        })
+}
+
+// LEXER
+// ------------------------------------------------------------------------------------------------
+
+extern {
+    type Error = Error;
+    type Location = usize;
```
grjte, line 123: Do we still need this Location?

parser/src/ast.rs
```diff
@@ -0,0 +1,100 @@
...
+
+/// Source is divided into SourceSections.
+/// There are 4 types of Source Sections:
+/// - AirDef: Name of the air constraints module.
+/// - TraceCols: Trace Columns representing columns of the execution trace.
+/// - TransitionConstraints: Transition Constraints to be enforced on the trace columns defined
+///   in the TraceCols section.
+/// - BoundaryConstraints: Boundary Constraints to be enforced on the boundaries of columns defined
+///   in the TraceCols section. Currently there are two types of boundaries, First and Last
+///   representing the first and last rows of the column.
```
grjte, line 28: nit: I would swap the order of Boundary and Transition constraints here and in the enum below and in the struct delarations further down

parser/src/tests/errors.rs
```diff
@@ -0,0 +1,72 @@
...
+#[test]
+fn error_invalid_token() {
+    let source = "
+    transition_constraints:
+        enf clk'' = clk + 1";
+    // Only one "'" is allowed at the end of identifier to indicate next row of the column.
+    // Should fail if more than one "'" is present in an identifier
+    let error = Error::ScanError(45..46);
+    build_parse_test!(source).expect_error(error);
+}
```
grjte, line 16: We should be able to simplify this so that the source only contains the expression under test. Since scan errors should appear before parse errors, we should be able to get the `ScanError` without including the section keyword.

I would make this change for all other tests as well - make the source string as small as possible while still including the error

parser/src/tests/errors.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[test]
+fn error_identifier_with_invalid_characters() {
+    let source = "
+    transition_constraints:
+        enf clk@' = clk@ + 1";
```
grjte, line 22: I would only use the `@` once. Otherwise once we start handling multiple errors this will need to be changed. I'd rather have this have a single error and then when we want to test multiple errors later we can add new tests for that.

If you want, you could instead add one new test with a source string that contains 2 errors to show that it fails at the first `ScanError` and only returns one error. Then when we add handling of multiple errors we'll expect that test to fail & we'll replace/update it

parser/src/tests/errors.rs
```diff
@@ -0,0 +1,72 @@
...
+    let source = "enf clk' = clk + 1";
+    build_parse_test!(source).expect_unrecognized_token();
+}
+
+#[test]
+fn error_identifier_starting_with_int() {
+    // Identifiers cannot start with numeric characters.
+    let source = "
+    transition_constraints:
+        enf 1clk' = 1clk + 1";
```
grjte, line 70: shouldn't this be a `ScanError` rather than failing at the parser level?

grjte (changes requested):
Thanks @tohrnii , this is looking great! I left a bunch of minor things inline, but overall think it's nicely structured and feels very readable and extensible now.

One last larger change I would make is to change the way the test module is structured. Instead of having a test module for errors and a separate one for correct ASTs, I would split it up into the following:

- `tests/scanner.rs`or `tests/lexer.rs`: test correct tokenization and ScanError in one module
- `tests/parser.rs`: test ParseError and correct AST output in another module. Then as this grows we can easily split this module into submodules for the various sections, etc.

My reasoning is that I think it's easier to see tests for validity and errors side by side so we can identify more easily if we've covered all cases.

parser/src/lexer.rs
```diff
@@ -0,0 +1,91 @@
...
+    Dot,
+
+    #[token("+")]
+    Plus,
+
+    #[error]
+    // skip whitespaces, tabs and newlines
+    #[regex(r"[ \t\n\f]+", logos::skip)]
+    // skip comments
+    #[regex(r"#.*\n?", logos::skip)]
```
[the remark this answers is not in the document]
grjte: Although actually it's fine to leave this for a future PR

grjte:
One more thing - it woule

air-dsl/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "air-dsl"
+version = "0.1.0"
+edition = "2021"
```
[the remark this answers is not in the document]
tohrnii: I think rust 2021 is the latest edition. I've changed the version to 0.0.1.

parser/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "parser"
+version = "0.1.0"
+edition = "2021"
+
+[build-dependencies]
+lalrpop = "0.19.7"
+
+[dependencies]
+lalrpop-util = { version = "0.19.7", features = ["lexer"] }
```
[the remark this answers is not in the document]
tohrnii: Yeah we don't need it anymore. Removed.

parser/src/lexer.rs
```diff
@@ -0,0 +1,91 @@
...
+    #[token("enf")]
+    Enf,
+
+    // used to represent the first row of the column to which a boundary constraint is applied.
+    #[token("first")]
+    First,
+
+    // used to represent the last row of the column to which a boundary constraint is applied.
+    #[token("last")]
+    Last,
```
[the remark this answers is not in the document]
tohrnii: Thanks. Done.

parser/src/grammar.lalrpop
tohrnii: Lalrpop requires tokens in the form `(location_start, token, location_end)` so we need to expose logos' location type to lalrpop.

parser/src/lexer.rs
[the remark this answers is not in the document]
tohrnii: Agreed, I was thinking to handle it in a future PR as well so I've kept it as is for now.

parser/src/tests/errors.rs
tohrnii: Sorry, I should have made it clearer. Since logos has recoverable errors so it doesn't fail while scanning. Since we don't have recoverable errors in lalrpop yet, the parser returns the first error it encounters so if we remove "transition_constraints" in the above, it will return an error there since there's no match for source section keyword.
I've changed the `expect_error` function check for the first failing error while scanning and match against that and hence changed the test to what you suggested to make the tests more atomic.
This should be improved when we have recoverable errors at parser level as well in the future.

parser/src/tests/errors.rs
tohrnii: Makes sense. Done.

parser/src/tests/errors.rs
tohrnii: Thanks for the catch. Logos tokenizes this into `Number(1)` and `Ident(clk)` so this doesn't fail at the lexer level. I've added an explanation for this test and added a few other tests to cover some potential confusing cases.

tohrnii:

air-dsl/Cargo.toml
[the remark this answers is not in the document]
grjte: Sorry, good point, thanks!

parser/src/tests/errors.rs
grjte: Ok we can revisit it then, thanks for explaining!

parser/src/tests/errors.rs
grjte: Great, thank you!

parser/src/ast.rs
```diff
@@ -0,0 +1,101 @@
...
+pub struct AirDef {
+    pub name: Identifier,
+}
+
+/// [TraceCols] contains the trace columns representing columns of the execution trace.
+/// TraceCols is made up of TraceColsGrp which represents groups of columns of a specific type.
+#[derive(Debug, PartialEq)]
+pub struct TraceCols {
+    pub cols: Vec<TraceColsGrp>,
+}
```
bobbinth, line 48: Since we have only two possible types of trace columns (`main` and `aux`), could this be something like:
```Rust
pub struct TraceCols {
    pub main_cols: Vec<Expr>,
    pub aux_cols: Vec<Expr>,
}
```
Or does this complicate things?

parser/src/ast.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// [TraceColsGrp] represents a group of columns of a specfic type.
+/// There are two types of groups:
+/// - MainTraceCols
+/// - AuxiliaryTraceCols
+#[derive(Debug, PartialEq)]
+pub struct TraceColsGrp {
+    pub cols_grp_type: TraceColsGrpType,
+    pub cols: Vec<Expr>,
```
bobbinth, line 57: Does this imply that `cols` could contain any expressions? If so, should we limit it as there is only very restricted set of options which could be used here? Specifically, for this PR we could restrict this only to a list of identifiers. For example:
```
trace_columns:
    main: [a, b, c, d]
```
In the future PRs, we could add more complex syntax. For example:
```
trace_columns:
    main: [a, b[2], c[3], d, ..5]
```
Or something like that - but we should probably discuss first what syntax we want to allow.

parser/src/ast.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Stores the transition constraints to be enforced to the trace column values.
+#[derive(Debug, PartialEq)]
+pub struct TransitionConstraints {
+    pub transition_constraints: Vec<Constraint>,
+}
+
+/// Stores the expression corresponding to the transition or boundary constraint.
+#[derive(Debug, PartialEq)]
+pub struct Constraint {
+    pub expr: Expr,
```
bobbinth, line 89: I think this might be too permissive. A constraint must always contain an equality expression. For example: `a = b`. Should this be defined as:
```
pub struct Constraint {
    pub lhs: Expr,
    pub rhs: Expr,
}
```

Also, should there be different types for transition vs. boundary constraints? For example, `enf a.first = b` is a valid boundary constraint, but is not a valid transition constraint. Or will we do these types of checks later on?

bobbinth (changes requested):
Looks great! Thank you! I left a few comments and questions inline.

parser/src/ast.rs
tohrnii: Makes sense. This seems a better way to handle it. I've changed it to what you suggested where an error is thrown if neither main nor auxiliary columns are declared.

parser/src/ast.rs
tohrnii: Makes sense. Since this has been changed to what you suggested above now, main and auxiliary columns can only contain identifiers for now.

parser/src/ast.rs
tohrnii: I've changed the constraint definition to what you suggested to make it more restrictive.

> Also, should there be different types for transition vs. boundary constraints? For example, enf a.first = b is a valid boundary constraint, but is not a valid transition constraint. Or will we do these types of checks later on?

I think it might be better to handle it after the AST is built otherwise it seems like the grammar would become much more complex (unless I'm missing something?) and it will be easier to do these checks once the AST is built. I'll think on this more though.

bobbinth: > I think it might be better to handle it after the AST is built otherwise it seems like the grammar would become much more complex

Agreed! Let's do these checks later.

parser/src/ast.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub lhs: Expr,
+    pub rhs: Expr,
+}
+
+/// Arithmetic expressions representing transition or boundary constraint.
+#[derive(Debug, PartialEq)]
+pub enum Expr {
+    Add(Box<Expr>, Box<Expr>),
+    Subtract(Box<Expr>, Box<Expr>),
+    Equal(Box<Expr>, Box<Expr>),
```
bobbinth, line 82: Do we need `Equal` to be a part of the expression? I can think of only a couple of places where we'd use it:

1. Constraint expression - which should already be covered.
2. Variable assignment - which we'll probably cover in a similar manner as the constraint expression.

parser/src/ast.rs
bobbinth: Actually, thinking about it some more, I wonder if it is actually not too complicated. Here is how it might work:

We could change the current `Constraint` to `TransitionConstraint` and it would look like so:
```Rust
pub struct TransitionConstraint {
    pub lhs: Expr,
    pub rhs: Expr,
}
```

We could add a new struct for boundary constraint which could look something like:
```Rust
pub struct BoundaryConstraint {
    pub column: Identifier,
    pub boundary: Boundary,
    pub value: Expr,
}
```

Then, `TransitionConstraints` and `BoundaryConstraints` would look like so:
```Rust
pub struct TransitionConstraints {
    pub transition_constraints: Vec<TransitionConstraint>,
}

pub struct BoundaryConstraints {
    pub boundary_constraints: Vec<BoundaryConstraints>,
}
```
And finally, we would be able to remove `Boundary` variant from the `Expr` enum.

Do you think this would complicate things too much?

tohrnii: Yeah, sorry, I was mistaken, this will work. I have changed it to what you suggested.

parser/src/ast.rs
tohrnii: I was keeping it for variable assignment but makes sense to handle it the same way as constraint expression, so I've removed it. Thanks.

bobbinth (approved):
Thank you! All looks good to me.

parser/src/ast.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/// Name of the air constraints module.
+#[derive(Debug, Eq, PartialEq)]
+pub struct AirDef {
+    pub name: Identifier,
+}
+
+/// [TraceCols] contains the trace columns representing columns of the execution trace.
+/// TraceCols is made up of TraceColsGrp which represents groups of columns of a specific type.
```
grjte, line 44: It looks like `TraceColsGrp` has been removed, right? In that case, this comment should be updated.

grjte (changes requested):
Thanks @tohrnii it looks great! I left 3 very small final things inline and then I think it's ready to merge!

parser/src/ast.rs
tohrnii: Sorry, fixed.

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,154 @@
...
+Expr2: Expr = {
+    <lexpr: Expr1> "+" <rexpr: Expr1> => Expr::Add(Box::new(lexpr), Box::new(rexpr)),
+    <lexpr: Expr1> "-" <rexpr: Expr1> => Expr::Subtract(Box::new(lexpr), Box::new(rexpr)),
+    Expr1
+}
+
+Expr1: Expr = {
+    <n: Num> => Expr::Int(Token::Number(n.to_string())),
+    <s: Identifier> => Expr::Variable(s),
+}
```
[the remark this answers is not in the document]
tohrnii: Do you think it makes sense to keep the `Expr` suffix in names to make it clear or would it be better to keep it concise?
I've made the following changes - 
`Expr` -> `Expr`
`Expr2` -> `ArithmeticExpr`
`Expr1` -> `AtomExpr`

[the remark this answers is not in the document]
grjte: I think this is fine - it gives the clarity I was looking for. We can always shrink them later if we want

grjte (approved):
Looks good, thank you!

PR #17, opened 2022-10-20
Parser crate refactor

grjte:
This PR refactors the parser crate so to keep things clean as we add the next functionality and tests.

- moves lexer, lalrpop parser/grammar, and ast each to a dedicated module
- moves tests into the respective modules
- adds a public `parse` function for use outside the crate and refactors `ir` and CLI to use it, removing lexer visibility
- refactors parser tests into modules by AST groupings so we can have more visibility on test coverage and add new tests more easily
- updates the README to reflect the new parser interface

A few notes on testing:
- To the 4th point above: we'll want to do something similar with the lexer
- we may want to consider something like [expect_test](https://docs.rs/expect-test/latest/expect_test/) for full file parsing tests, but this should be revisited later


parser/src/lexer/tests.rs
```diff
@@ -1,11 +1,30 @@
...
+// TEST HELPERS
+// ================================================================================================
+
+fn expect_valid_tokenization(source: &str, expected_tokens: Vec<Token>) {
+    let tokens: Vec<Token> = Lexer::new(source).collect();
+    assert_eq!(tokens, expected_tokens);
+}
+
+/// Checks that source is valid and asserts that appropriate error is returned if there
+/// is a problem while scanning or parsing the source.
```
tohrnii, line 15: Should this be "while scanning the source."?

parser/src/parser/tests/utils.rs
```diff
@@ -49,38 +46,24 @@ impl ParseTest {
     // TEST METHODS
     // --------------------------------------------------------------------------------------------
 
-    pub fn expect_valid_tokenization(&self, expected_tokens: Vec<Token>) {
-        let tokens: Vec<Token> = Lexer::new(self.source.as_str()).collect();
-        assert_eq!(tokens, expected_tokens);
-    }
-
     /// Checks that source is valid and asserts that appropriate error is returned if there
     /// is a problem while scanning or parsing the source.
```
tohrnii, line 50: Should this be "while parsing the source."?

parser/src/lib.rs
```diff
@@ -1,11 +1,19 @@
...
+use error::Error;
+
+mod lexer;
+use lexer::{Lexer, Token};
+
+mod parser;
+use crate::parser::SourceParser;
+
+/// Parses the provided source and returns the AST.
+pub fn parse(source: &str) -> Result<ast::Source, lalrpop_util::ParseError<usize, Token, Error>> {
```
tohrnii, line 16: nit: Should we import module items instead of using full paths for better readability?

tohrnii (approved):
Thank you @grjte, looks great! I've left a couple nits about comments.

parser/src/lib.rs
tohrnii: No that makes sense. Thanks for the clarification.

Issue #12, opened 2022-10-19
Add support for multiplication


grjte:
Multiplication needs to be added to the entire pipeline

- [ ] scanning
- [ ] parsing
- [ ] IR
- [ ] codgen

PR #21, opened 2022-10-23
Add support for multiplication

tohrnii:
Addressing #12.
Adds multiplication to entire pipeline.

- [x] scanning
- [x] parsing
- [x] IR
- [x] codgen (doesn't include unit test for codegen)


grjte (changes requested):
Thanks @tohrnii, it looks good! My comments are mostly related to doing more rigorous testing - I think this is a good time for us to start addressing edge cases more systematically.

One other general nit: when a `clippy` or `rustfmt` error is introduce in a particular PR, I would prefer to rebase the fix. A "fix" commit is fine if the error was introduced in a different PR, but in a case like this it's a simple rebase, and it will keep the git history cleaner.

parser/src/lexer/tests.rs
```diff
@@ -45,11 +45,16 @@ fn keywords_with_identifiers() {
 
 #[test]
 fn multi_arithmetic_ops() {
-    let source = "enf clk' - clk - 1 = 0";
+    let source = "enf clk' * clk - clk' - clk - 1 = 0";
```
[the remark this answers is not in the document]
tohrnii: I've renamed these to `chained_<op>_name` but I'm not sure if there's much utility in doing extensive testing for lexer for chained operations since these should only cause error at the parser level. Maybe all of lexer tests should be structured differently than I've done in the updated PR. Please let me know your thoughts on this.

parser/src/parser/grammar.lalrpop
```diff
@@ -87,8 +87,13 @@ TransitionConstraint: TransitionConstraint = {
...
 Expr: Expr = {
-    <lexpr: Expr> "+" <rexpr: NextExpr> => Expr::Add(Box::new(lexpr), Box::new(rexpr)),
-    <lexpr: Expr> "-" <rexpr: NextExpr> => Expr::Subtract(Box::new(lexpr), Box::new(rexpr)),
+    <lexpr: Expr> "+" <rexpr: MulExpr> => Expr::Add(Box::new(lexpr), Box::new(rexpr)),
+    <lexpr: Expr> "-" <rexpr: MulExpr> => Expr::Subtract(Box::new(lexpr), Box::new(rexpr)),
+    MulExpr
+}
+
+MulExpr: Expr = {
+    <lexpr: NextExpr> "*" <rexpr: NextExpr> => Expr::Multiply(Box::new(lexpr), Box::new(rexpr)),
```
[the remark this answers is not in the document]
tohrnii: Thank you. Fixed.

parser/src/parser/tests/expressions.rs
```diff
@@ -5,6 +5,30 @@ use super::{
 
 // SECTIONS
 // ================================================================================================
+#[test]
+fn multiplication() {
+    // the operation must be put into a source section, or parsing will fail
+    let source = "
+    transition_constraints:
+        enf clk' * clk - 1 = 0";
```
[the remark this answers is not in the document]
tohrnii: Please let me know if it's ok now.

codegen-winter/src/air/transition_constraints.rs
```diff
@@ -51,43 +51,38 @@ impl Codegen for NodeIndex {
...
             Operation::Add(l_idx, r_idx) => {
                 let lhs = l_idx.to_string(graph);
-                strings.push(lhs);
 
                 // output Add followed by Neg as "-"
-                let r_idx = if let Operation::Neg(n_idx) = graph.node(r_idx).op() {
-                    strings.push(String::from("-"));
-                    n_idx
+                let rhs = if let Operation::Neg(n_idx) = graph.node(r_idx).op() {
+                    format!("-{}", n_idx.to_string(graph))
```
grjte, line 71: We actually need to put parentheses around the expression being negated. There was an error here before.

The most basic fix would be this:
```
format!("- ({})", n_idx.to_string(graph))
```

It would be nice to do something more sophisticated and only put the parenthesis if expression represented by `n_idx` is actually an operation. In that case, this part should be pulled into its own helper method

codegen-winter/src/air/transition_constraints.rs
```diff
@@ -51,43 +51,38 @@ impl Codegen for NodeIndex {
...
                 };
-
+                format!("{}{}", lhs, rhs)
+            }
+            Operation::Mul(l_idx, r_idx) => {
+                let lhs = l_idx.to_string(graph);
                 let rhs = r_idx.to_string(graph);
-
-                strings.push(rhs);
+                format!("{}*{}", lhs, rhs)
```
grjte, line 80: Similar comment to the one for negation above - We need to have parentheses around the sub-expressions if they are addition expressions. It would be nice to have a shared helper for this and the negation case above

parser/src/lexer/tests.rs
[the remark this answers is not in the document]
grjte: I think you're right about most of the complex tests being at the parser level. We can audit the tests later to assess which edge cases we're missing and which tests are redundant. For now, I think what you have is good, and I like the restructure.

grjte (changes requested):
Thanks @tohrnii ! The changes all look great. Unfortunately I think there's a bug that I didn't catch in the last pass (it was actually there previously). I left 2 comments inline. Once those are fixed, we can merge this. If you want, feel free to do the simple naive solution now so we can get this merged and open an issue for a more elegant solution

codegen-winter/src/air/transition_constraints.rs
grjte: It might make sense to do the more sophisticated fix in a future PR so we can merge this one.

codegen-winter/src/air/transition_constraints.rs
grjte: It might make sense to do the more sophisticated fix in a future PR so we can merge this one and just do the following for now:

```
format!("({}) * ({})", lhs, rhs)
```

codegen-winter/src/air/transition_constraints.rs
```diff
@@ -50,44 +50,40 @@ impl Codegen for NodeIndex {
...
             }
             Operation::MainTraceNextRow(col_idx) => {
-                strings.push(format!("next[{}]", col_idx));
+                format!("next[{}]", col_idx)
             }
             Operation::Neg(idx) => {
-                strings.push(String::from("-"));
                 let str = idx.to_string(graph);
-                strings.push(str);
+                format!("-{}", str)
```
grjte, line 65: Ah sorry, I missed one! We need to do the same thing with this one too.

grjte (changes requested):
Sorry @tohrnii there's one more instance of the same issue...

grjte:
@tohrnii can you also open an issue for implementing a more elegant solution to nested expressions during code generation?

grjte (approved):
It looks great now. Thanks @tohrnii!

Issue #13, opened 2022-10-19
Add support for exponentiation


grjte:
Exponentiation needs to be added to the entire pipeline, with unit tests at each stage.

- [x] scanning
- [x] parsing
- [x] IR
- [x] codgen

grjte:
Closed by #22

PR #22, opened 2022-10-24
Add support for exponentiation

tohrnii:
Addressing #13.
Adds exponentiation to entire pipeline.

- [x] scanning
- [x] parsing
- [x] IR
- [x] codgen (doesn't include unit test for codegen)

### DIFF
--- a/codegen-winter/src/air/boundary_constraints.rs
+++ b/codegen-winter/src/air/boundary_constraints.rs
@@ -55,15 +55,18 @@ trait Codegen {
 impl Codegen for Expr {
     fn to_string(&self) -> String {
         match self {
-            Self::Constant(value) => format!("Felt::new({})", value),
+            Self::Const(value) => format!("Felt::new({})", value),
             Self::Add(lhs, rhs) => {
                 format!("{} + {}", lhs.to_string(), rhs.to_string())
             }
-            Self::Subtract(lhs, rhs) => {
+            Self::Sub(lhs, rhs) => {
                 format!("{} - {}", lhs.to_string(), rhs.to_string())
             }
-            Self::Multiply(lhs, rhs) => {
+            Self::Mul(lhs, rhs) => {
                 format!("{} * {}", lhs.to_string(), rhs.to_string())
+            }
+            Self::Exp(lhs, rhs) => {
+                format!("({}).exp({})", lhs.to_string(), rhs)
             }
             _ => {
                 unimplemented!("unreachable code")
--- a/codegen-winter/src/air/transition_constraints.rs
+++ b/codegen-winter/src/air/transition_constraints.rs
@@ -53,7 +53,7 @@ impl Codegen for Operation {
     // TODO: Only add parentheses in Add and Mul if the expression is an arithmetic operation.
     fn to_string(&self, graph: &AlgebraicGraph) -> String {
         match self {
-            Operation::Constant(value) => format!("E::from({})", value),
+            Operation::Const(value) => format!("E::from({})", value),
             Operation::MainTraceCurrentRow(col_idx) => {
                 format!("current[{}]", col_idx)
             }
@@ -80,9 +80,9 @@ impl Codegen for Operation {
                 let rhs = r_idx.to_string(graph);
                 format!("({}) * ({})", lhs, rhs)
             }
-            _ => {
-                unimplemented!()
-                // TODO: Exp
+            Operation::Exp(l_idx, r_idx) => {
+                let lhs = l_idx.to_string(graph);
+                format!("({}).exp({})", lhs, r_idx)
             }
         }
     }
--- a/ir/src/transition_constraints/graph.rs
+++ b/ir/src/transition_constraints/graph.rs
@@ -35,7 +35,7 @@ impl AlgebraicGraph {
     pub fn degree(&self, index: &NodeIndex) -> u8 {
         // recursively walk the subgraph and compute the degree from the operation and child nodes
         match self.node(index).op() {
-            Operation::Constant(_)
+            Operation::Const(_)
             | Operation::MainTraceCurrentRow(_)
             | Operation::MainTraceNextRow(_) => 1_u8,
             Operation::Neg(index) => self.degree(index),
@@ -55,13 +55,13 @@ impl AlgebraicGraph {
         trace_columns: &TraceColumns,
     ) -> Result<NodeIndex, SemanticError> {
         match expr {
-            Expr::Constant(value) => Ok(self.insert_op(Operation::Constant(value))),
+            Expr::Const(value) => Ok(self.insert_op(Operation::Const(value))),
             Expr::Next(Identifier(ident)) => {
                 let index = trace_columns.get_column_index(&ident)?;
                 // insert the next row column node.
                 Ok(self.insert_op(Operation::MainTraceNextRow(index)))
             }
-            Expr::Variable(Identifier(ident)) => {
+            Expr::Var(Identifier(ident)) => {
                 // since variable definitions are not possible yet, the identifier must match one of
                 // the declared trace columns.
                 let index = trace_columns.get_column_index(&ident)?;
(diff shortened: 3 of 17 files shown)


tohrnii:
Currently the exponent can only be a constant but in examples written by @bobbinth in the original discussion also allows the exponent to be a variable ([example](https://github.com/maticnetwork/miden/discussions/254#discussioncomment-3293836)). Also, I'm wondering if there is any utility to allow more complex expressions as exponent, for example `col ^ (m + n)`. What do you think @grjte?

ir/src/transition_constraints/graph.rs
```diff
@@ -84,6 +84,18 @@ impl AlgebraicGraph {
...
+            Expr::Exp(lhs, rhs) => {
+                // add base subexpression.
+                let lhs = self.insert_expr(*lhs, trace_columns)?;
+                // add exponent subexpression.
+                if let Expr::Constant(exponent) = *rhs {
+                    Ok(self.insert_op(Operation::Exp(lhs, exponent as usize)))
+                } else {
+                    Err(SemanticError::InvalidExprType(
+                        "Exponent can only be a constant".to_string(),
+                    ))
```
grjte, line 96: I think this error would be more useful if it includes information about the `lhs` and `rhs`.

parser/src/ast/mod.rs
```diff
@@ -48,6 +48,7 @@ pub enum Expr {
     Next(Identifier),
     Add(Box<Expr>, Box<Expr>),
     Subtract(Box<Expr>, Box<Expr>),
+    Exp(Box<Expr>, Box<Expr>),
```
grjte, line 51: this is super nitty, but since all of the other variants are full words, I think maybe we should use `Exponent` instead of `Exp`. Either that, or we could switch to `Sub`, `Mul`, `Const`, `Var`, which would be fine too (and maybe better?) Feel free to pick whichever you prefer, as long as we're consistent. (In fact, in the graph in the IR I probably should have used `Const`)

parser/src/lexer/mod.rs
```diff
@@ -81,6 +81,9 @@ pub enum Token {
     #[token("-")]
     Minus,
 
+    #[token("^")]
+    Exp,
+
```
grjte, line 86: nit: same thought about `Exp` vs `Exponent` as I left in the ast

codegen-winter/src/air/transition_constraints.rs
```diff
@@ -83,8 +83,17 @@ impl Codegen for Operation {
...
             }
+            Operation::Exp(l_idx, r_idx) => {
+                let lhs = l_idx.to_string(graph);
+                strings.push(lhs);
+
+                strings.push(String::from("^"));
+
+                let rhs = r_idx.to_string();
+                strings.push(rhs);
+            }
```
grjte, line 94: I have the same thought here that I left in the `mul` PR - I think using `format!` would be better.

parser/src/lexer/tests.rs
```diff
@@ -45,11 +45,16 @@ fn keywords_with_identifiers() {
 
 #[test]
 fn multi_arithmetic_ops() {
-    let source = "enf clk' - clk - 1 = 0";
+    let source = "enf clk'^2 - clk' - clk - 1 = 0";
```
grjte, line 48: Same comment here that I left in the `Mul` PR about modifying this test vs. adding new tests

parser/src/parser/grammar.lalrpop
```diff
@@ -87,8 +87,13 @@ TransitionConstraint: TransitionConstraint = {
...
 Expr: Expr = {
-    <lexpr: Expr> "+" <rexpr: NextExpr> => Expr::Add(Box::new(lexpr), Box::new(rexpr)),
-    <lexpr: Expr> "-" <rexpr: NextExpr> => Expr::Subtract(Box::new(lexpr), Box::new(rexpr)),
+    <lexpr: Expr> "+" <rexpr: ExpExpr> => Expr::Add(Box::new(lexpr), Box::new(rexpr)),
+    <lexpr: Expr> "-" <rexpr: ExpExpr> => Expr::Subtract(Box::new(lexpr), Box::new(rexpr)),
+    ExpExpr
+}
+
+ExpExpr: Expr = {
+    <lexpr: NextExpr> "^" <rexpr: NextExpr> => Expr::Exp(Box::new(lexpr), Box::new(rexpr)),
```
grjte, line 96: I don't think we want to allow a `NextExpr` on the right hand side here. Even when we support variables in the future, we don't want to allow trace columns here. I think this is actually the best place to restrict the rhs to a constant - then you don't have to check it later, since it will be handled during parsing.

parser/src/parser/tests/expressions.rs
```diff
@@ -5,6 +5,33 @@ use super::{
 
 // SECTIONS
 // ================================================================================================
+#[test]
+fn exponentiation() {
+    // the operation must be put into a source section, or parsing will fail
+    let source = "
+    transition_constraints:
+        enf clk'^2 + clk - 1 = 0";
```
grjte, line 13: Same thoughts I left in the `Mul` PR about more rigorous testing of edge cases & different combinations of operations.

grjte (changes requested):
Thanks @tohrnii! I left a few comments that are redundant with the Mul PR, so it makes sense to read that one first. Mostly it's the same theme about now being a good time to introduce more rigorous testing.

Regarding your comment - I agree with restricting the exponent to a constant for now. I think we can add the option for it to be a variable in the future when we add support for variables (since currently, identifiers are always trace columns). Similar thought with the complex expressions - I think this should be added when support for parens is added, but for now the only atoms allowed inside the parens should be constants. In the future when variable support is added, we would then allow atoms that are either identifiers or constants, and we'll then need to do more type-checking in the IR (to tell the difference between allowed identifiers like variables and disallowed ones like trace columns)

grjte:
> Similar thought with the complex expressions - I think this should be added when support for parens is added, but for now the only atoms allowed inside the parens should be constants. In the future when variable support is added, we would then allow atoms that are either identifiers or constants, and we'll then need to do more type-checking in the IR (to tell the difference between allowed identifiers like variables and disallowed ones like trace columns)

Actually, thinking about the complexity of excluding variables now when we'll want to include them again later - I don't think it's worth it. I think even in the parens PR we should just keep the exponent rhs as a constant, and support for complex expressions can be added when variables are added (probably v0.2). What do you think?

parser/src/ast/mod.rs
tohrnii: I agree keeping the concise version seems better. So changed these to `Sub`, `Mul`, `Const`, `Var`, etc.

parser/src/lexer/mod.rs
tohrnii: Went with `Exp` and changed the rest and `Number` to `Num`.

parser/src/parser/grammar.lalrpop
tohrnii: Makes sense. Thanks.

codegen-winter/src/air/boundary_constraints.rs
```diff
@@ -55,16 +55,19 @@ trait Codegen {
...
+            Self::Sub(lhs, rhs) => {
                 format!("{} - {}", lhs.to_string(), rhs.to_string())
             }
-            Self::Multiply(lhs, rhs) => {
+            Self::Mul(lhs, rhs) => {
                 format!("{} * {}", lhs.to_string(), rhs.to_string())
             }
+            Self::Exp(lhs, rhs) => {
+                format!("({}).exp({})", lhs.to_string(), rhs)
+            }
```
grjte, line 70: I think parentheses need to be added to the other expressions above like in the `Mul` PR. These can be revisited at the same time those ones are

grjte (changes requested):
Looks great, thank you @tohrnii! There's just one small issue which I left inline (the same one we just looked at for codegen in the mul PR)

nit: no need to change for this PR, but I would prefer for things like keyword renaming to be in its own `refactor` commit in the future

codegen-winter/src/air/boundary_constraints.rs
grjte: Actually, since this was a pre-existing issue I'll fix it separately and approve this PR

grjte (approved):

Issue #20, opened 2022-10-23
add support for parentheses


grjte:
We need to add parentheses `(` `)` to support grouping and more complex expressions

- [x] parser
- [x] codegen

grjte:
Closed by #23

PR #23, opened 2022-10-24
Add support for parenthesis

tohrnii:
Addressing #20.
Adds support for parentheses.
- [x] parser
- [x] codegen

### DIFF
--- a/parser/src/lexer/mod.rs
+++ b/parser/src/lexer/mod.rs
@@ -105,6 +105,12 @@ pub enum Token {
     #[token(".")]
     Dot,
 
+    #[token("(")]
+    Lparen,
+
+    #[token(")")]
+    Rparen,
+
     // UNDEFINED TOKENS AND TOKENS TO IGNORE
     // --------------------------------------------------------------------------------------------
     /// Error is returned on encountering unrecognized tokens.
--- a/parser/src/parser/tests/expressions.rs
+++ b/parser/src/parser/tests/expressions.rs
@@ -136,6 +136,29 @@ fn multi_multiplication() {
 }
 
 #[test]
+fn ops_with_parens() {
+    // the operation must be put into a source section, or parsing will fail
+    let source = "
+    transition_constraints:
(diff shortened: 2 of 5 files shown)


parser/src/lexer/mod.rs
```diff
@@ -102,6 +102,12 @@ pub enum Token {
     #[token(".")]
     Dot,
 
+    #[token("(")]
+    Lpar,
+
+    #[token(")")]
+    Rpar,
+
```
grjte, line 110: nit: for some reason, I find this strange to read, and if I didn't see the token next to it I'm not sure that I'd immediately understand what it means. Maybe it would be clearer just do `Lparen`, since it's not that long and "paren" is already an abbreviation (so it's still consistent with our use of abbreviations)

parser/src/parser/tests/expressions.rs
```diff
@@ -188,3 +229,34 @@ fn multi_arithmetic_ops_different_precedence() {
...
 }
+
+#[test]
+fn multi_arithmetic_ops_different_precedence_w_parens() {
+    // the operation must be put into a source section, or parsing will fail
+    let source = "
+    transition_constraints:
+        enf clk' + clk * (2 - 1) = 0";
+    // The precedence order of operations here is:
+    // 1. Parantheses
```
grjte, line 240: nit: There's a typo - this should be Parentheses

grjte (changes requested):
@tohrnii this looks great, thank you! I left 2 tiny nits. Can you rebase and update the grammar to handle the conflicts from the merged Exp PR? It might also be nice to add one more test that includes both exponentiation and parentheses.

tohrnii:
Thank you @grjte for the review. 
>  It might also be nice to add one more test that includes both exponentiation and parentheses.

I've updated the previous multi ops test that tests precedence between parens and exp. Please let me know if there should be multiple tests here instead.

grjte (approved):
Looks great, thanks!

Issue #15, opened 2022-10-19
add auxiliary trace handling


grjte:
Once #14 is handled, we'll need to add the IR and codegen for the auxiliary trace

- [x] add auxiliary trace to the IR
- [x] add aux trace boundary constraint handling
- [x] add aux trace transition constraint handling with random values in algebraic graph and handling of base vs. extension field
- [x] add aux trace codegen

grjte:
@tohrnii for all of the codegen you can reference the existing Miden air for the signatures.

One important note: with boundary constraints, there's a different type for main vs. aux assertion methods. When implementing `to_string` for constants, it should use `Felt::from` for assertions put into `get_assertions` and `E::from` for assertions put into `get_aux_assertions`

grjte:
Closed by #39

Issue #14, opened 2022-10-19
add support for random values


grjte:
We need access to random values when evaluating transition constraints over the auxiliary trace. The main complication with these is that they may be values in the extension field (and will be, in Miden's case), whereas everything that's been implemented so far (for the main trace) only requires values from the base field.

We also need to determine how random values are going to be referenced. Based on the [previous discussion](https://github.com/maticnetwork/miden/discussions/254#discussioncomment-3302668), it seems like we should refer to them by `$rand[0], $rand[1]`, etc.

- [x] lexer
- [x] parser
- [x] IR
- [x] codegen

grjte:
Closed by #39

Issue #11, opened 2022-10-19
Add handling of periodic columns


grjte:
Periodic columns are used for transition constraints. The syntax has already been defined, but we need to add support for them

- [x] add parsing of periodic columns
- [x] add periodic columns to IR (and IR's AlgebraicGraph)
- [x] update degree calculation for periodic columns
- [x] add codegen of periodic columns
- [x] update codegen of transition constraints to use periodic columns

grjte:
Closed by #40

Issue #10, opened 2022-10-19
Add handling of public inputs


grjte:
We need to add a way to specify boundary constraints that rely on public inputs.

- [x] clarify syntax for declaring & referring to public inputs in the Air DSL
- [x] handle parsing of public input declarations and references
- [x] add public inputs to IR
- [x] handle codegen of public inputs and boundary constraints that depend on public inputs

grjte:
@bobbinth I think we should keep the initial implementation of public inputs as simple as possible and add complexity iteratively, so I'm inclined towards fixed size vectors as [described here](https://github.com/maticnetwork/miden/discussions/254#discussioncomment-3444589). We can revisit things after we've added the other elements needed for v0.1 (periodic columns, auxiliary trace constraints, and random values).

To address the 2 points related to public inputs there:
> We may not know the size of public inputs ahead of time. This would mean we'd need to add support for vectors of unknown size.

I think we should revisit this later.

> If we need a public input which is just a single value, what's the best way to denote its type? Or should we assume that everything is a vector and something like foo: [1] is fine?

I think we should start by treating it as a vector and using [1] and then revisit later.

bobbinth:
Yes, I agree with the approach.

grjte:
Great! In that case, the initial declaration format for public inputs will be this:

```
pub_inputs:
  program_hash: [4]
  stack_inputs: [16]
  stack_outputs: [16]
```

grjte:
Closed by #39

PR #24, opened 2022-10-24
IR discussion for public inputs, periodic columns, random values, and aux constraints

grjte:
This purpose of this PR is for a high-level discussion of the IR for public inputs (#10), periodic columns (#11), random values (#14), and auxiliary constraint handling (#15). These are the rest of the features required for `v0.1`.

Once we've agreed on the general approach for the IR, I'll split this into smaller PRs instead. The relevant files for discussion are the ones in the `ir` crate, and this is probably most easily viewed directly rather than via diffs. The one major change to the parser was to split the `ast::Expr` expression type into `BoundaryExpr` and `TransitionExpr`.

notes:
- this hasn't been cleaned up or documented yet - I'll tidy it up when splitting into separate PRs.
- The IR still makes use of the same `BoundaryExpr` struct that is declared in `parser::ast`. It feels cleaner for the IR to have its own type, but we don't actually need anything different, so the same type is used in the `parser`, `ir` and `codegen`. We could move the type definition to the `ir` instead and then import it to the `parser`. Thoughts?
- the `TransitionConstraintDegree` that this uses is almost exactly the same as in Winterfell. We're not using `min_blowup_factor` or `get_evaluation_degree`, but we may want that in the future when we're supporting verifier code in Miden assembly and solidity. Any thoughts about clean reuse of this or a different approach?


ir/src/transition_constraints/graph.rs
```diff
@@ -125,21 +231,30 @@ impl Node {
...
+    /// An identifier for a cell in the specified column in the current row in the auxiliary trace.
+    /// The inner value is the index of the column within the trace.
+    AuxTraceCurrentRow(usize),
+    /// An identifier for a cell in the specified column in the next row in the auxiliary trace. The
+    /// inner value is the index of the column within the trace.
+    AuxTraceNextRow(usize),
+    /// An identifier for a periodic value from a specified periodic column. The inner value is the
+    /// index of the periodic column within the declared periodic columns. The periodic value made
+    /// available from the specified column is based on the current row of the trace.
+    PeriodicColumn(usize),
```
bobbinth, line 249: Would it make sense to track both the index as well as the cycle length for a periodic column here? This might make degree computation a bit more straight-forward.

ir/src/transition_constraints/mod.rs
```diff
@@ -21,17 +39,33 @@ impl TransitionConstraints {
...
+    }
+
+    /// Returns all transition constraints against the main execution trace as a vector of
+    /// [NodeIndex] where each index is the tip of the subgraph representing the constraint within
+    /// the [AlgebraicGraph].
+    pub fn main_constraints(&self) -> &[NodeIndex] {
+        &self.main_constraints
+    }
+
+    /// Returns a vector of the degrees of the transition contraints.
```
bobbinth, line 56: nit: typo in "contraints".

ir/src/transition_constraints/mod.rs
```diff
@@ -21,17 +39,33 @@ impl TransitionConstraints {
     // --- PUBLIC ACCESSORS -----------------------------------------------------------------------
 
     /// Returns a vector of the degrees of the transition contraints.
```
bobbinth, line 41: nit: typo in "contraints".

parser/src/ast/boundary_constraints.rs
```diff
@@ -1,4 +1,6 @@
-use super::{Expr, Identifier};
+use std::fmt::Display;
+
+use super::Identifier;
```
bobbinth, line 3: minor nit: extra blank line.

ir/src/lib.rs
```diff
@@ -71,38 +84,73 @@ impl AirIR {
...
+            .aux_degrees(&self.periodic_cycle_lens())
+    }
+
+    pub fn aux_transition_constraints(&self) -> &[NodeIndex] {
+        self.transition_constraints.aux_constraints()
     }
 
-    pub fn main_transition_graph(&self) -> &AlgebraicGraph {
-        self.main_transition_constraints.graph()
+    pub fn transition_graph(&self) -> &AlgebraicGraph {
```
bobbinth, line 152: minor nit (feel free to ignore): I don't love the name `transition_graph`. Something like `transition_constraint_expressions` would be more correct, but it is rather long - so, don't love that either.

parser/src/ast/mod.rs
```diff
@@ -14,9 +20,14 @@ pub use transition_constraints::*;
 pub struct Source(pub Vec<SourceSection>);
 
 /// Source is divided into SourceSections.
-/// There are 4 types of Source Sections:
+/// There are 6 types of Source Sections:
 /// - AirDef: Name of the air constraints module.
 /// - TraceCols: Trace Columns representing columns of the execution trace.
+/// - PublicInputs: Public inputs are each represented by a fixed-size array. They are not required,
+///   and there is no limit to the number of public inputs that can be specified.
```
bobbinth, line 27: This could be done in the future PRs: public inputs should actually be required. An AIR without public inputs is kind of meaningless.

ir/src/symbol_table.rs
```diff
@@ -0,0 +1,178 @@
...
+        columns: &[PeriodicColumn],
+    ) -> Result<(), SemanticError> {
+        for (index, column) in columns.iter().enumerate() {
+            validate_cycles(column)?;
+
+            self.insert_symbol(column.name(), IdentifierType::PeriodicColumn(index))?;
+            self.periodic_columns.push(column.values().to_vec());
+        }
+
+        Ok(())
```
bobbinth, line 109: Comment for the future: what should we do if two periodic columns are identical (same cycle length and values)? Should it be an error? A warning? Or something else?

bobbinth (changes requested):
Thank you! The overall structure looks great! I left a few comments inline - but all of them are fairly minor.

To address your questions:
- I do think we'll end up with a different struct in IR instead of using the same `BoundaryExpr` in both parser and IR  - but I think it is fine to keep as is now. We can make the call once we've implemented more complex boundary constraints (e.g., the ones needed for stack overflow table).
- I think the way you've implemented `TransitionConstraintDegree` works well for now - so, I wouldn't change it. I'm a bit hesitant to use the struct directly from Winterfell - but I don't have strong motivation for this hesitation.

examples/aux_trace.air
```diff
@@ -0,0 +1,15 @@
+def AuxTraceAir
+
+trace_columns:
+    aux: [p0, p1]
```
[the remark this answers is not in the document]
grjte: Yep, true! I wrote that in the docs actually. I just wanted to see how things were looking here, but I'll remove this file & we'll make it an error if `main` isn't defined

codegen-winter/src/air/boundary_constraints.rs
```diff
@@ -48,23 +49,17 @@ pub(super) fn add_fn_get_assertions(impl_ref: &mut Impl, ir: &AirIR) {
...
 /// Code generation trait for generating Rust code strings from boundary constraint expressions.
-trait Codegen {
-    fn to_string(&self) -> String;
-}
-
-impl Codegen for Expr {
+impl Codegen for BoundaryExpr {
     fn to_string(&self) -> String {
         match self {
             Self::Constant(value) => format!("Felt::new({})", value),
```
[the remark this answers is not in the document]
grjte: Hm I think it was left that way because our main trace assertions are Felt, but yes, it needs to be updated as part of adding codegen for aux assertions, or we need to handle them differently

codegen-winter/src/air/boundary_constraints.rs
```diff
@@ -1,6 +1,6 @@
-use super::{AirIR, Impl};
+use super::{AirIR, Codegen, Impl};
 
-use ir::Expr;
+use ir::BoundaryExpr;
```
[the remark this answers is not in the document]
grjte: Yep, these are all automatic imports. I'll clean them up, thanks

ir/src/symbol_table.rs
grjte: Good point. Probably a warning? But we could ask also fix it internally

ir/src/transition_constraints/graph.rs
grjte: Yes, that's better, thanks

examples/aux_trace.air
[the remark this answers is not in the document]
grjte: I opened #29

parser/src/ast/mod.rs
grjte: I've opened #30

ir/src/symbol_table.rs
grjte: I've opened #31

ir/src/lib.rs
grjte: I could call it `constraint_graph` if you prefer that?

bobbinth: Not sure if it is better, to be honest :) I'm fine keeping it as is.

ir/src/boundary_constraints.rs
```diff
@@ -1,6 +1,6 @@
 use std::collections::BTreeMap;
 
-use super::{Expr, Identifier, SemanticError, TraceColumns};
+use super::{BoundaryExpr, Identifier, IdentifierType, SemanticError, SymbolTable};
```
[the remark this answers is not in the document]
grjte: I've moved the `BTreeMap` import to `lib` and imported from there so we can update easily if we decide to.

grjte:
Closing this PR, as the work has been moved to #28, #32, and #33

PR #25, opened 2022-10-24
docs: add mdbook with lang specs & impl details

grjte:
This PR includes a first pass at documentation of the language specification and implementation for v0.1, including basic information on everything except the sections describing the `boundary_constraints` and `transition_constraints` source sections.


docs/src/description/declarations.md
```diff
@@ -0,0 +1,74 @@
...
+
+The following is a valid `trace_columns` block:
+
+```
+trace_columns:
+    main: [a, b, c]
+    aux: [d, e]
+```
+
+In the above example, the main execution trace for the Air has 3 columns which can be referenced by `a`, `b`, and `c`. Internally, these identifiers will reference the trace columns with indices 0, 1, and 2 respectively. Similarly, the auxiliary execution trace has 2 columns which can be referenced by `d` and `e`.
```
bobbinth, line 19: nit: I would capitalize "AIR".

docs/src/description/declarations.md
```diff
@@ -0,0 +1,74 @@
...
+
+## public_inputs
+
+A `public_inputs` section contains declarations for public inputs. Currently, each public input must be provided as a vector of a fixed size, but there is no limit to how many of them can be declared within the `public_inputs` section.
+
+**Public inputs are required.** There must be at least one pubic input declared.
+
+Each public input is described by an identifier and an array length `n` in the following format:
```
bobbinth, line 27: nit: probably should be "array of length `n`" (with "of").

docs/src/description/declarations.md
```diff
@@ -0,0 +1,74 @@
+# Type declaration sections
+
+## trace_columns
```
bobbinth, line 3: nit: I wonder if we should use just regular text for section names here and later. For example, "Trace columns" instead of "trace_columns".

bobbinth (approved):
Looks good! Thank you! I left a few minor comments inline - but PR is good even without addressing them.

docs/src/description/syntax.md
```diff
@@ -0,0 +1,119 @@
...
+  - `last`: used to access the value of a trace column at the last row of the trace. _It may only be used when defining boundary constraints._
+- `enf`: used to describe a single [constraint](./constraints.md).
+- `public_inputs`: used to declare the [source section](./structure.md) where the [public inputs are declared](./declarations.md). _They may only be referenced when defining boundary constraints._
+- `periodic_columns`: used to declare the [source section](./structure.md) where the [periodic columns are declared](./declarations.md). _They may only be referenced when defining transition constraints._
+- `trace_columns`: used to declare the [source section](./structure.md) where the [execution trace is described](./declarations.md).
+  - `main`: used to declare the main execution trace
+  - `aux`: used to declare the auxiliary execution trace
+- `transition_constraints`: used to declare the [source section](./structure.md) where the [transition constraints are described](./constraints.md#transition_constraints).
+  - `$rand`: used to [access random values](#accessors) provided by the verifier. _It may only be referenced when defining transition constraints._
```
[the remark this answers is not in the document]
grjte: Great point, thanks!

docs/src/description/declarations.md
grjte: Actually, I intended it as "Each public input is described by an identifier and an array length." I'll put `n` in parentheses.

docs/src/description/declarations.md
grjte: Sure, I'll update

PR #33, opened 2022-10-25
IR for v01 features (pub inputs, periodic columns, rand values)

grjte:
This PR handles adds public inputs (https://github.com/0xPolygonMiden/air-dsl/issues/10), periodic columns (https://github.com/0xPolygonMiden/air-dsl/issues/11), and random values (https://github.com/0xPolygonMiden/air-dsl/issues/14) to the IR and makes some minor `codegen` changes so things compile.

Left for the future:
- refactor the periodic column handling in the symbol table as suggested [here](https://github.com/0xPolygonMiden/air-dsl/pull/24#discussion_r1003608800)
- fix an error with the degree calculation for cycles (can be combined with the previous point)
- extensive testing of the IR


ir/src/transition_constraints/degree.rs
```diff
@@ -0,0 +1,87 @@
+// The [TransitionConstraintDegree] struct and documentation contained in this file were duplicated
+// from the Winterfell STARK prover library (https://github.com/novifinancial/winterfell), which is
+// licensed under the MIT license. The implementation in this file is a subset of the Winterfell
+// code.
+
+// The original code is available in the Winterfell library in the `air` crate:
+// https://github.com/novifinancial/winterfell/blob/main/air/src/air/transition/degree.rs
```
bobbinth, line 7: If this is intended to be a module level comment, i think the syntax is `//!`.

bobbinth (approved):
All looks good! Thank you! Just one minor doc-related comment.

ir/src/transition_constraints/degree.rs
grjte: Fixed, thank you!

Issue #16, opened 2022-10-19
Add end-to-end testing of full transpilation pipeline


grjte:
We should set up a structure for testing full files and add a few tests of full air files for v0.1 - some simple, and some that use the full v0.1 specification (e.g. `system.air` and `binary.air` that @tohrnii previously defined)

We may want to consider something like [expect_test](https://docs.rs/expect-test/latest/expect_test/) for this

tohrnii:
@grjte Can this issue be closed?

grjte:
Closed by #42

PR #34, opened 2022-10-26
Examples for testing and demonstration

grjte:
This PR adds simple example AIR definitions to the `examples` directory for binary constraints, public inputs, periodic columns, and constraints against the auxiliary trace using random values.

They can be used for demonstration purposes with the CLI or for testing for #16.


tohrnii (approved):
Looks good. Thank you @grjte, this is really helpful.

PR #35, opened 2022-10-26
fix: return column index for boundary conditions from accessors

tohrnii:
Small fix in the boundary constraints IR. We need the column index for the constraints while accessing it from IR during codegen for setting the column values in assertions. We should return a `Vec<(usize, &BoundaryExpr)>` instead of `Vec<&BoundaryExpr>` from access methods.

I didn't add any tests but tested it with a sample file:

```
trace_columns:
    main: [clk, fmp, ctx]
  
transition_constraints:
    enf clk' = clk + 1
  
boundary_constraints:
    enf clk.first = 0
    enf ctx.first = 0
    enf ctx.last = 1
```
Before fix:
```
let mut result = Vec::new();
result.push(Assertion::single(0, 0, Felt::from(0)));
result.push(Assertion::single(1, 0, Felt::from(0)));
let last_step = self.last_step();
result.push(Assertion::single(0, last_step, Felt::from(1)));
result
```

After fix: 
```
let mut result = Vec::new();
result.push(Assertion::single(0, 0, Felt::from(0)));
result.push(Assertion::single(2, 0, Felt::from(0)));
let last_step = self.last_step();
result.push(Assertion::single(2, last_step, Felt::from(1)));
result
```

### DIFF
--- a/ir/src/boundary_constraints.rs
+++ b/ir/src/boundary_constraints.rs
@@ -33,13 +33,13 @@ impl BoundaryConstraints {
     }
 
     /// Returns all of the boundary constraints for the first row of the main trace.
-    pub fn main_first(&self) -> Vec<&BoundaryExpr> {
-        self.main_first.values().collect()
+    pub fn main_first(&self) -> Vec<(usize, &BoundaryExpr)> {
+        self.main_first.iter().map(|(k, v)| (*k, v)).collect()
     }
 
     /// Returns all of the boundary constraints for the final row of the main trace.
-    pub fn main_last(&self) -> Vec<&BoundaryExpr> {
-        self.main_last.values().collect()
+    pub fn main_last(&self) -> Vec<(usize, &BoundaryExpr)> {
+        self.main_last.iter().map(|(k, v)| (*k, v)).collect()
     }
 
     /// Returns the total number of boundary constraints for the aux trace.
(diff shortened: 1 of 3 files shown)


ir/src/boundary_constraints.rs
```diff
@@ -33,13 +33,21 @@ impl BoundaryConstraints {
...
 
     /// Returns all of the boundary constraints for the first row of the main trace.
-    pub fn main_first(&self) -> Vec<&BoundaryExpr> {
-        self.main_first.values().collect()
+    pub fn main_first(&self) -> Vec<(usize, &BoundaryExpr)> {
+        let mut constraints = Vec::new();
+        for (k, v) in &self.main_first {
+            constraints.push((*k, v))
+        }
+        constraints
```
grjte, line 41: nit: I would replace this with the following code to simplify it, and I would make this change for all 4 methods modified in this file.

```
        self.main_first.iter().map(|(k, v)| (*k, v)).collect()
```

grjte (changes requested):
Great catch @tohrnii! I left one nit inline. Also, it would be great to use this example as a test in the future. Can you add a note to the codegen unit testing issue so that we include this example once the unit tests are set up?

ir/src/boundary_constraints.rs
tohrnii: Thanks. Fixed.

grjte (approved):
Looks great, thanks again!

PR #36, opened 2022-10-26
Complete v0.1 mdbook

grjte:
This PR completes the documentation for v0.1. It documents how to define constraints in the `boundary_constraints` and `transition_constraints` sections and makes a few other minor updates for clarity.

Remaining:
- we'll update the name throughout these docs when we update the name throughout the codebase just prior to releasing v0.1


docs/src/description/constraints.md
```diff
@@ -2,8 +2,158 @@
 
 ## Boundary constraints (`boundary_constraints`)
 
-TODO
+The `boundary_constraints` section consists of expressions describing the expected value of columns in the main or auxiliary traces at the specified boundary. Column boundaries can be selected using boundary accessors. Valid boundary accessors are `.first`, which selects the first cell of the column to which it is applied, and `.last`, which selects the last cell of the column column to which it is applied.
+
+**Boundary constraints are optional.** It is equally valid to define an empty `boundary_constraints` section or to omit the `boundary_constraints` section entirely.
```
bobbinth, line 7: I think this is fine for now, but in the future we should change this. A computation without any boundary constraints is kind of meaningless - so, we should make them mandatory.

docs/src/description/constraints.md
```diff
@@ -2,8 +2,158 @@
...
+
+Boundary constraints that are defined against auxiliary columns or that use random values from the built-in `$rand` array will be identified as auxiliary constraints.
```
bobbinth, line 9: I'd say that boundary constraints which use `$rand` values must be defined against auxiliary trace columns (defining them against main trace columns should be an error).

docs/src/description/constraints.md
```diff
@@ -2,8 +2,158 @@
...
+    # these are main constraints that use public input values.
+    enf a.first = stack_inputs[0]
+    enf a.last = stack_outputs[0]
+
+    # these are auxiliary constraints that use public input values.
+    enf p0.first = stack_inputs[1]
+    enf p0.last = stack_outputs[1]
+
+    # these are auxiliary constraints that use random values from the verifier.
+    enf b.first = a + $rand[0]
```
bobbinth, line 77: This should be an error.

docs/src/description/constraints.md
```diff
@@ -2,8 +2,158 @@
...
+    stack_outputs: [16]
+
+boundary_constraints:
+    # these are main constraints that use public input values.
+    enf a.first = stack_inputs[0]
+    enf a.last = stack_outputs[0]
+
+    # these are auxiliary constraints that use public input values.
+    enf p0.first = stack_inputs[1]
+    enf p0.last = stack_outputs[1]
```
bobbinth, line 74: Technically, this should be valid - but I'll need to think about it a bit more.

docs/src/description/constraints.md
```diff
@@ -2,8 +2,158 @@
...
+    enf b.first = a + $rand[0]
+    enf p1.first = (stack_inputs[2] + $rand[0]) * (stack_inputs[3] + $rand[1])
+```
 
 ## Transition constraints (`transition_constraints`)
 
-TODO
+The `transition_constraints` section consists of expressions describing constraints that must be true at each row of the execution trace in order for the proof to be valid.
+
+**Transition constraints are optional.** It is equally valid to define an empty `transition_constraints` section or to omit the `transition_constraints` section entirely.
```
bobbinth, line 85: It's ok to leave it like this for now - but in the future we should change this. A computation without transition constraints is probably meaningless. So, we should make transition constraints mandatory (i.e., there should be at least one).

docs/src/description/constraints.md
```diff
@@ -2,8 +2,158 @@
...
+```
+def TransitionConstraintsExample
+
+trace_columns:
+    main: [a, b]
+    aux: [p]
+
+public_inputs:
+    # at least one public inputs is required.
+    program_hash: [4]
```
bobbinth, line 112: nit: since we don't care about what is actually here should we say something like:
```
public_inputs:
    <omitted for bravety>
```

docs/src/description/constraints.md
```diff
@@ -2,8 +2,158 @@
...
+
+transition_constraints:
+    # this is a main constraint that uses a periodic column.
+    enf a' = k * a
+
+    # this is an auxiliary constraint that uses a periodic column.
+    enf p0' = k * p0
+
+    # these are auxiliary constraints that use random values from the verifier.
+    enf b = a + $rand[0]
```
bobbinth, line 157: This probably should be invalid as this constraint should never be satisfied: the right hand side is an extension field element, while the left hand side is a base field element. If the value in `$rand[0]` is truly random, these will never be equal.

We'll need to think about a general rule which can describe this.

bobbinth (changes requested):
Thank you! Looks good! I left a few comments inline, many of them can be addresses in future PRs.

docs/src/description/constraints.md
grjte: I'm glad I'm catching all the edge cases so we can discuss them! :D

docs/src/description/constraints.md
grjte: Great - we should create a "bug" issue for this with a note to update the docs when it is resolved

docs/src/description/constraints.md
grjte: Yeah - I aimed to demonstrate everything that's currently syntactically valid. This is a harder one to check than the boundary constraints, so it probably merits its own issue

docs/src/description/constraints.md
grjte: I added #44

docs/src/description/constraints.md
grjte: I added #45

docs/src/description/constraints.md
grjte: I've opened #46

docs/src/description/constraints.md
grjte: I've opened #46

docs/src/description/constraints.md
grjte: I've opened #47

grjte:
@bobbinth this PR is ready for a final review. In addition to the points previously addressed, I made a few more minor tweaks to incorporate fixes added while this was pending and also a to address a few small points @tohrnii noticed. 

@tohrnii your review is optional. (I know you are unlikely to be available today/tomorrow)

docs/book.toml
```diff
@@ -3,4 +3,7 @@ authors = ["grjte"]
 language = "en"
 multilingual = false
 src = "src"
-title = "Polygon Miden's AirScript"
+title = "AirScript by Polygon Miden"
+
+[output.linkcheck]
+warning-policy = "ignore"
```
bobbinth, line 9: With this line, I couldn't build the book locally. But once I deleted it, it worked fine. Maybe there are some broken links somewhere?

bobbinth (approved):
Looks good! Thank you! I left one comment regarding build the `mdBook` inline.

Also, we probably shouldn't check in `.DS_Store` file.

grjte:
Great, thanks for catching both of these! I'll fix before merging

PR #39, opened 2022-10-26
Add codegen for public inputs and auxiliary columns

tohrnii:
Adds codegen for public inputs and auxiliary columns.
Partly Addressing #10, #14 and #15.

### DIFF
--- a/codegen-winter/src/air/boundary_constraints.rs
+++ b/codegen-winter/src/air/boundary_constraints.rs
@@ -21,20 +21,21 @@ pub(super) fn add_fn_get_assertions(impl_ref: &mut Impl, ir: &AirIR) {
         let assertion = format!(
             "result.push(Assertion::single({}, 0, {}));",
             col_idx,
-            constraint.to_string()
+            constraint.to_string(false)
         );
         get_assertions.line(assertion);
     }
 
     // add the constraints for the last boundary.
     let last_constraints = ir.main_last_boundary_constraints();
+
     if !last_constraints.is_empty() {
         get_assertions.line("let last_step = self.last_step();");
         for (col_idx, constraint) in last_constraints {
             let assertion = format!(
                 "result.push(Assertion::single({}, last_step, {}));",
                 col_idx,
-                constraint.to_string()
+                constraint.to_string(false)
             );
             get_assertions.line(assertion);
         }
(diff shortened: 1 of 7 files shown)


codegen-winter/src/air/mod.rs
```diff
@@ -109,19 +121,19 @@ let context = AirContext::new_multi_segment(
...
+    new.line("Self { context, stack_inputs: public_inputs.stack_inputs, stack_outputs: public_inputs.stack_outputs, overflow_addrs: public_inputs.overflow_addrs }");
 }
 
 // RUST STRING GENERATION
 // ================================================================================================
 
 /// Code generation trait for generating Rust code strings from boundary constraint expressions.
 pub trait Codegen {
-    fn to_string(&self) -> String;
+    fn to_string(&self, is_aux_assertion: bool) -> String;
```
grjte, line 132: nit: I would name this differently - maybe `is_aux_constraint`

codegen-winter/src/air/mod.rs
```diff
@@ -109,19 +121,19 @@ let context = AirContext::new_multi_segment(
     new.line(context);
 
     // return initialized Self.
-    new.line("Self { context }");
+    new.line("Self { context, stack_inputs: public_inputs.stack_inputs, stack_outputs: public_inputs.stack_outputs, overflow_addrs: public_inputs.overflow_addrs }");
```
grjte, line 124: Ah there's a tricky thing here - these things shouldn't be hard-coded. Yes, they are the public inputs that we are using in Miden, but the public inputs in the Air struct should only depend on the public inputs that are declared in the air file.

codegen-winter/src/air/boundary_constraints.rs
```diff
@@ -21,23 +21,46 @@ pub(super) fn add_fn_get_assertions(impl_ref: &mut Impl, ir: &AirIR) {
...
+        );
+        get_assertions.line(assertion);
+    }
+
+    // add the constraints for the auxiliary columns for the first boundary.
+    for (col_idx, constraint) in ir.aux_first_boundary_constraints() {
+        let assertion = format!(
+            "result.push(Assertion::single({}, 0, {}));",
+            col_idx,
+            constraint.to_string(true)
```
grjte, line 34: These ones and the aux_last_boundary_constraints should both be added to a `get_aux_assertions` function instead, so there should be a new function `add_fn_get_aux_assertions` for handling that

codegen-winter/src/air/mod.rs
```diff
@@ -28,7 +35,10 @@ fn add_air_struct(scope: &mut Scope, name: &str) {
     scope
         .new_struct(name)
         .vis("pub")
-        .field("context", "AirContext<Felt>");
+        .field("context", "AirContext<Felt>")
+        .field("stack_inputs", "Vec<Felt>")
+        .field("stack_outputs", "Vec<u64>")
+        .field("overflow_addrs", "Vec<u64>");
```
grjte, line 41: the base implementation of the public inputs struct needs to depend on the public inputs that are defined in the original Air file. They shouldn't be hard-coded

codegen-winter/src/air/public_inputs.rs
```diff
@@ -0,0 +1,27 @@
...
+/// Updates the provided scope with a public inputs.
+pub(super) fn add_public_inputs_struct(scope: &mut Scope, name: &str) {
+    // define the custom Air struct.
+    scope
+        .new_struct(name)
+        .vis("pub")
+        .field("program_hash", "Digest")
+        .field("stack_inputs", "Vec<Felt>")
+        .field("stack_outputs", "Vec<u64>")
+        .field("overflow_addrs", "Vec<u64>");
```
grjte, line 12: Similar to the comment I left in the other spots - these need to be dynamically defined based on the provided AirIR

grjte (changes requested):
Thanks @tohrnii it's a great start! There are a couple of significant points that I left inline. Everything aside from these 2 points is looking good though

codegen-winter/src/air/boundary_constraints.rs
tohrnii: Thanks, fixed.

codegen-winter/src/air/mod.rs
tohrnii: Fixed.

codegen-winter/src/air/public_inputs.rs
tohrnii: Fixed.

codegen-winter/src/air/mod.rs
```diff
@@ -2,34 +2,45 @@ use super::{AirIR, Impl, Scope};
...
+    let air_struct = scope
         .new_struct(name)
         .vis("pub")
         .field("context", "AirContext<Felt>");
 
+    // add public inputs
+    for (pub_input, pub_input_size) in ir.public_inputs() {
+        air_struct.field(pub_input, format!("[Felt; {}]", pub_input_size));
+    }
+
```
tohrnii, line 43: @grjte Since we already have a PublicInputs struct defined, should this just have that as the field or should we not define the PublicInputs struct?

ir/src/symbol_table.rs
```diff
@@ -35,6 +35,8 @@ pub(super) struct SymbolTable {
     /// A map of the Air's periodic columns using the index of the column within the declared
     /// periodic columns as the key and the vector of periodic values as the value
     periodic_columns: Vec<Vec<u64>>,
+
+    public_inputs: Vec<(String, usize)>,
```
grjte, line 39: nit: I would add either a doc comment or a TODO here so we fill it in later

ir/src/lib.rs
```diff
@@ -81,11 +82,14 @@ impl AirIR {
             }
         }
 
+        let public_inputs = symbol_table.get_public_inputs();
+
```
grjte, line 86: Since we also need the periodic columns, I think it would be better to replace the `into_periodic_columns` method with a single `into_declarations` method that consumes the `symbol_table` and returns a tuple of `(public_inputs, periodic_columns)`

ir/src/lib.rs
```diff
@@ -24,6 +24,7 @@ pub struct AirIR {
     periodic_columns: Vec<Vec<u64>>,
     boundary_constraints: BoundaryConstraints,
     transition_constraints: TransitionConstraints,
+    public_inputs: Vec<(String, usize)>,
```
grjte, line 27: nit: I would move this above the constraints, probably above the `periodic_columns`

codegen-winter/src/air/mod.rs
grjte: I think either way is fine. Let's stick with what you've done for now, where it's directly in the struct, since it's the simplest option. When we revisit public inputs to think about more complex implementations we can think about changing this.

codegen-winter/src/air/public_inputs.rs
```diff
@@ -0,0 +1,28 @@
...
+    // add a constructor for public inputs
+    base_impl
+        .new_fn("new")
+        .arg("program_hash", "Digest")
+        .arg("stack_inputs", "Vec<Felt>")
+        .arg("stack_outputs", "Vec<u64>")
+        .arg("overflow_addrs", "Vec<u64>")
+        .vis("pub")
+        .ret("Self")
+        .line("Self { program_hash, stack_inputs, stack_outputs, overflow_addrs }");
```
grjte, line 27: I think it's just an oversight, but these argument names shouldn't be hard-coded, and all of the types should be the same: `Vec<Felt>`

grjte (changes requested):
This looks much closer. I left a few more comments inline

ir/src/lib.rs
tohrnii: Thanks, fixed.

codegen-winter/src/air/public_inputs.rs
tohrnii: Thanks, fixed. Should the types be fixed size arrays or vectors?

grjte: Actually you can make them fixed size arrays. Good point :)

codegen-winter/src/air/public_inputs.rs
```diff
@@ -0,0 +1,33 @@
+use ir::AirIR;
+
+use super::Scope;
+
+/// Updates the provided scope with a public inputs.
+pub(super) fn add_public_inputs_struct(scope: &mut Scope, ir: &AirIR) {
+    let name = "PublicInputs";
+    // define the custom Air struct.
```
grjte, line 8: nit: this comment should probably say "PublicInputs" - same for the other comments in this file

codegen-winter/src/air/public_inputs.rs
```diff
@@ -0,0 +1,33 @@
+use ir::AirIR;
+
```
grjte, line 2: minor nit: there's an extra line here

grjte (changes requested):
This looks great now! 3 minor nits, then we're done!

grjte:
If I'm not mistaken, this also takes care of #14 , right? So #10, #14, #15 can all be closed with this PR?

tohrnii:
> If I'm not mistaken, this also takes care of #14 , right? So #10, #14, #15 can all be closed with this PR?

Ah yes, it should.

grjte (approved):
Thanks @tohrnii, great work!

PR #40, opened 2022-10-27
Add codegen for periodic columns

tohrnii:
Partly addressing #11.

### DIFF
--- a/codegen-winter/src/air/periodic_columns.rs
+++ b/codegen-winter/src/air/periodic_columns.rs
@@ -1,0 +1,32 @@
+use super::{AirIR, Impl};
+use ir::PeriodicColumns;
+
+pub(super) fn add_fn_get_periodic_column_values(impl_ref: &mut Impl, ir: &AirIR) {
+    // define the function.
+    let get_periodic_column_values = impl_ref
+        .new_fn("get_periodic_column_values")
+        .arg_ref_self()
+        .ret("Vec<Vec<Felt>>");
+
+    // output the periodic columns.
+    get_periodic_column_values.line(ir.periodic_columns().to_string());
+}
+
+/// Code generation trait for generating Rust code strings from Periodic Columns.
+trait Codegen {
+    fn to_string(&self) -> String;
+}
+
+impl Codegen for PeriodicColumns {
+    fn to_string(&self) -> String {
+        let mut columns = vec![];
+        for column in self {
+            let mut rows = vec![];
+            for row in column {
+                rows.push(format!("Felt::from({})", row));
+            }
+            columns.push(format!("vec![{}]", rows.join(", ")));
+        }
+        format!("vec![{}]", columns.join(", "))
+    }
+}
--- a/ir/src/lib.rs
+++ b/ir/src/lib.rs
@@ -15,14 +15,17 @@ pub use transition_constraints::{NodeIndex, TransitionConstraintDegree};
 mod error;
 use error::SemanticError;
 
+pub type PublicInputs = Vec<(String, usize)>;
+pub type PeriodicColumns = Vec<Vec<u64>>;
+
 /// Internal representation of an AIR.
 ///
 /// TODO: docs
 #[derive(Default, Debug)]
 pub struct AirIR {
     air_name: String,
-    public_inputs: Vec<(String, usize)>,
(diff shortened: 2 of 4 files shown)


codegen-winter/src/air/periodic_columns.rs
```diff
@@ -0,0 +1,12 @@
...
+
+pub(super) fn add_fn_get_periodic_column_values(impl_ref: &mut Impl, ir: &AirIR) {
+    // define the function.
+    let get_periodic_column_values = impl_ref
+        .new_fn("get_periodic_column_values")
+        .arg_ref_self()
+        .ret("Vec<Vec<Felt>>");
+
+    // output the periodic columns.
+    get_periodic_column_values.line(format!("{:?}", ir.periodic_columns()));
```
grjte, line 11: We shouldn't use a debug print here. If formatting this way, it should be `"{}"`. Also, I tested the output, and it returned an array with arrays of integers. We need a vector and `Felt`.

grjte (changes requested):
Thanks @tohrnii it looks good overall, but there's one issue with types.

codegen-winter/src/air/periodic_columns.rs
```diff
@@ -0,0 +1,31 @@
...
+    // output the periodic columns.
+    get_periodic_column_values.line(format!("{}", ir.periodic_columns().to_string()));
+}
+
+/// Code generation trait for generating Rust code strings from Periodic Columns.
+trait Codegen {
+    fn to_string(&self) -> String;
+}
+
+impl Codegen for Vec<Vec<u64>> {
```
tohrnii, line 19: Should this be a named type? We can define alias `PeriodicColumns` for `Vec<Vec<u64>>`?

codegen-winter/src/air/periodic_columns.rs
tohrnii: Thanks, fixed.

codegen-winter/src/air/periodic_columns.rs
grjte: Sure, sounds good! It might make sense to do this in the IR and do the same thing with the public inputs output type. What do you think?

tohrnii: Sounds good.

tohrnii: Done.

ir/src/lib.rs
```diff
@@ -15,14 +15,17 @@ pub use transition_constraints::{NodeIndex, TransitionConstraintDegree};
 mod error;
 use error::SemanticError;
 
+pub type PublicInputs = Vec<(String, usize)>;
+pub type PeriodicColumns = Vec<Vec<u64>>;
```
bobbinth, line 19: Should these be `Felt`'s even at the IR level already?

We don't need to change this now, but it might be helpful in the future when may want to evaluate constraints at a random point.

codegen-winter/src/air/periodic_columns.rs
```diff
@@ -0,0 +1,32 @@
...
+        let mut columns = vec![];
+        for column in self {
+            let mut rows = vec![];
+            for row in column {
+                rows.push(format!("Felt::from({})", row));
+            }
+            columns.push(format!("vec![{}]", rows.join(", ")));
+        }
+        format!("vec![{}]", columns.join(", "))
+    }
```
bobbinth, line 31: A thought for the future: instead of doing conversions from `u64` to `Felt` here, we could do the following:

1. Define static arrays for all periodic columns such that these arrays already contain `Felt`'s.
2. Inside `get_periodic_column_values()` convert these arrays to vectors (which should be just memory copy operations) and return.

The performance benefit of the above is probably very minor - so, definitely not a high priority item.

bobbinth (approved):
Looks good to me! I left a couple of comments for the future inline.

codegen-winter/src/air/periodic_columns.rs
tohrnii: Makes sense. Thanks. I've added an issue for this in #43.

tohrnii:
I'm merging this for now. There are some bugs in the generated rust code which will be addressed in #41.

PR #41, opened 2022-10-27
Fix compilation issues

tohrnii:


tohrnii:
@bobbinth @grjte The generated rust files after this PR can be viewed in PR #42 in the `air-dsl/tests/` directory.

air-dsl/Cargo.toml
```diff
@@ -10,3 +10,6 @@ codegen-winter = { package = "codegen-winter", path = "../codegen-winter", versi
 structopt = "0.3.26"
 env_logger = "0.9"
 log = { version = "0.4", default-features = false }
+winter-math = { package = "winter-math", version = "0.4", default-features = false }
+winter-utils = { package = "winter-utils", version = "0.4", default-features = false }
+winter-air = { package = "winter-air", version = "0.4", default-features = false }
```
grjte, line 15: nit: these imports don't seem to be required by the changes in this PR. I'm guessing maybe they're needed for the testing PR instead?

codegen-winter/src/air/boundary_constraints.rs
```diff
@@ -98,14 +98,14 @@ impl Codegen for BoundaryExpr {
         match self {
             Self::Const(value) => {
                 if is_aux_constraint {
-                    format!("E::from({})", value)
+                    format!("E::from({} as u64)", value)
```
grjte, line 101: This cast shouldn't be necessary, since we know that value is always a u64. We should use "_u64" instead of casting so that it is defined with the correct type in the first place, e.g.:

```
format!("E::from({}_u64)", value)
```

This should be updated in all equivalent cases where the `as u64` cast is used throughout these air changes

codegen-winter/src/air/boundary_constraints.rs
```diff
@@ -98,14 +98,14 @@ impl Codegen for BoundaryExpr {
...
                 } else {
-                    format!("Felt::from({})", value)
+                    format!("Felt::new({})", value)
                 }
             }
             Self::PubInput(name, index) => format!("self.{}[{}]", name, index),
             Self::Rand(index) => {
-                format!("aux_rand_elements[{}]", index)
+                format!("aux_rand_elements.get_segment_elements({})[0]", index)
```
grjte, line 108: I think the index is being put in the wrong place here. First we want to get the random elements for the 1st aux segment (at index 0), then we want to index into them by `index`.

```
format!("aux_rand_elements.get_segment_elements(0)[{}]", index)
```

codegen-winter/src/air/public_inputs.rs
```diff
@@ -29,4 +29,21 @@ pub(super) fn add_public_inputs_struct(scope: &mut Scope, ir: &AirIR) {
     for (pub_input, pub_input_size) in ir.public_inputs() {
         new_fn.arg(pub_input, format!("[Felt; {}]", pub_input_size));
     }
+
+    add_serializable_impl(scope, pub_inputs_values)
+}
+
+fn add_serializable_impl(scope: &mut Scope, pub_input_values: Vec<String>) {
```
grjte, line 36: nit: it would be good to add a doc comment here

codegen-winter/src/air/transition_constraints.rs
```diff
@@ -40,18 +40,18 @@ pub(super) fn add_fn_evaluate_aux_transition(impl_ref: &mut Impl, ir: &AirIR) {
...
-        .arg("main_frame", "&EvaluationFrame<F>")
-        .arg("aux_frame", "&EvaluationFrame<E>")
-        .arg("_periodic_values", "&[F]")
-        .arg("aux_rand_elements", "&AuxTraceRandElements<E>")
-        .arg("result", "&mut [E]");
+        .arg("main_frame", "&EvaluationFrame<E>")
+        .arg("aux_frame", "&EvaluationFrame<F>")
+        .arg("_periodic_values", "&[E]")
+        .arg("aux_rand_elements", "&AuxTraceRandElements<F>")
+        .arg("result", "&mut [F]");
```
grjte, line 50: These are now backwards from what we normally do and the way the signature is defined in Winterfell and in the existing Miden Air, so I think we should stick to the previous lettering.

codegen-winter/src/air/transition_constraints.rs
```diff
@@ -40,18 +40,18 @@ pub(super) fn add_fn_evaluate_aux_transition(impl_ref: &mut Impl, ir: &AirIR) {
     // define the function.
     let evaluate_aux_transition = impl_ref
         .new_fn("evaluate_aux_transition")
-        .generic("E: FieldElement<BaseField = Felt> + ExtensionOf<F>")
-        .generic("F: FieldElement<BaseField = Felt>")
+        .generic("E: FieldElement<BaseField = Felt>")
+        .generic("F: FieldElement<BaseField = Felt> + ExtensionOf<E>")
```
grjte, line 44: These are now backwards from what we usually do, so we should stick to opposite lettering that we had before (`F` for the base field and `E` for extension). I assume you made the change due to compilation errors, but maybe if you reverse the order of the generics the issue will be resolved?

```
        .generic("F: FieldElement<BaseField = Felt>")
        .generic("E: FieldElement<BaseField = Felt> + ExtensionOf<F>")
```

If possible, it would actually be better to have a "where" clause exactly like in Miden/Winterfell, if `codegen` supports that

codegen-winter/src/air/transition_constraints.rs
```diff
@@ -94,7 +94,7 @@ impl Codegen for Operation {
                 format!("periodic_values[{}]", col_idx)
             }
             Operation::RandomValue(idx) => {
-                format!("aux_rand_elements[{}]", idx)
+                format!("aux_rand_elements.get_segment_elements({})[0]", idx)
```
grjte, line 97: I think the index is being put in the wrong place here. First we want to get the random elements for the 1st aux segment (at index 0), then we want to index into them by `idx`.

```
format!("aux_rand_elements.get_segment_elements(0)[{}]", idx)
```

codegen-winter/src/imports.rs
```diff
@@ -3,16 +3,18 @@ use super::Scope;
...
-    scope.import("winter_air::Assertion", "Assertion");
-    scope.import("winter_air::AirContext", "AirContext");
-    scope.import("winter_air::Air", "Air");
+    scope.import("winter_air", "TransitionConstraintDegree");
+    scope.import("winter_air", "TraceInfo");
+    scope.import("winter_air", "ProofOptions as WinterProofOptions");
+    scope.import("winter_air", "EvaluationFrame");
+    scope.import("winter_air", "Assertion");
+    scope.import("winter_air", "AirContext");
+    scope.import("winter_air", "Air");
```
grjte, line 12: nit: since the rest of it is nicely alphabetized, I would put these 3 below the "AuxTraceRandElements" import

grjte (changes requested):
Thanks for fixing these! I left a few comments inline

air-dsl/Cargo.toml
tohrnii: I was using them to test for compilation issues. Removed.

codegen-winter/src/air/boundary_constraints.rs
tohrnii: Fixed, thanks.

codegen-winter/src/air/boundary_constraints.rs
tohrnii: Fixed, thanks.

codegen-winter/src/air/transition_constraints.rs
tohrnii: Fixed, thanks. Codegen does support where clauses.

codegen-winter/src/air/transition_constraints.rs
tohrnii: Fixed, thanks. Codegen does support where clauses.

codegen-winter/src/air/transition_constraints.rs
tohrnii: Fixed, thanks.

grjte (approved):
This looks good to me. Thanks @tohrnii!

codegen-winter/src/air/transition_constraints.rs
grjte: Nice, that's great

Issue #38, opened 2022-10-26
refactor IR for periodic columns


grjte:
- [x] Add the cycle length of the periodic column to the `IdentifierType::PeriodicColumn` enum variant
- [x] update the insertion of periodic columns into the graph to include the cycle length
- [x] change the computation of degrees so that nothing needs to be passed to the method

grjte:
Closed by #49

PR #49, opened 2022-10-28
Refactor IR for periodic columns & update degree computation

grjte:
This PR addresses #38, refactoring the IR for periodic columns and simplifying the computation of degrees for transition constraints with cycles


bobbinth (approved):
All looks good! Thank you!

Issue #27, opened 2022-10-25
Better handling of parantheses for nested expressions during codegen


tohrnii:
During codegen for an operation, we need to add parentheses around compound expressions. However, parentheses should not be added in case the expression is not an arithmetic operation. We should add a check in the [to_string()](https://github.com/0xPolygonMiden/air-dsl/blob/main/codegen-winter/src/air/transition_constraints.rs#L54) method to add the parentheses only if the expression is an arithmetic operation.

tohrnii:
Closed by #155

PR #50, opened 2022-10-28
fix: only add parens around arithmetic operations

tohrnii:
Addressing #27

### DIFF
--- a/codegen-winter/src/air/boundary_constraints.rs
+++ b/codegen-winter/src/air/boundary_constraints.rs
@@ -93,7 +93,6 @@ pub(super) fn add_fn_get_aux_assertions(impl_ref: &mut Impl, ir: &AirIR) {
 
 /// Code generation trait for generating Rust code strings from boundary constraint expressions.
 impl Codegen for BoundaryExpr {
-    // TODO: Only add parentheses in Add/Sub/Mul/Exp if the expression is an arithmetic operation.
     fn to_string(&self, is_aux_constraint: bool) -> String {
         match self {
             Self::Const(value) => {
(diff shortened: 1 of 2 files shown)


codegen-winter/src/air/boundary_constraints.rs
```diff
@@ -109,28 +108,48 @@ impl Codegen for BoundaryExpr {
...
     }
 }
+
+/// Checks whether the boundary expression is an arithmetic operation.
+fn is_arithmetic_expr(boundary_expr: &BoundaryExpr) -> bool {
+    matches!(
+        boundary_expr,
+        BoundaryExpr::Add(_, _) | BoundaryExpr::Sub(_, _) | BoundaryExpr::Mul(_, _)
+    )
+}
```
grjte, line 155: This method is going to produce extra parentheses. For example, `a * b * c` will result in `a * (b * c)` unnecessarily. I think a better way to do this would be to think about operator precedence (or binding power), and use parens when the sub expression has a lower precedence. We could probably do something elegant like reverse Pratt parsing here.

grjte (changes requested):
Thanks @tohrnii. This works, but I think there could be a better approach. I left my thoughts inline, but I also think we should pause this and address #68 first, since those changes will affect this trait.

codegen-winter/src/air/boundary_constraints.rs
tohrnii: Got it. Thank you. Will look into this later this week.

tohrnii:
superseded by #115

Issue #45, opened 2022-10-28
Throw error when transition_constraints is empty or omitted


grjte:
We should make it a requirement to have transition constraints. When this is updated, the mdbook docs section about `transition_constraints` should also be updated.

- [ ] update in the parser: transition_constraints sections should have at least one transition constraint
- [ ] update in the IR: at least one transition_constraints source section should be defined
- [ ] update in mdbook docs as noted here: https://github.com/0xPolygonMiden/air-dsl/pull/36#discussion_r1006190244

tohrnii:
Closed by #52.

Issue #44, opened 2022-10-28
Throw error when boundary_constraints is empty or omitted


grjte:
We should make it a requirement to have boundary constraints. When this is updated, the mdbook docs section about `boundary_constraints` should also be updated.

- [ ] update in the parser: boundary_constraints sections should have at least one boundary constraint
- [ ] update in the IR: at least one boundary_constraints source section should be defined
- [ ] update in mdbook docs as noted here: https://github.com/0xPolygonMiden/air-dsl/pull/36#discussion_r1006185096

tohrnii:
Closed by #52.

Issue #30, opened 2022-10-25
Throw error when `public_inputs` is empty or omitted


grjte:
At least one public input should be required, so this source section should be required and non-empty.

We should add tests and errors for the following cases:
- [ ]  no public_inputs section - this error & test should probably happen in the `ir`
- [x] public_inputs section is empty - this error & test can happen in the `parser`

grjte:
The second item was addressed in #32

tohrnii:
Closed by #32 and #52.

PR #51, opened 2022-10-28
Throw error when sections are empty or omitted

tohrnii:
Addressing #45, #44 and #30.
- Regarding throwing error when boundary_constraints or transition_constraints are empty, that case was already handled but I have included tests in parser for the same.
- Changes to the docs for transition constraints and boundary constraints are missing in this PR as they should be done after #36 is merged.


tohrnii:
Superseded by #52.

Issue #29, opened 2022-10-25
Throw error when `trace_columns` do not contain `main` declaration


grjte:
Currently it's possible to declare a trace columns source section without declaring the shape of the main execution trace. This shouldn't be possible. A new test should be added when it is fixed.

This is wrong but currently transpiles without error
```
trace_columns:
    aux: [a, b]
```

grjte:
We should also throw an error when the trace_columns source section is not defined. I don't think this case is currently handled. It probably belongs in the `ir`

tohrnii:
Closed by #52.

PR #52, opened 2022-10-28
Throw error when sections are empty or omitted

tohrnii:
Addressing #45, #44 and #30 and #29.
- Regarding throwing error when boundary_constraints or transition_constraints are empty, that case was already handled but I have included tests in parser for the same.
- Changes to the docs for transition constraints and boundary constraints are missing in this PR as they should be done after #36 is merged.

### DIFF
--- a/ir/src/error.rs
+++ b/ir/src/error.rs
@@ -6,5 +6,5 @@ pub enum SemanticError {
     IndexOutOfRange(String),
     TooManyConstraints(String),
     InvalidPeriodicColumn(String),
-    MissingSourceSection(String),
+    MissingDeclaration(String),
 }
--- a/ir/src/error.rs
+++ b/ir/src/error.rs
@@ -6,4 +6,5 @@ pub enum SemanticError {
     IndexOutOfRange(String),
     TooManyConstraints(String),
     InvalidPeriodicColumn(String),
+    MissingSourceSection(String),
 }
--- a/ir/src/helpers.rs
+++ b/ir/src/helpers.rs
@@ -1,0 +1,67 @@
+use crate::error::SemanticError;
+
+/// Struct to store existence of sections in the AIR source code
+pub(super) struct SourceValidator {
+    trace_columns_exists: bool,
+    public_inputs_exists: bool,
+    boundary_constraints_exists: bool,
+    transition_constraints_exists: bool,
+}
+
+impl SourceValidator {
+    pub fn new() -> Self {
+        SourceValidator {
+            trace_columns_exists: false,
+            public_inputs_exists: false,
+            boundary_constraints_exists: false,
+            transition_constraints_exists: false,
+        }
+    }
+
+    /// If the section exists, sets the provided section boolean to true.
+    pub fn exists(&mut self, section: &str) {
+        match section {
+            "trace_columns" => self.trace_columns_exists = true,
+            "public_inputs" => self.public_inputs_exists = true,
+            "boundary_constraints" => self.boundary_constraints_exists = true,
+            "transition_constraints" => self.transition_constraints_exists = true,
+            _ => unreachable!(),
+        }
+    }
+
+    /// Returns a MissingSourceSection error if the provided section is missing.
+    pub fn check(&self, section: &str) -> Result<(), SemanticError> {
+        match section {
+            "trace_columns" => {
+                if !self.trace_columns_exists {
+                    return Err(SemanticError::MissingSourceSection(
+                        "trace_columns section is missing".to_string(),
+                    ));
+                }
+            }
+            "public_inputs" => {
+                if !self.public_inputs_exists {
+                    return Err(SemanticError::MissingSourceSection(
+                        "public_inputs section is missing".to_string(),
+                    ));
+                }
+            }
+            "boundary_constraints" => {
+                if !self.boundary_constraints_exists {
+                    return Err(SemanticError::MissingSourceSection(
+                        "boundary_constraints section is missing".to_string(),
+                    ));
+                }
+            }
+            "transition_constraints" => {
+                if !self.transition_constraints_exists {
+                    return Err(SemanticError::MissingSourceSection(
+                        "transition_constraints section is missing".to_string(),
+                    ));
+                }
+            }
+            _ => unreachable!(),
+        }
+        Ok(())
+    }
+}
--- a/ir/src/helpers.rs
+++ b/ir/src/helpers.rs
@@ -1,6 +1,6 @@
 use crate::error::SemanticError;
 
-/// Struct to store existence of sections in the AIR source code
+/// Struct to help with validation of AIR source.
 pub(super) struct SourceValidator {
     trace_columns_exists: bool,
     public_inputs_exists: bool,
@@ -18,7 +18,7 @@ impl SourceValidator {
         }
     }
 
-    /// If the section exists, sets the provided section boolean to true.
+    /// If the declaration exists, sets corresponding boolean flag to true.
     pub fn exists(&mut self, section: &str) {
         match section {
             "trace_columns" => self.trace_columns_exists = true,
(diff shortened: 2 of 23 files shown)


ir/src/error.rs
```diff
@@ -6,4 +6,6 @@ pub enum SemanticError {
     IndexOutOfRange(String),
     TooManyConstraints(String),
     InvalidPeriodicColumn(String),
+    MissingSection(String),
```
grjte, line 9: nit: I would prefer `MissingSourceSection`

ir/src/lib.rs
```diff
@@ -36,12 +36,21 @@ impl AirIR {
     /// Consumes the provided source and generates a matching AirIR.
     pub fn from_source(source: &ast::Source) -> Result<Self, SemanticError> {
         let ast::Source(source) = source;
+
+        // ensure that the source contains trace columns section
+        if !contains_trace_cols(source) {
+            return Err(SemanticError::MissingSection(
+                "Trace Columns section is missing".to_string(),
```
grjte, line 43: nit: I would prefer to use the syntax of the missing section, e.g. "'trace_columns' section is missing" or "Trace Columns section `trace_columns` is missing"

ir/src/lib.rs
```diff
@@ -168,6 +182,58 @@ impl AirIR {
...
+
+/// Returns true if the source contains trace columns section.
+fn contains_trace_cols(source: &Vec<ast::SourceSection>) -> bool {
+    for section in source {
+        if matches!(section, ast::SourceSection::TraceCols(_)) {
+            return true;
+        }
+    }
+    false
+}
```
grjte, line 235: Since we can get the information about whether we have trace columns or not for free (when we process them in the match statement inside one of the iterators in `from_source`), I don't think it's optimal to iterate over all the sections here. 

Instead of the current approach, I think it might be simpler to just declare a verification struct with some booleans for each of the sections that are required. They would be initialized to false, then we'd just update them when the section is processed during the declaration and constraint handling. Finally, after all the source sections are processed, we can report an error based on the values of these booleans.

For example, in `from_source` we could do something like this:

```
let validator = SourceValidator::new()

// process the declarations of identifiers first, using a single symbol table to enforce
// uniqueness.
for section in source {
            match section {
                ...
                ast::SourceSection::TraceCols(columns) => {
                    validator.exists("trace_columns");
                    ...
                }
                ...
           }
}

// then process the constraints & validate them against the symbol table.
...

validator.check();

...
```

We would call "validator.exists" for all required source sections. In the future, if we needed to handle other types of error processing, we could incorporate that into this easily.

ir/src/lib.rs
```diff
@@ -168,6 +182,58 @@ impl AirIR {
     }
 }
 
+// === HELPERS ====================================================================================
+
+/// Returns an error if a public inputs section is not defined.
+fn validate_public_inputs(public_inputs: &PublicInputs) -> Result<(), SemanticError> {
+    if public_inputs.is_empty() {
+        return Err(SemanticError::MissingSection(
+            "Public Inputs section is missing".to_string(),
```
grjte, line 191: nit: Similar comment to the one above for the `trace_columns`, and same thing for boundary constraints and transition constraints below

ir/src/symbol_table.rs
```diff
@@ -69,6 +69,11 @@ impl SymbolTable {
         &mut self,
         columns: &[Identifier],
     ) -> Result<(), SemanticError> {
+        if columns.is_empty() {
+            return Err(SemanticError::MissingMainTraceColumns(
+                "Main Trace Columns not defined".to_string(),
+            ));
+        }
```
grjte, line 76: I think this check should be handled in the parser instead. It should be easy to add it there, since there's already an error raised when both trace section definitions are missing

grjte (changes requested):
Thanks for addressing these! I think maybe we could simplify things a bit and give more room to handle future errors cleanly by introducing a struct. I left a suggestion inline along with a few other nits

ir/src/lib.rs
tohrnii: Fixed, thanks.

ir/src/lib.rs
tohrnii: Fixed, thanks.

ir/src/symbol_table.rs
tohrnii: Fixed, thanks. There is one other issue. The parser expects the main trace columns to be defined before aux trace columns. Should we allow different ordering? If not, maybe the error should be more descriptive. I've opened #55 for this.

grjte: @tohrnii thanks for noticing and addressing this. Note for the future - the resolution is for this ordering to be required and to make an adjustment to the docs, which is in #36

ir/src/lib.rs
```diff
@@ -75,18 +84,26 @@ impl AirIR {
...
         }
 
         let (public_inputs, periodic_columns) = symbol_table.into_declarations();
 
+        // validate sections
+        validator.check("trace_columns")?;
+        validator.check("public_inputs")?;
+        validator.check("boundary_constraints")?;
+        validator.check("transition_constraints")?;
+
```
grjte, line 106: I think we should handle this as one call `validator.check()` so that when we have other things that need to be processed and then validated in the future it can all be captured in the same call

ir/src/lib.rs
```diff
@@ -193,6 +213,30 @@ mod tests {
...
+        trace_columns:
+            main: [clk]
+        transition_constraints:
+            enf clk' = clk + 1";
+
+        let parsed = parse(source).expect("Parsing failed");
+        let result = AirIR::from_source(&parsed);
+        assert!(result.is_err());
+    }
+
```
grjte, line 239: Public inputs are missing here and in some of the other tests, so that may affect these tests. In the future we'll want to test exact errors like we do in Miden VM, but that can wait until we've revised our unit testing as part of v0.2

parser/src/parser/grammar.lalrpop
```diff
@@ -41,7 +41,9 @@ TraceCols: TraceCols = {
     {
         (Some(main_cols), Some(aux_cols)) => Ok(TraceCols { main_cols, aux_cols }),
         (Some(main_cols), None) => Ok(TraceCols { main_cols, aux_cols: vec![] }),
-        (None, Some(aux_cols)) => Ok(TraceCols { main_cols: vec![], aux_cols }),
+        (None, Some(_aux_cols)) => Err(ParseError::User {
+            error: Error::ParseError(MissingMainTraceCols("Main Trace Columns definition is required".to_string()))
+        }),
```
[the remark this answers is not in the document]
tohrnii: Fixed, thanks.

ir/src/helpers.rs
```diff
@@ -0,0 +1,67 @@
...
+            "boundary_constraints" => self.boundary_constraints_exists = true,
+            "transition_constraints" => self.transition_constraints_exists = true,
+            _ => unreachable!(),
+        }
+    }
+
+    /// Returns a MissingSourceSection error if the provided section is missing.
+    pub fn check(&self, section: &str) -> Result<(), SemanticError> {
+        match section {
+            "trace_columns" => {
```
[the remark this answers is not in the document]
tohrnii: Fixed, thanks.

ir/src/error.rs
```diff
@@ -6,4 +6,5 @@ pub enum SemanticError {
     IndexOutOfRange(String),
     TooManyConstraints(String),
     InvalidPeriodicColumn(String),
+    MissingSourceSection(String),
```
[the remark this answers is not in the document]
tohrnii: Fixed, thanks.

ir/src/lib.rs
tohrnii: Fixed, thanks.

grjte (approved):
LGTM thank you!

PR #57, opened 2022-11-03
Minor docs fix

tohrnii:
This PR makes 3 minor changes to the docs.
- Fixes a link
- Adds mdbook-linkcheck
- Adds clarification regarding declaring main trace columns before aux trace columns


grjte:
@tohrnii these changes are good, but since the PR requires a rebase and there's another pending docs PR that has more significant updates which also needed a rebase and minor adjustments based on recent fixes, I just addressed these minor changes there instead. I'll tag you for an optional review if you have time

PR #58, opened 2022-11-03
Add bitwise air constraints with currently supported syntax

tohrnii:
This PR adds a modified version of existing bitwise air constraints to work with currently supported syntax and adds an integration test to check for expected code generation.


air-script/tests/bitwise/bitwise.air
```diff
@@ -0,0 +1,56 @@
...
+
+    # Enforce that input is decomposed into valid bits
+    enf a0^2 - a0 = 0
+    enf b0^2 - b0 = 0
+    enf a1^2 - a1 = 0
+    enf b1^2 - b1 = 0
+    enf a2^2 - a2 = 0
+    enf b2^2 - b2 = 0
+    enf a3^2 - a3 = 0
+    enf b3^2 - b3 = 0
```
grjte, line 32: nit: I would prefer to group the `a` bits together and the `b` bits together. It's obviously not so significant in an example, but if we use this for our actual constraints it would be much clearer to do it that way, and we are likely to use this as a basis for our constraints in the future

air-script/tests/bitwise/bitwise.air
```diff
@@ -0,0 +1,56 @@
...
+
+    # Enforce that the values in the columns a and b are exactly equal to the aggregation of binary
+    # values contained in the individual bit columns.
```
grjte, line 35: nit: I would prefer if the comment specifies that these constraints are only applied in the first row. For example, the existing comment in our implementation is "The values in column a in the first row should be the aggregation of the decomposed bit columns a0..a3." (with an analogous comment for b)

air-script/tests/bitwise/bitwise.air
```diff
@@ -0,0 +1,56 @@
...
+    # Enforce that for each row except the last, the aggregated output value must equal the previous
+    # aggregated output value in the next row.
+    enf k1 * (z - zp') = 0
+
+    # Enforce that for all rows the value in the z column is computed by multiplying the previous
+    # output value (from the zp column in the current row) by 16 and then adding it to the bitwise
+    # operation applied to the row's set of bits of a and b. The entire constraint must also be
+    # multiplied by the operation selector flag to ensure it is only applied for the appropriate
+    # operation.
+    enf (1 - s) * (z - (zp * 16 + 2^0 * a0 * b0 + 2^1 * a1 * b1 + 2^2 * a2 * b2 + 2^3 * a3 * b3)) + s * (z - (zp * 16 + a0 + b0 - 2^0 * a0 * b0 + a1 + b1 - 2^1 * a1 * b1 + a2 + b2 - 2^2 * a2 * b2 + a3 + b3 - 2^3 * a3 * b3)) = 0
```
grjte, line 56: 1. The implementation of the `XOR` constraint doesn't look correct here.
2. It's a bit hard to follow what's going on, since the current syntax forces this all to be combined into one row. I think adding a bit more to the comment specifying that it's enforcing `AND` when `s = 0` and `XOR` when `s = 1` might help

tohrnii:
Thank you @grjte for the review. Made suggested changes and fixes.
> I'm not sure that it makes sense to have the duplication or to have unsupported syntax in our examples (although we can use a variation of that file for the next version)

I've removed the examples directory entirely as we have all those files in the integration tests now.

grjte (changes requested):
Thanks @tohrnii, all the bitwise work looks good now. I don't think we should delete all the examples though, since in the docs & CLI examples we mention that they can be used for testing transpilation with the CLI. Maybe it would be best to just restore them for this PR and then submit a separate PR that has just one example which could be the [example from the docs](https://0xpolygonmiden.github.io/air-script/description/example.html) that includes all the syntax. What do you think?

tohrnii:
> I don't think we should delete all the examples though, since in the docs & CLI examples we mention that they can be used for testing transpilation with the CLI. Maybe it would be best to just restore them for this PR and then submit a separate PR that has just one example which could be the example from the docs that includes all the syntax. What do you think?

Makes sense. I've opened #74 based on your suggestion.

air-script/tests/bitwise/bitwise.rs
```diff
@@ -0,0 +1,102 @@
...
+        let mut result = Vec::new();
+        result
+    }
+
+    fn evaluate_transition<E: FieldElement<BaseField = Felt>>(&self, frame: &EvaluationFrame<E>, periodic_values: &[E], result: &mut [E]) {
+        let current = frame.current();
+        let next = frame.next();
+        result[0] = (current[0]).exp(E::PositiveInteger::from(2_u64)) - (current[0]) - (E::from(0_u64));
+        result[1] = (periodic_values[1]) * (next[0] - (current[0])) - (E::from(0_u64));
+        result[2] = (current[3]).exp(E::PositiveInteger::from(2_u64)) - (current[3]) - (E::from(0_u64));
```
bobbinth, line 78: Comment for the future: exponentiations could be really expensive, especially for constant-time implementations (like we are currently using). So, whenever possible, we should replace them with multiplications or more specialized operations. For example, here, instead of doing `(current[0]).exp(E::PositiveInteger::from(2_u64))` we should be doing `(current[0] * current[0])` or `current[0].square()`.

grjte:
@tohrnii is this ready for me to review again?

grjte (approved):
Thanks @tohrnii, this looks good!

Issue #60, opened 2022-11-03
Add support for declaring named constants (scalars, vectors, matrices)


tohrnii:
We should add support for constants. The syntax was defined by @bobbinth in the original Constraints Description Language discussion [here](https://github.com/maticnetwork/miden/discussions/254#discussioncomment-3284984). Taken directly from @bobbinth's original post:

Constants could be of the following three types:
- scalars
- vectors
- matrices

The constants should be declared in a separate section.
 
Declaring constants could be done like so:
```
constants:
  a: 123 // a scalar constant
  b: [1, 2, 3] // a vector constant
  c: [
    [1, 2, 3],
    [4, 5, 6],
  ] // a matrix constant
```

Referring to vector constants inside expressions could be done with index notation:
```
let x = b[1]
let y = c[0][2]
```

Something like this could also be possible:
```
let x = c[1][1..3] // this sets x to [5, 6]
```

Other open suggestions by @grjte and @bobbinth to easily distinguish between trace columns and constants:
- Use CAPITALS for constant names. If we do this, we should decide if we want to show an error to the user or a warning to use correct convention.
- Another option is to prefix trace columns with a some special symbol.

This task involves making changes to the parser, IR and codegen.

- [ ] parser
- [ ] IR
- [ ] codegen

tohrnii:
Add validation for column number mismatch for rows to the IR. For eg, the following should throw an error:
```
A: [[1, 2, 3], [4, 5]]
```

tohrnii:
Closed by #63, #94 and #95.

PR #63, opened 2022-11-04
Add support for parsing of constants

tohrnii:
Partly addressing #60.

This PR adds support for constants to the parser. This doesn't yet produce valid IR for constants. That will be handled in 
#67.

Questions:
1. Should we enforce using CAPITALS for constant names? If yes, should we throw an error or show a warning to the user?
2. Should we have any special characters for accessing constant values to ensure we know what the element is while building the AST instead of knowing after cross referencing it in the symbol table while building IR?


bobbinth:
> Should we enforce using CAPITALS for constant names? If yes, should we throw an error or show a warning to the user?

I'm in favor of this. It is kind of consistent with Rust. But also, I don't feel strongly about this.

> Should we have any special characters for accessing constant values to ensure we know what the element is while building the AST instead of knowing after cross referencing it in the symbol table while building IR?

What would be an example of using special characters here? If it is prefixing every constant reference with a special character, I don't think we need it (especially if we use capital letters for constant names).

tohrnii:
> If it is prefixing every constant reference with a special character, I don't think we need it (especially if we use capital letters for constant names).

Agreed. I meant that as an option if we don't go with using CAPITALS. But agreed that this doesn't seem like a good solution. I have gone with using a generic enum type in this PR while building the AST (for eg `VecElem` to access any kind of vector element (constant, variable, public input, etc) instead of using specific enum type like `PubInput` for public input value).

parser/src/ast/boundary_constraints.rs
```diff
@@ -61,9 +61,14 @@ impl Display for Boundary {
...
-    /// index of the cell.
-    PubInput(Identifier, usize),
+    /// Represents any named constant or variable.
+    Var(Identifier),
+    /// Represents an element inside a constant or variable vector. The index is the index of
+    /// this value inside the vector.
+    VecElem(Identifier, usize),
+    /// Represents an element inside a constant or variable matrix. Indices idx_row and idx_col
+    /// are the indices of this value inside the matrix.
+    MatrixElem(Identifier, usize, usize),
```
grjte, line 71: nit: as mentioned before, I think we should create a new atomic type and use it both here and in the transition constraints (and for grouped columns in the future as well)

parser/src/parser/tests/constants.rs
```diff
@@ -0,0 +1,62 @@
...
+    ])]);
+    build_parse_test!(source).expect_ast(expected);
+}
+
+#[test]
+fn error_empty_constants_section() {
+    let source = "
+    constants:
+    ";
+    assert!(build_parse_test!(source).parse().is_err());
```
grjte, line 61: Can you add more tests for edge cases? What happens if the matrix rows/columns have different numbers of elements? What happens if a vector contains 1 elements and 1 vector? What happens with `a: [[1],]`? etc, etc.

parser/src/parser/tests/boundary_constraints.rs
```diff
@@ -83,7 +87,7 @@ fn boundary_constraint_with_pub_input() {
             boundary_constraints: vec![BoundaryConstraint::new(
                 Identifier("clk".to_string()),
                 Boundary::First,
-                BoundaryExpr::PubInput(Identifier("a".to_string()), 0),
+                BoundaryExpr::VecElem(Identifier("a".to_string()), 0),
```
grjte, line 90: the name of this test should be updated

parser/src/parser/grammar.lalrpop
```diff
@@ -58,6 +60,22 @@ AuxCols: Vec<Identifier> = {
     "aux" ":" "[" <aux_cols: CommaElems<Identifier>> "]" => aux_cols,
 }
 
+// CONSTANTS
+// ================================================================================================
+
+// Constants section is optional but cannot be empty. Atleast one constant is required.
```
grjte, line 66: nit: "At least" instead of "Atleast". Also, I would clarify this a bit by saying "If the section is declared, then at least one constant is required." I would also say "The constants section" instead of "Constants section"

parser/src/parser/grammar.lalrpop
```diff
@@ -119,7 +137,10 @@ BoundaryAtom: BoundaryExpr = {
     <lexpr: BoundaryAtom> "^" <num: Num_u64> => BoundaryExpr::Exp(Box::new(lexpr), num),
     "$rand" "[" <n: Num_u64> "]" => BoundaryExpr::Rand(n as usize),
     <n: Num_u64> => BoundaryExpr::Const(n),
-    <ident: Identifier> "[" <index: Num_u64> "]" => BoundaryExpr::PubInput(ident, index as usize)
+    <ident: Identifier> => BoundaryExpr::Var(ident),
+    <ident: Identifier> "[" <idx: Num_u64> "]" => BoundaryExpr::VecElem(ident, idx as usize),
+    <ident: Identifier> "[" <idx_row: Num_u64> "]" "[" <idx_col: Num_u64> "]" =>
+        BoundaryExpr::MatrixElem(ident, idx_row as usize, idx_col as usize)
```
grjte, line 143: I think we should break this out into something reusable that we can also use for transition constraints and in the future for column groupings.

I'd prefer something like a `VectorAccess` whose index is a number or another vector access. (I haven't thought about this a ton yet though so lmk if you want to discuss)

grjte:
I think there's a general question here about whether we should be using column-major or row-major form. I know Miden is column-major, but Winterfell isn't always, and many other projects may not be. Row-major is more common in general, and we are aiming for AirScript to have more general utility than just for Miden.

@bobbinth curious about your thoughts

parser/src/ast/transition_constraints.rs
```diff
@@ -33,6 +33,8 @@ impl TransitionConstraint {
 pub enum TransitionExpr {
     Const(u64),
     Var(Identifier),
+    VecElem(Identifier, usize),
+    MatrixElem(Identifier, usize, usize),
```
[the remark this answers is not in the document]
grjte: Tbh, I'm not 100% convinced by the naming here, since a vector access may not be an element - it might be another vector. Do you have other ideas for this? Lookup, access, index, ... (I'm not in love with anything I've come up with)

grjte (changes requested):
Thanks for your work @tohrnii! I've left several comments inline, mostly about ways to use more shared abstractions and reduce near-duplications.

One other comment, though - the tests are mostly for the perfect case, which makes them great as examples and sanity checks, but doesn't give us the level of confidence in the code that I think we'd like to have. Can you add more edge cases for all of them? I added examples in one case, but the other test modules should be updated too

grjte:
> > Should we enforce using CAPITALS for constant names? If yes, should we throw an error or show a warning to the user?
>
> I'm in favor of this. It is kind of consistent with Rust. But also, I don't feel strongly about this.

I think we should probably do this. It's the convention for Python too, which means it's the convention for both languages we're referencing syntactically.

(This actually makes me wonder if we should really have a separate source section for these. Maybe they should just be declared at the top of the file uppercase, like the convention in Python and Rust. Having the lowercase source section followed by the uppercase constants feels weird.)

> > Should we have any special characters for accessing constant values to ensure we know what the element is while building the AST instead of knowing after cross referencing it in the symbol table while building IR?
>
> What would be an example of using special characters here? If it is prefixing every constant reference with a special character, I don't think we need it (especially if we use capital letters for constant names).

No, I don't think we should have special characters for accessing constant values.

parser/src/ast/transition_constraints.rs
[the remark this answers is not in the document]
Al-Kindi-0: I think "access" is the best of these, but I agree that it is not satisfying.

tohrnii:
> (This actually makes me wonder if we should really have a separate source section for these. Maybe they should just be declared at the top of the file uppercase, like the convention in Python and Rust. Having the lowercase source section "constants:" followed by the uppercase constants feels weird.)

@grjte I agree with your point however since currently all of our code is organized into sections, I think it might also look a little weird to only have constants declared outside a section. Maybe we should only do this if we are planning to have some other part of the code outside of the current source section structure as well. What do you think?

parser/src/ast/constants.rs
```diff
@@ -0,0 +1,32 @@
+// CONSTANTS
+// ================================================================================================
+
+use super::Identifier;
+
+/// Stores a constant's name and value. There are three types of constants:
+/// - Scalar: 123
+/// - Vector: \[1, 2, 3\]
+/// - Matrix: \[\[1, 2, 3\], \[4, 5, 6\]\]
```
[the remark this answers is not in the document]
tohrnii: Rustdoc doesn't seem to care about the escapes and outputs correct text with or without backslashes however the documentation on hovering on the struct in visual studio considers it a link without the escapes. Maybe there's a trick to fix this problem without escaping for visual studio but I couldn't find it.

parser/src/ast/transition_constraints.rs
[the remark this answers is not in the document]
tohrnii: I'm not sure if it's any better but other options probably are entity or item?

parser/src/parser/tests/boundary_constraints.rs
tohrnii: I went with adding a `public_inputs` section instead and kept the name same.

parser/src/parser/tests/constants.rs
bobbinth: I would say we should probably throw an error for matrixes have different number of elements in a row.

tohrnii:

I've also added enforcing the use of capitalized names for constants in this PR. The error message and corresponding enum variant can be improved though but I can't think of a better one.

parser/src/parser/tests/constants.rs
grjte: I agree with Bobbin, and also with you that we would do this at the IR level

parser/src/ast/constants.rs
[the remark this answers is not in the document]
grjte: I see, no that's fine. Thanks for explaining

parser/src/parser/grammar.lalrpop
```diff
@@ -119,7 +158,11 @@ BoundaryAtom: BoundaryExpr = {
     <lexpr: BoundaryAtom> "^" <num: Num_u64> => BoundaryExpr::Exp(Box::new(lexpr), num),
     "$rand" "[" <n: Num_u64> "]" => BoundaryExpr::Rand(n as usize),
     <n: Num_u64> => BoundaryExpr::Const(n),
-    <ident: Identifier> "[" <index: Num_u64> "]" => BoundaryExpr::PubInput(ident, index as usize)
+    <ident: Identifier> => BoundaryExpr::Elem(ident),
+    <ident: Identifier> "[" <idx: Num_u64> "]" => 
+        BoundaryExpr::VecElem(VectorAccess::new(ident, idx)),
+    <ident: Identifier> "[" <row_idx: Num_u64> "]" "[" <col_idx: Num_u64> "]" =>
+        BoundaryExpr::MatrixElem(MatrixAccess::new(ident, row_idx, col_idx))
```
grjte, line 165: Since most of this is reused for the transition expressions, I would pull most of it out of here so it's all in one place. Here's how I would approach this:

First, create 2 new atoms for indexing and size. These will be useful for column groupings as well. We can immediately each time we index to use `<idx: Index>` instead (e.g. indexing into `rand`). We can update the declaration of public inputs to use `Size`.
```
Size: u64 = {
    "[" <size: Num_u64> "]" => size
}

Index: usize = {
    "[" <idx: Num_u64> "]" => idx as usize
}
```

Next, I'd create rules for accessing vectors and matrices.
```
VectorAccess: VectorAccess = {
    <ident: Identifier> <idx: Index> => VectorAccess::new(ident, idx)
}

MatrixAccess: MatrixAccess = {
    <ident: Identifier> <row: Index> <col: Index> => MatrixAccess::new(ident, row, col)
}
```

Finally, I'd replace the boundary and transition constraint parsing rules to use `VectorAccess` and `MatrixAccess`. This will also make those rules easier to read.

```
BoundaryAtom: BoundaryExpr = {
    "(" <BoundaryExpr> ")",
    <lexpr: BoundaryAtom> "^" <num: Num_u64> => BoundaryExpr::Exp(Box::new(lexpr), num),
    "$rand" <n: Index> => BoundaryExpr::Rand(n as usize),
    <n: Num_u64> => BoundaryExpr::Const(n),
    <ident: Identifier> => BoundaryExpr::Elem(ident),
    <vector_access: VectorAccess> => BoundaryExpr::VecElem(vector_access),
    <matrix_access: MatrixAccess> => BoundaryExpr::MatrixElem(matrix_access)
}
```

parser/src/parser/grammar.lalrpop
grjte: I've left a code sample in the latest review

grjte (changes requested):
Thanks @tohrnii! This is a partial review of everything except the tests. I'll review the test changes separately.

parser/src/parser/grammar.lalrpop
tohrnii: Thank you. Fixed.

grjte:
> @grjte I've changed `VectorAccess` to `VecAccess` and `Matrix` to `MatAccess` but haven't changed it for variants of `ConstantType` since then `Scalar` will probably need to be shortened as well. Please let me know if this change should be reverted for uniformity or if `MatAccess` doesn't look good.

Hmm yeah, I probably would revert for uniformity personally. Matrix is hard to abbreviate well

grjte (changes requested):
Thanks @tohrnii! Just a few small final things inline

parser/src/parser/grammar.lalrpop
```diff
@@ -166,6 +212,32 @@ CommaElems<T>: Vec<T> = {
...
+    }
+}
+
+Size: u64 = {
+    "[" <size: Num_u64> "]" => size
+}
+
+Index: usize = {
+    "[" <idx: Num_u64> "]" => idx as usize
+}
```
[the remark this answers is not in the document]
tohrnii: Sorry, fixed.

parser/src/parser/tests/constants.rs
```diff
@@ -0,0 +1,151 @@
...
+            Identifier("A".to_string()),
+            ConstantType::Vector(vec![1, 2, 3, 4]),
+        ),
+        Constant::new(
+            Identifier("B".to_string()),
+            ConstantType::Vector(vec![5, 6, 7, 8]),
+        ),
+    ])]);
+    build_parse_test!(source).expect_ast(expected);
+}
```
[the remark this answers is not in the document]
tohrnii: Yeah that was a mistake. I probably copied it to add a failure test. Removed.

tohrnii:
> Hmm yeah, I probably would revert for uniformity personally. Matrix is hard to abbreviate well
Done.

Overcastan (approved):
Looks good! Thank you!

parser/src/lexer/tests/constants.rs
```diff
@@ -0,0 +1,171 @@
...
+        Token::Comma,
+        Token::Num("8".to_string()),
+        Token::Rsqb,
+        Token::Rsqb,
+    ];
+    expect_valid_tokenization(source, tokens);
+}
+
+#[test]
+fn constants_access_inside_boundary_expr() {
```
Al-Kindi-0, line 104: Do we have some failing tests?

Al-Kindi-0 (approved):
Looks great, thank you! I left only a question on testing.

parser/src/lexer/tests/constants.rs
tohrnii: Thank you. I added a few tests with comments that should fail at the parser or IR level. I don't think there's anything specific to constants that would fail at the lexer level (please let me know if you think there are failure cases to add here). There are a lot of redundant tests for lexer and we should refactor these in the future and probably remove most of them.

parser/src/ast/mod.rs
```diff
@@ -60,8 +64,74 @@ pub struct TraceCols {
...
+        Self {
+            name,
+            row_idx,
+            col_idx,
+        }
+    }
+
+    /// Returns the name of the matrix.
+    pub fn name(&self) -> &str {
+        &self.name.0
```
grjte, line 125: This should also be `&self.name.name()` now that we have it

grjte (approved):
Looks good! There's one final thing that should be adjusted, but I'm approving so I don't block things. Please make that adjustment before merging. Thanks for all your work on this!

parser/src/ast/mod.rs
tohrnii: Ah sorry, I thought I changed both instances. Fixed.

PR #70, opened 2022-11-10
Update Cargo.toml files for v0.1 release

bobbinth:
This PR adds a bunch of fields to Cargo.toml files which are needed to have for publishing to crates.io. It also alphabetizes all dependencies.

### DIFF
--- a/air-script/Cargo.toml
+++ b/air-script/Cargo.toml
@@ -1,19 +1,28 @@
 [package]
 name = "air-script"
 version = "0.1.0"
+description="AirScript language compiler"
+authors = ["miden contributors"]
+readme="README.md"
+license = "MIT"
+repository = "https://github.com/0xPolygonMiden/air-script"
+documnetation = "https://0xpolygonmiden.github.io/air-script/"
+categories = ["compilers", "cryptography"]
+keywords = ["air", "stark", "zero knowledge", "zkp"]
 edition = "2021"
+rust-version = "1.65"
 
 [[bin]]
 name = "airc"
 path = "src/main.rs"
 
 [dependencies]
-parser = { package = "air-parser", path = "../parser", version = "0.1.0" }
-ir = { package = "air-ir", path = "../ir", version = "0.1.0" }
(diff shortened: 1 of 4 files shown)


codegen/winterfell/Cargo.toml
```diff
@@ -1,7 +1,15 @@
 [package]
 name = "air-codegen-winter"
 version = "0.1.0"
+description="Winterfell code genreator for the AirScript languge"
```
grjte, line 4: nit: There's are a couple of typo. It should be "generator" instead of "genreator" and "language" instead of "languge"

ir/Cargo.toml
```diff
@@ -1,7 +1,15 @@
 [package]
 name = "air-ir"
 version = "0.1.0"
+description="Intermediate represenation for the AirScript language"
```
grjte, line 4: nit: There's a typo here. It should say "representation" instead of "represenation"

air-script/Cargo.toml
```diff
@@ -1,19 +1,28 @@
...
 name = "air-script"
 version = "0.1.0"
+description="AirScript language compiler"
+authors = ["miden contributors"]
+readme="README.md"
+license = "MIT"
+repository = "https://github.com/0xPolygonMiden/air-script"
+documnetation = "https://0xpolygonmiden.github.io/air-script/"
+categories = ["compilers", "cryptography"]
+keywords = ["air", "stark", "zkp"]
```
grjte, line 11: should we also have "zero knowledge" or "zero knowledge proofs" instead of just "zkp"?

grjte (approved):
Thanks for updating these! There are a few typos that should be fixed. Also, I wonder about the keywords & if we should include things that are more searchable (e.g. zero knowledge instead of just zkp or even zk in addition to zkp), but it's not a blocker.

codegen/winterfell/Cargo.toml
bobbinth: Sorry for all the typos! (apparently, my spell-checker doesn't work in `.toml` files). Fixed!

PR #71, opened 2022-11-10
Update keywords in Cargo.toml files

bobbinth:
Seems like we can't have spaces in keywords. So, updating `zero knowledge` to `zero-knowledge`.

Also, added badges to the main README file.


grjte (approved):
Looks good!

PR #74, opened 2022-11-15
Remove duplicate examples and only keep one example from docs in the examples directory

tohrnii:
This PR removes duplicate airscript files from examples directory and adds one example from the docs to the examples directory. Also includes a tiny fix for a rendering issue in the docs.


grjte (changes requested):
@tohrnii this looks great, and thanks for catching the rendering error! 

I think the only thing that's missing is that all examples related to the CLI need to be updated to use `example.air` instead of `system.air`.

I think this happens in the docs and in some of the READMEs, so the best thing is probably to search the entire repo for `system.air`

grjte:
minor nit: I just noticed that the commit message is longer than one line. In this case, it would be great if you can shorten it so it fits in the summary line, but then add more detailed bullet points below it, e.g.

```
chore: update AirScript examples

- remove examples that are included in integration tests
- add a file with the example from the docs
```

tohrnii:

grjte (approved):
LGTM thank you!

Issue #78, opened 2022-11-22
Formatting issue


Al-Kindi-0:
Formatting issue in:

1. Syntax overview/ built-in variables
2. Syntax overview/ Delimiters and special characters
3. AirScript example

grjte:
Closed by #81

PR #81, opened 2022-11-23
Fix doc rendering issue

tohrnii:
Addressing #78.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -3,7 +3,7 @@ name: AirScript book
 on:
   push:
     branches:
-      - main
+      - next
 
 jobs:
   deploy:
--- a/docs/src/description/syntax.md
+++ b/docs/src/description/syntax.md
@@ -6,7 +6,7 @@ This page specifies the basic syntax, types, and keywords of AirScript.
 
 AirScript defines the following keywords:
 
-- `$rand`: used to access random values provided by the verifier.
+- `\$rand`: used to access random values provided by the verifier.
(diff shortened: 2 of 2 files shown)


grjte (changes requested):
This fixes look good to me! However, the mdbook only deploys from main. We should probably do one of the following:

1. keep docs in sync with next, and change the CI `book.yml` workflow to deploy when the `next` branch is updated
2. merge this to `main` instead and then rebase `next`

I think this project is young enough and fast-moving enough that I'd probably go for the first option. What do you think?

tohrnii:
@grjte Agreed! I made the change to `book.yml`.

Al-Kindi-0:
Thank you @tohrnii ! 
A couple of questions: 

1. Is \$rand the correct syntax or is it $rand and \$ is just for .md formatting?
2. Does this address the issue in the example in section 2.5?

tohrnii:
@Al-Kindi-0 To answer your questions:
1. It's just for .md formatting. The correct syntax in AirScript is without the back slash.
2. The fix for that was included in https://github.com/0xPolygonMiden/air-script/pull/74 however it may not be getting reflected right now since the change was made to `next` branch but the docs are being pulled from `main` branch. That has been fixed in this PR so should be fixed after this PR is merged.

Al-Kindi-0:
Thank you!
In line 91, should that be "last row" instead?

tohrnii:
@Al-Kindi-0 Thank you. Fixed.

Al-Kindi-0 (approved):

grjte (approved):
Looks good to me, thanks!

PR #91, opened 2022-12-02
fix codegen for auxiliary constraint degrees

grjte:
This PR fixes the declaration of the constraint degrees for constraints in the auxiliary trace, which was previously declaring empty vector in all cases.

Issue #54, opened 2022-10-30
Consider making trace reference more general in the IR


bobbinth:
Currently, in the [Operation](https://github.com/0xPolygonMiden/air-dsl/blob/main/ir/src/transition_constraints/graph.rs#L245) enum the way we reference trace columns is via `MainTraceCurrentRow` and `MainTraceNextRow` (and similar for auxiliary trace). This hard-codes the evaluation frame to work only with two consecutive rows. Instead, we could make the structure more general by putting the row offset into the enum itself. For example, the enum could look like this:

```Rust
pub enum Operation {
    Const(u64),
    MainTraceRow(usize, usize),
    AuxTraceRow(usize, usize),
    ...
}
```
In the `MainTraceRow` and `AuxTraceRow` the first `usize` would refer to the column offset and the second `usize` would refer to the row offset. So, for example: `MainTraceRow(1, 0)` would mean column 1 at the current row, but `MainTraceRow(1, 1)` would mean column 1 at the next row.

grjte:
I like this overall, and it seems like the priority is bumped up by #84.

One thought I have thought is that I don't love having the offset data as usizes where order matters. I wonder if maybe instead we want some kind of light struct that has the Trace row information, where we can also move the information about which trace it's in. This would also make things more flexible if we wanted more than just Main and Auxiliary traces in the future.

Here's a rough idea, though the names could be improved:

```
pub enum Operation {
    Const(u64),
    TraceElement(TraceAccess), // replaces `MainTraceRow` and `AuxTraceRow`
    ...
}


struct TraceAccess {
    trace: TraceType
    col_idx: usize,
    row_offset: usize,
}

pub enum TraceType {
    Main,
    Aux
}
```

grjte:
Another thought - to make it easy to expand beyond main and aux traces, we could identify the trace with a number indicating how many commitments came before the trace (or how many times we got fresh randomness from the verifier), e.g. our current `Main` would be trace `0`, our current `Aux` would be trace `1`, etc. Then at least the IR part of this supports more traces (although language syntax + parser would need adjustment for future changes)

Then we'd have something like:

```
struct TraceAccess {
    trace: u8,
    col_idx: usize,
    row_offset: usize,
}
```

bobbinth:
I like the latter approach as it does not tie us into the terminology of `main` and `aux` names.

The only suggestion is that maybe we follow the naming convention of winterfell and call these `trace_segment`. So, the struct would look like this:
[...]

PR #92, opened 2022-12-02
Make the trace reference more general in the IR

grjte:
This PR makes the trace reference more general in the IR, addressing issue #54. This PR depends on the fix in #91 and should not be merged first.

The IR has been generalized everywhere except:
- the insertion of random values into the graph. This can't change until we have a way to describe different sets of randomness in the AirScript syntax.
- the way we store boundary constraints. This needs more thinking and probably deserves its own issue

### DIFF
--- a/codegen/winterfell/src/air/mod.rs
+++ b/codegen/winterfell/src/air/mod.rs
@@ -99,7 +99,7 @@ fn add_fn_new(impl_ref: &mut Impl, ir: &AirIR) {
 
     // define the transition constraint degrees of the main trace `main_degrees`.
     let main_degrees = ir
-        .main_degrees()
+        .constraint_degrees(0)
         .iter()
         .map(|degree| degree.to_string(false))
         .collect::<Vec<_>>();
@@ -110,7 +110,7 @@ fn add_fn_new(impl_ref: &mut Impl, ir: &AirIR) {
 
     // define the transition constraint degrees of the aux trace `aux_degrees`.
     let aux_degrees = ir
-        .aux_degrees()
+        .constraint_degrees(1)
         .iter()
         .map(|degree| degree.to_string(true))
         .collect::<Vec<_>>();
--- a/codegen/winterfell/src/air/transition_constraints.rs
+++ b/codegen/winterfell/src/air/transition_constraints.rs
@@ -25,7 +25,7 @@ pub(super) fn add_fn_evaluate_transition(impl_ref: &mut Impl, ir: &AirIR) {
 
     // output the constraints.
     let graph = ir.transition_graph();
-    for (idx, constraint) in ir.main_transition_constraints().iter().enumerate() {
+    for (idx, constraint) in ir.transition_constraints(0).iter().enumerate() {
         evaluate_transition.line(format!(
             "result[{}] = {};",
             idx,
@@ -56,7 +56,7 @@ pub(super) fn add_fn_evaluate_aux_transition(impl_ref: &mut Impl, ir: &AirIR) {
 
     // output the constraints.
     let graph = ir.transition_graph();
-    for (idx, constraint) in ir.aux_transition_constraints().iter().enumerate() {
+    for (idx, constraint) in ir.transition_constraints(1).iter().enumerate() {
         evaluate_aux_transition.line(format!(
             "result[{}] = {};",
             idx,
@@ -85,12 +85,15 @@ impl Codegen for Operation {
     fn to_string(&self, graph: &AlgebraicGraph) -> String {
         match self {
             Operation::Const(value) => format!("E::from({}_u64)", value),
-            Operation::MainTraceCurrentRow(col_idx) | Operation::AuxTraceCurrentRow(col_idx) => {
-                format!("current[{}]", col_idx)
-            }
-            Operation::MainTraceNextRow(col_idx) | Operation::AuxTraceNextRow(col_idx) => {
-                format!("next[{}]", col_idx)
-            }
+            Operation::TraceElement(trace_access) => match trace_access.row_offset() {
+                0 => {
+                    format!("current[{}]", trace_access.col_idx())
+                }
+                1 => {
+                    format!("next[{}]", trace_access.col_idx())
+                }
+                _ => panic!("Winterfell doesn't support row offsets greater than 1."),
+            },
             Operation::PeriodicColumn(col_idx, _) => {
                 format!("periodic_values[{}]", col_idx)
             }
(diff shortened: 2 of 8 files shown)


codegen/winterfell/src/air/mod.rs
```diff
@@ -108,7 +108,14 @@ fn add_fn_new(impl_ref: &mut Impl, ir: &AirIR) {
     ));
 
     // define the transition constraint degrees of the aux trace `aux_degrees`.
-    new.line("let aux_degrees = Vec::new();");
+    let mut aux_degrees: Vec<String> = Vec::new();
+    for degree in ir.constraint_degrees(1).iter() {
+        aux_degrees.push(degree.to_string(false));
+    }
```
bobbinth, line 114: stylistic nit: this could be done in a more idiomatic way as follows:
```Rust
let aux_degrees = ir
    .constraint_degrees(1)
    .iter()
    .map(|degree| degree.to_string(false))
    .collect::<Vec<_>>();
```

ir/src/symbol_table.rs
```diff
@@ -20,16 +21,48 @@ impl Display for IdentifierType {
...
+        self.col_idx
+    }
+}
+
 /// SymbolTable for identifiers to track their types and information and enforce uniqueness of
 /// identifiers.
 #[derive(Default, Debug)]
 pub(super) struct SymbolTable {
+    /// The number of trace segments in the AIR.
+    num_trace_segments: TraceSegment,
```
bobbinth, line 64: nit: I found it a bit odd that the type of `num_trace_segments` is `TraceSegment`. I know it's just an alias for `u8`, but it looks strange because I think of `TraceSegment` as in index of trace segment. Maybe this should be just `u8` or `usize`?

ir/src/transition_constraints/mod.rs
```diff
@@ -15,57 +15,56 @@ pub const MIN_CYCLE_LENGTH: usize = 2;
...
-    // --- PUBLIC ACCESSORS -----------------------------------------------------------------------
+    // --- CONSTRUCTOR ----------------------------------------------------------------------------
 
-    /// Returns a vector of the degrees of the transition constraints.
-    pub fn main_degrees(&self) -> Vec<TransitionConstraintDegree> {
-        self.main_constraints
-            .iter()
-            .map(|entry_index| self.graph.degree(entry_index))
-            .collect()
+    pub fn new(num_trace_segments: TraceSegment) -> Self {
```
bobbinth, line 34: Same as the previous comment about type of `num_trace_segments` being `TraceSegment`. Probably `usize` is better.

bobbinth (approved):
Looks good! Thank you! I left a few non-blocking comments inline.

ir/src/symbol_table.rs
```diff
@@ -43,6 +76,12 @@ pub(super) struct SymbolTable {
 }
 
 impl SymbolTable {
+    /// Sets the number of trace segments, based on the current number of trace segments and the
+    /// provided trace segment identifier, which is indexed from zero.
```
tohrnii, line 80: minor nit: Maybe this can be rephrased a bit to clarify that num_trace_segments is set to max of the trace segment encountered till now but it's easy to look at the code and see what's happening at a glance so maybe not required.

tohrnii (approved):
Thank you @grjte! Looks great! I left a few very minor nits inline.

codegen/winterfell/src/air/mod.rs
```diff
@@ -99,7 +99,7 @@ fn add_fn_new(impl_ref: &mut Impl, ir: &AirIR) {
 
     // define the transition constraint degrees of the main trace `main_degrees`.
     let mut main_degrees: Vec<String> = Vec::new();
-    for degree in ir.main_degrees().iter() {
+    for degree in ir.constraint_degrees(0).iter() {
         main_degrees.push(degree.to_string(false));
     }
```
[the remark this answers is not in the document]
grjte: Good point, thanks!

codegen/winterfell/src/air/mod.rs
grjte: Fixed!

Al-Kindi-0 (approved):
Looks great, thank you @grjte !

codegen/winterfell/src/air/transition_constraints.rs
```diff
@@ -85,12 +85,16 @@ impl Codegen for Operation {
...
-            }
+            Operation::TraceElement(trace_access) => match trace_access.row_offset() {
+                0 => {
+                    format!("current[{}]", trace_access.col_idx())
+                }
+                1 => {
+                    format!("next[{}]", trace_access.col_idx())
+                }
+                // Winterfell doesn't support row offsets greater than 1.
+                _ => unimplemented!(),
```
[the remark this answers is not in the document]
grjte: I debated this a bit too. We don't currently have a `Result` type for our `to_string`, and I was thinking that `panic` makes sense in this case. `unimplemented` is just a wrapper around `panic` which also seemed logical, but a more specific message that what's missing is Winterfell support also makes sense. For now, I'll change it to `panic` with a more specific message.

ir/src/symbol_table.rs
grjte: I've attempted to clarify.

ir/src/boundary_constraints.rs
```diff
@@ -80,13 +81,16 @@ impl BoundaryConstraints {
...
+            IdentifierType::TraceColumn(column) => match column.trace_segment() {
+                0 => match constraint.boundary() {
+                    ast::Boundary::First => self.main_first.insert(column.col_idx(), expr),
+                    ast::Boundary::Last => self.main_last.insert(column.col_idx(), expr),
+                },
+                1 => match constraint.boundary() {
+                    ast::Boundary::First => self.aux_first.insert(column.col_idx(), expr),
+                    ast::Boundary::Last => self.aux_last.insert(column.col_idx(), expr),
+                },
+                _ => unimplemented!(),
```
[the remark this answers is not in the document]
grjte: I think `unimplemented` makes sense here, because the only way we should get something else is if the IR gets an AST with more trace segments, which is not currently possible given the syntax of AirScript. Therefore a SemanticError doesn't really make sense. Once we support the trace segments in the language, we'll need to implement this (and refactor the boundary contraints handling in the IR)

codegen/winterfell/src/air/transition_constraints.rs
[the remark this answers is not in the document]
grjte: (Also, we're planning to add support for more trace segments to Winterfell at some point, which is probably the other reason I originally went with `unimplemented`)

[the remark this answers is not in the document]
tohrnii: Makes sense. Thanks.

PR #94, opened 2022-12-04
Add constants support to IR

tohrnii:
Partly addressing #60.


ir/src/transition_constraints/graph.rs
```diff
@@ -187,9 +195,69 @@ impl AlgebraicGraph {
...
+                vector_access.name(),
+                vector_access.idx(),
+                vector_access_type
+            ))),
+        }
+    }
+
+    /// Validates and adds a matrix access to the graph.
+    /// Returns an error if the matrix identifier's value is not of a supported type.
+    fn insert_matrix_access(
```
bobbinth, line 237: Similar comments here as for `insert_vector_access()`.

ir/src/transition_constraints/graph.rs
```diff
@@ -165,16 +176,13 @@ impl AlgebraicGraph {
         }
     }
 
-    fn insert_variable(
+    fn insert_elem(
```
bobbinth, line 179: Should this maybe be `insert_symbol()` or maybe `insert_symbol_access()` to be more consistent with `insert_vector_access()` and `insert_matrix_access()`?

bobbinth (changes requested):
Looks good! Thank you! I left a few comments inline.

ir/src/transition_constraints/graph.rs
tohrnii: Changed it to `insert_symbol_access()`.

ir/src/transition_constraints/graph.rs
```diff
@@ -232,6 +300,9 @@ impl Node {
 #[derive(Debug, Eq, PartialEq)]
 pub enum Operation {
     Const(u64),
+    ScalarConstant(u64),
+    VectorConstant(VectorAccess),
+    MatrixConstant(MatrixAccess),
```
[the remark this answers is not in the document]
tohrnii: Thanks. I made a slight variation and went with using `VectorAccess` and `MatrixAccess` structs as arguments.
```
pub enum ConstantValue {
    Inline(u64),
    Scalar(String),
    Vector(VectorAccess),
    Matrix(MatrixAccess),
}
```

@grjte had a slight preference of using structs here instead of fixing index order. Please let me know if you prefer to use struct fields as arguments instead or if there are any performance implications in using struct here (I couldn't find anything on that).

ir/src/transition_constraints/graph.rs
```diff
@@ -187,9 +195,69 @@ impl AlgebraicGraph {
...
+    /// Validates and adds a vector access to the graph.
+    /// Returns an error if the vector identifier's value is not of a supported type.'
+    fn insert_vector_access(
+        &mut self,
+        symbol_table: &SymbolTable,
+        vector_access: &VectorAccess,
+    ) -> Result<(TraceSegment, NodeIndex), SemanticError> {
+        // check that vector access is valid
+        symbol_table.validate_vector_access(vector_access.name(), vector_access.idx())?;
+        let vector_access_type = symbol_table.get_type(vector_access.name())?;
```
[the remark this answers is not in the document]
tohrnii: Thanks. I changed it to pass `access` struct directly but I think keeping it as `validate_vector_access()` seems better. I moved the symbol table lookup above the validation check and pass the `IdentifierType` to the `validate_vector_access()` method so that we have to do the lookup only once. Using `access_vector_element()` in cases where we ignore the value was looking a little weird. Please let me know if you think otherwise and would prefer `access_vector_element()` instead.

ir/src/transition_constraints/graph.rs
```diff
@@ -187,9 +195,69 @@ impl AlgebraicGraph {
...
+                let trace_segment = 0;
+                let node_index = self.insert_op(Operation::VectorConstant(vector_access.clone()));
+                Ok((trace_segment, node_index))
+            }
+            _ => Err(SemanticError::InvalidUsage(format!(
+                "Vector Access {}[{}] was declared as a {} which is not a supported type.",
+                vector_access.name(),
+                vector_access.idx(),
+                vector_access_type
+            ))),
```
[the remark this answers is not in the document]
tohrnii: Thanks. I made the change for some errors but we should probably do this for other semantic errors as well? If yes, I'll open an issue for it.

ir/src/transition_constraints/graph.rs
tohrnii: Same as above. The signature now looks like:
```
pub(super) fn validate_matrix_access(
        &self,
        matrix_access: &MatrixAccess,
        matrix_access_type: &IdentifierType,
    ) -> Result<(), SemanticError>
```

ir/src/tests/mod.rs
```diff
@@ -0,0 +1,462 @@
+// TODO: add checks for the correctness of the AirIR that is built.
+#[cfg(test)]
+mod tests {
```
bobbinth, line 3: since we are already inside `tests` module, you don't need another `tests` module here.

ir/src/transition_constraints/graph.rs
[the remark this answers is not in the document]
bobbinth: This could work but then I don't think `SymbolTable` is the right place for these functions since we are not using `self` inside these functions at all. If you want to keep it this way, I'd move the function somewhere else - e.g., either as a helper function in `graph.rs` or maybe as method on `VectorAccess` and `MatrixAccess`.

ir/src/tests/mod.rs
tohrnii: Ah, sorry, fixed.

ir/src/transition_constraints/graph.rs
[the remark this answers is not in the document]
bobbinth: I think this works. In general, we are much less concerned about performance here - so, w/e makes the code more clear is usually better.

It does feel a bit weird that we use structs from the parser in the IR, but I guess we do it in a few other places already. Maybe there should be a `air-script-core` or `air-script-common` crate which defines such commonly used structs.

ir/src/transition_constraints/graph.rs
```diff
@@ -231,7 +299,7 @@ impl Node {
 /// A transition constraint operation or value reference.
 #[derive(Debug, Eq, PartialEq)]
 pub enum Operation {
-    Const(u64),
+    Constant(ConstantValue),
```
bobbinth, line 302: I would add doc comment for `Constant` in the same way we do for other variants here.

ir/src/transition_constraints/graph.rs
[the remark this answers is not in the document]
bobbinth: I'd probably add a constructor for error here and for the one inside `insert_matrix_access()` and then make a separate PR for the rest.

ir/src/transition_constraints/graph.rs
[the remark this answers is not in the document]
tohrnii: Makes sense. I've opened #98 for this.

ir/src/boundary_constraints.rs
```diff
@@ -126,7 +126,12 @@ fn validate_expression(
 ) -> Result<(), SemanticError> {
     match expr {
         BoundaryExpr::VectorAccess(vector_access) => {
-            symbol_table.validate_public_input(vector_access.name(), vector_access.idx())
+            let vector_access_type = symbol_table.get_type(vector_access.name())?;
+            symbol_table.validate_vector_access(vector_access, vector_access_type)
```
bobbinth, line 130: I guess since `validate_vector_access()` it can't really go into `graph.rs` - but even here it feels weird that we first get the type and then ask `SymbolTable` to validate it, even though there is no `SymbolTable`-specific logic in validation code.

I think I'd probably still prefer using `access_vector_element()` and `access_matrix_element()` as it encapsulates the logic more cleanly.

bobbinth (changes requested):
Thank you! I left a couple more comments inline.

ir/src/transition_constraints/graph.rs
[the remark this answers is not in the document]
tohrnii: Would it make sense to keep the functions as public helper functions in `symbol_table.rs` file itself since we use them in `graph.rs` and `boundary_constraints.rs` and we already have other helper validation functions in `symbol_table.rs`?

ir/src/boundary_constraints.rs
tohrnii: Got it. Will change it to that.

Al-Kindi-0 (approved):
Looks good to me, @tohrnii , thank you!

PR #96, opened 2022-12-05
Make the trace reference more general in the IR

tohrnii:
#92 was merged to this branch rather than `next` by mistake. This PR contains the work done in #92.


bobbinth (approved):
All looks good!

Issue #61, opened 2022-11-03
Add support for intermediate variables (scalars, vectors, matrices)


tohrnii:
We should add support for variables. The syntax was defined by @bobbinth in the original Constraints Description Language discussion [here](https://github.com/maticnetwork/miden/discussions/254#discussioncomment-3284984). Taken from @bobbinth's original post:

A variable is defined using a `let` keyword. For e.g., `let a = b * c`

Variables could be of the following three types:
- scalars
- vectors
- matrices

Declaring variables could be done like so:
```
trace_columns:
  main: [a, b, c]
transition_constraints:
  let x = a + 123 // a scalar variable
  let y = [a + 1, b + 2, c + 3] // a vector variable
  let z = [
    [a + 1, b + 2, c + 3],
    [a + 4, b + 5, c + 6],
  ] // a matrix variable
```

Referring to vector variables inside expressions could be done with index notation:
```
let a = [m, n]
let x = a[0] + 1
```

We could build variables from column references as well:
```
trace_columns:
  main: [a, b]

let x = [a, b]   // x is a vector with current values from a and b
let y = [a', b'] // y is a vector with next values from a and b
```

Variables should only be defined in `boundary_constraints` and `transition_constraints` sections.

This task involves making changes to the parser, IR and codegen.

- [ ] parser
- [ ] IR
- [ ] codegen

tohrnii:
Variable support is remaining for boundary constraints. The issue has been moved to #106.

Closed by #100.

PR #100, opened 2022-12-08
Variables IR and Codegen

tohrnii:
Partly addressing #61.


ir/src/transition_stmts/graph.rs
```diff
@@ -296,6 +351,8 @@ impl Node {
 pub enum Operation {
     /// An inlined or named constant with identifier and access indices.
     Constant(ConstantValue),
+    /// A variable with [VariableValue] and root index of the variable's expression value.
+    Variable(VariableValue, NodeIndex),
```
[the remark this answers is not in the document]
grjte: I agree. Let's keep them out of the graph

ir/src/tests/mod.rs
```diff
@@ -228,6 +228,32 @@ fn transition_constraints_with_constants() {
...
+        let b = [a, a*a]
+        let c = [[clk' - clk, clk - a], [1 + 8, 2^2]]
+        enf c[0][0] = 1
+    boundary_constraints:
+        enf clk.first = A
+        enf clk.last = B[0] + C[0][1]";
+
+    let parsed = parse(source).expect("Parsing failed");
+    let result = AirIR::from_source(&parsed);
+    println!("{:?}", result);
```
Al-Kindi-0, line 253: Is this intentional?

ir/src/symbol_table.rs
```diff
@@ -279,6 +312,30 @@ impl SymbolTable {
...
+                        ));
+                    }
+                    if matrix_access.col_idx() >= matrix[0].len() {
+                        return Err(SemanticError::matrix_access_out_of_bounds(
+                            matrix_access,
+                            matrix.len(),
+                            matrix[0].len(),
+                        ));
+                    }
+                    Ok(symbol_type)
```
grjte, line 331: This whole section seems to be a duplication of the code in the `Constant` case above. Could we use a helper function instead?

ir/src/symbol_table.rs
```diff
@@ -241,6 +257,23 @@ impl SymbolTable {
...
+            IdentifierType::TransitionVariable(transition_variable) => {
+                if let TransitionVariableType::Vector(vector) = transition_variable.value() {
+                    if vector_access.idx() < vector.len() {
+                        Ok(symbol_type)
+                    } else {
+                        Err(SemanticError::vector_access_out_of_bounds(
+                            vector_access,
+                            vector.len(),
+                        ))
+                    }
```
grjte, line 269: This section seems to be a duplication of the code in the `Constant` case above. Could we use a helper function instead?

ir/src/boundary_stmts.rs
```diff
@@ -0,0 +1,175 @@
...
+
+/// A struct containing all of the boundary constraints to be applied at each of the 2 allowed
+/// boundaries (first row and last row). The constraints are stored in a vector of tuples where
+/// each index of the vector corresponds to the trace segment of the column the boundary constraint
+/// applies to. For ease of code generation and evaluation, constraints are sorted into maps by the
+/// boundary. This also simplifies ensuring that there are no conflicting constraints sharing a
+/// boundary and column index.
+#[derive(Default, Debug)]
+pub(crate) struct BoundaryStmts {
+    boundary_constraints: Vec<(BTreeMap<usize, BoundaryExpr>, BTreeMap<usize, BoundaryExpr>)>,
```
grjte, line 17: I think this approach makes sense for the trace segment handling. Wdyt about defining a type alias for `BTreeMap<usize, BoundaryExpr>`?

ir/src/boundary_stmts.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    // --- ACCESSORS ------------------------------------------------------------------------------
+
+    /// Returns the number of boundary constraints for the specified trace segment.
+    pub fn num_boundary_constraints(&self, trace_segment: TraceSegment) -> usize {
+        if self.boundary_constraints.len() <= trace_segment.into() {
+            0
+        } else {
+            self.boundary_constraints[trace_segment as usize].0.len()
+                + self.boundary_constraints[trace_segment as usize].1.len()
```
grjte, line 36: Maybe it would be clearer to use a struct instead of a tuple so we could refer to first and last instead of 0 and 1. Wdyt?

ir/src/boundary_stmts.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// Add a boundary statement from the AST. The statement can be either be a variable or a
+    /// constraint. In case it is a constraint, it is added to the boundary_constraints vector
+    /// in the relevant trace segment. Variables are currently not supported for boundary
+    /// constraints.
+    ///
+    /// # Errors
+    /// Returns an error if:
+    /// - The identifier specified for the boundary constraint column has not been declared or has
+    ///   been declared with the wrong type.
+    /// - The constraint expression is contains invalid public input references.
```
grjte, line 85: nit: extra word "is"

ir/src/boundary_stmts.rs
```diff
@@ -0,0 +1,175 @@
...
+                        },
+                        1 => match constraint.boundary() {
+                            ast::Boundary::First => self.boundary_constraints[1]
+                                .0
+                                .insert(column.col_idx(), expr),
+                            ast::Boundary::Last => self.boundary_constraints[1]
+                                .1
+                                .insert(column.col_idx(), expr),
+                        },
+                        _ => unimplemented!(),
```
grjte, line 119: With the new refactor making everything more generic we should be able to reduce redundancy and do something more elegant here.

ir/src/boundary_stmts.rs
```diff
@@ -0,0 +1,175 @@
...
+        Ok(())
+    }
+}
+
+/// Recursively validates the BoundaryExpression.
+///
+/// # Errors
+/// Returns an error if the expression includes a reference to a public input that hasn't been
+/// declared or to an invalid index in an existing public input.
+/// TODO: Complete implementation of validation.
```
grjte, line 151: What's missing in the implementation of the validation? Is there an issue tracking this?

ir/src/transition_stmts/graph.rs
```diff
@@ -87,8 +88,9 @@ impl AlgebraicGraph {
     /// added recursively to reuse existing matching nodes.
     pub(super) fn insert_expr(
         &mut self,
-        symbol_table: &SymbolTable,
+        symbol_table: &mut SymbolTable,
```
grjte, line 91: I don't think the symbol table needs to be mutable, does it? It doesn't seem to be changed anywhere inside this module

ir/src/transition_stmts/graph.rs
```diff
@@ -200,6 +205,28 @@ impl AlgebraicGraph {
...
+                        );
+                        Ok((trace_segment, node_index))
+                    }
+                } else {
+                    Err(SemanticError::InvalidUsage(format!(
+                        "Identifier {} was declared as a {} which is not a supported type.",
+                        ident, elem_type
+                    )))
+                }
+            }
```
grjte, line 229: There's quite a bit of similarity/near-redundancy between this, `insert_vector_access`, and `insert_matrix_access, so I think there's probably a better way. Can you take a look at refactoring a bit or ping me if you want to discuss?

grjte (changes requested):
Looks good, thanks @tohrnii! I like the improvement to the boundary constraint handling to make them more generic. I left a few comments inline

ir/src/transition_stmts/graph.rs
tohrnii: Thanks, Fixed. Forgot to change it after refactor.

ir/src/boundary_stmts.rs
tohrnii: I've defined an alias with the name `BoundaryConstraintsMap`. Please let me know if you have a better name suggestion.

ir/src/tests/mod.rs
tohrnii: No, it was a mistake. Fixed.

ir/src/boundary_stmts.rs
tohrnii: Thanks, please let me know if you had something else in mind.

ir/src/transition_stmts/graph.rs
tohrnii: I agree with you that there is some redundancy but I can't think of a way to reduce it much here.

ir/src/boundary_stmts.rs
tohrnii: That was for me to check before submitting this PR. I think all the cases should be covered now.

tohrnii:
Thank you @grjte and @Al-Kindi-0 for the review. Made suggested changes.

Al-Kindi-0 (approved):

ir/src/boundary_stmts.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// Returns the number of boundary constraints for the specified trace segment.
+    pub fn num_boundary_constraints(&self, trace_segment: TraceSegment) -> usize {
+        if self.boundary_constraints.len() <= trace_segment.into() {
+            0
+        } else {
+            self.boundary_constraints[trace_segment as usize]
+                .first
+                .len()
+                + self.boundary_constraints[trace_segment as usize].last.len()
+        }
```
grjte, line 72: nit: Since it's used 3 times, I would probably convert `trace_segment` to `usize` just once at the top, e.g. `let trace_segment = trace_segment as usize`

ir/src/boundary_stmts.rs
```diff
@@ -0,0 +1,201 @@
...
+    ///
+    /// # Errors
+    /// Returns an error if:
+    /// - The identifier specified for the boundary constraint column has not been declared or has
+    ///   been declared with the wrong type.
+    /// - The constraint expression contains invalid public input references.
+    /// - A boundary constraint has already been declared for the specified column and boundary.
+    pub(super) fn insert(
+        &mut self,
+        symbol_table: &mut SymbolTable,
```
grjte, line 124: nit: this shouldn't need to be mutable, since we're not supporting variables in boundary statements yet

ir/src/transition_stmts/graph.rs
```diff
@@ -174,10 +178,15 @@ impl AlgebraicGraph {
         }
     }
 
+    /// Adds a trace column, periodic column, named constant or a variable to the graph and returns
+    /// the node index and trace segment.
+    /// Returns an error if the identifier is not present in the symbol table or is not a supported
+    /// type.
```
grjte, line 184: nit: This part of the comment should be under an `# Errors` header

ir/src/transition_stmts/graph.rs
```diff
@@ -174,10 +178,15 @@ impl AlgebraicGraph {
...
     fn insert_symbol_access(
         &mut self,
         symbol_table: &SymbolTable,
         ident: &str,
+        variable_roots: &mut BTreeMap<VariableValue, (TraceSegment, NodeIndex)>,
```
grjte, line 189: nit: we should be able to use the new `VariableRoots` type here

ir/src/transition_stmts/graph.rs
grjte: Ok, let's leave it for now and revisit in the future once the tests are in place.

ir/src/transition_stmts/graph.rs
```diff
@@ -213,23 +244,46 @@ impl AlgebraicGraph {
         &mut self,
         symbol_table: &SymbolTable,
         vector_access: &VectorAccess,
+        variable_roots: &mut BTreeMap<VariableValue, (TraceSegment, NodeIndex)>,
```
grjte, line 247: nit: We should be able to use the new `VariableRoots` type here. Can you search the crate by `BTreeMap<VariableValue, (TraceSegment, NodeIndex)>` to make sure no other instances were missed?

grjte (changes requested):
Thanks @tohrnii! I have a few final comments, but they're all very small nits - should only take 5 min to update.

tohrnii:
Thanks @grjte. Fixed.

Overcastan (approved):

grjte (approved):
Looks good, thanks @tohrnii!

PR #101, opened 2022-12-10
Add support for trace col grouping

### LINKED ISSUE #83: Add support for grouping of trace columns to parser
We should allow users to name trace columns in a group as an array. For eg,
```
trace_columns:
    main: [x, y, a[4], z]
```

Here the main trace contains 7 columns where 4 are grouped under `a`. These columns can be accessed using their local indices inside `a` (for e.g., a[1] to refer to the second column in group `a`).

tohrnii:
Closes #83.

### DIFF
--- a/ir/src/symbol_table.rs
+++ b/ir/src/symbol_table.rs
@@ -4,7 +4,7 @@ use super::{
 };
 use parser::ast::{
     constants::{Constant, ConstantType},
-    Identifier, MatrixAccess, PeriodicColumn, PublicInput, VectorAccess,
+    Identifier, MatrixAccess, PeriodicColumn, PublicInput, TraceCol, VectorAccess,
 };
 use std::fmt::Display;
 
@@ -130,13 +130,18 @@ impl SymbolTable {
     pub(super) fn insert_trace_columns(
         &mut self,
         trace_segment: TraceSegment,
-        columns: &[Identifier],
+        columns: &[TraceCol],
     ) -> Result<(), SemanticError> {
         self.set_num_trace_segments(trace_segment);
 
-        for (idx, Identifier(name)) in columns.iter().enumerate() {
-            let trace_column = TraceColumn::new(trace_segment, idx);
-            self.insert_symbol(name, IdentifierType::TraceColumn(trace_column))?;
+        for (idx, trace_col) in columns.iter().enumerate() {
+            match trace_col {
+                TraceCol::Single(Identifier(name)) => {
+                    let trace_column = TraceColumn::new(trace_segment, idx);
+                    self.insert_symbol(name, IdentifierType::TraceColumn(trace_column))?;
+                }
+                TraceCol::Group(_, _) => todo!(),
+            }
         }
 
         Ok(())
--- a/ir/src/symbol_table.rs
+++ b/ir/src/symbol_table.rs
@@ -4,7 +4,7 @@ use super::{
 };
 use parser::ast::{
     constants::{Constant, ConstantType},
-    Identifier, MatrixAccess, PeriodicColumn, PublicInput, TraceCol, VectorAccess,
+    Identifier, MatrixAccess, PeriodicColumn, PublicInput, TraceCols, VectorAccess,
 };
 use std::fmt::Display;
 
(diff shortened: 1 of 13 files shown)


ir/src/symbol_table.rs
```diff
@@ -132,13 +132,18 @@ impl SymbolTable {
...
-            let trace_column = TraceColumn::new(trace_segment, idx);
-            self.insert_symbol(name, IdentifierType::TraceColumn(trace_column))?;
+        for (idx, trace_col) in columns.iter().enumerate() {
+            match trace_col {
+                TraceCol::Single(Identifier(name)) => {
+                    let trace_column = TraceColumn::new(trace_segment, idx);
+                    self.insert_symbol(name, IdentifierType::TraceColumn(trace_column))?;
+                }
+                TraceCol::Group(_, _) => todo!(),
+            }
```
grjte, line 146: I think it will end up being cleaner here, in the symbol table, and elsewhere in the IR if we don't have to deal with multiple variants like this. See my comments in the AST.

grjte (changes requested):
Thanks @tohrnii! The grammar looks good, but I think we may want to simplify the AST a bit to make life easier in the IR. See my comments inline and let me know if you have any questions

parser/src/ast/mod.rs
```diff
@@ -53,8 +53,22 @@ pub enum SourceSection {
...
+}
+
+/// [TraceCol] is used to represent a single or a group of columns in the execution trace.
+#[derive(Debug, Eq, PartialEq)]
+pub enum TraceCol {
+    Single(Identifier),
+    Group(Identifier, u64),
+}
+
+/// [TraceCol] is used to represent a single or a group of columns in the execution trace.
```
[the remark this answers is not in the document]
tohrnii: > This doc comment is the same as the one above, and it's not fully clear what the difference is between these. I'm guessing the usize in the GroupAccess refers to an index?

Sorry for the confusion, that is correct.

parser/src/ast/mod.rs
```diff
@@ -50,11 +50,55 @@ pub enum SourceSection {
...
-/// [TraceCols] contains the main and auxiliary trace columns of the execution trace.
+/// [Trace] contains the main and auxiliary trace columns of the execution trace.
+#[derive(Debug, Eq, PartialEq)]
+pub struct Trace {
+    pub main_cols: Vec<TraceCols>,
+    pub aux_cols: Vec<TraceCols>,
+}
+
+/// [TraceCols] is used to represent a single or a group of columns in the execution trace. For
+/// single columns, the size is 0. For groups, the size is the number of columns in the group.
```
Al-Kindi-0, line 61: Just for my own understanding, is the number of columns sufficient to pinpoint the exact columns (e.g. contiguous and starting from 0)?

grjte: Regarding this doc comment:
Shouldn't the size be 1 for single columns? (i.e. the number of columns in the group)

parser/src/ast/mod.rs
```diff
@@ -50,11 +50,55 @@ pub enum SourceSection {
...
+        &self.name
+    }
+
+    pub fn size(&self) -> u64 {
+        self.size
+    }
+}
+
+/// [TraceAccess] is used to represent a column inside a trace column group at the specified index
+/// within the group. For columns outside a group, the index is always 0.
```
grjte, line 83: nit: I think this is a bit confusing. Maybe something like the following would be clearer:

"[TraceAccess] is used to indicate a column in the trace by specifying its index within a set of [TraceCols] with the given identifier. If the identifier refers to a single column ([TraceCols] with size 1), then the index is always zero."

parser/src/parser/tests/trace_columns.rs
```diff
@@ -25,20 +28,50 @@ fn trace_columns_main_and_aux() {
...
 }
 
+#[test]
+fn trace_columns_groups() {
+    let source = "
+    trace_columns:
+        main: [clk, fmp, ctx, a[3]]
+        aux: [rc_bus, b[4], ch_bus]
+    transition_constraints:
+        enf a[1]' = 1";
```
grjte, line 52: We should probably also include an example of a trace column that isn't indexed into (e.g. `clk`) to ensure that it ends up with the correct index (e.g. as `TraceAccess::new(Identifier("clk".to_string()), 0)`)

parser/src/ast/mod.rs
tohrnii: @grjte Fixed.

@Al-Kindi-0's I'm not sure if you meant this but from your comment it seems like you mean whether column groups need to be contiguous? Currently they are assumed to be contiguous but  I'm wondering if it's possible for some projects that it might be meaningful to have columns in a group that are not contiguous? (@grjte ?). I don't know if any clean syntax is possible to support such use cases though.

ir/src/symbol_table.rs
```diff
@@ -132,13 +132,18 @@ impl SymbolTable {
...
         self.set_num_trace_segments(trace_segment);
 
-        for (idx, Identifier(name)) in columns.iter().enumerate() {
-            let trace_column = TraceColumn::new(trace_segment, idx);
-            self.insert_symbol(name, IdentifierType::TraceColumn(trace_column))?;
+        let mut col_idx = 0;
+        for trace_cols in trace {
+            let trace_column = TraceColumn::new(trace_segment, col_idx);
+            self.insert_symbol(
+                trace_cols.name().name(),
```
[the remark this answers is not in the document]
tohrnii: Makes sense. Fixed.

tohrnii:
Thank you @grjte and @Al-Kindi-0 for the review. Made suggested changes.

parser/src/ast/mod.rs
Al-Kindi-0: Thank you @grjte and @tohrnii , the answer of @grjte addresses my question. Regarding whether it might be useful to support non-contiguous cases, I can't think of cases where one might need it at the moment.

grjte (approved):
Looks good to me, thanks for all your work @tohrnii!

grjte (approved):

Overcastan (approved):

Al-Kindi-0 (approved):

PR #102, opened 2022-12-10
Add trace column grouping to IR

### LINKED ISSUE #85: Update IR to preserve column groupings for codegen
After support for grouping of trace columns is added in #83, we should preserve the group in IR as well so that it can be referenced by codegen with the column group's name.

tohrnii:
Closes #85.


ir/src/trace_column.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub fn col_idx(&self) -> usize {
+        self.col_idx
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct TraceColumnGroup {
+    trace_segment: TraceSegment,
+    size: usize,
+    start_col_idx: usize,
```
grjte, line 35: nit: instead of `start_col_idx` I'd prefer `offset`. It's more consistent with the way we've done things in the Miden VM code

ir/src/transition_constraints/graph.rs
```diff
@@ -159,21 +163,38 @@ impl AlgebraicGraph {
...
+                    let trace_access =
+                        TraceAccess::new(trace_segment, idx + column_group.start_col_idx(), 1);
+                    let node_index = self.insert_op(Operation::TraceElement(trace_access));
+                    Ok((trace_segment, node_index))
+                } else {
+                    Err(SemanticError::InvalidUsage(format!(
+                        "Identifier {} was declared as a {} not as a trace column group",
+                        ident, col_type
+                    )))
+                }
```
grjte, line 196: There's a lot of duplication here. As mentioned in my review of #101, I think it would be preferable to treat TraceAccess in the same way whether it's a group or a single column. When we're doing things in this graph, we just need a reference within the trace. The only place that really needs to know about the distinction between groups and single columns is the symbol table, because 1) we do validation there, and 2) we may want the column grouping information at the end for codegen.

ir/src/error.rs
```diff
@@ -4,6 +4,7 @@ use parser::ast::{MatrixAccess, VectorAccess};
 #[derive(Debug)]
 pub enum SemanticError {
     DuplicateIdentifier(String),
+    DuplicateTraceColumn(String),
```
grjte, line 7: I don't think this is ever used, is it?

ir/src/boundary_constraints.rs
```diff
@@ -95,6 +95,40 @@ impl BoundaryConstraints {
...
+                        1 => match constraint.boundary() {
+                            ast::Boundary::First => self
+                                .aux_first
+                                .insert(column_group.start_col_idx() + idx, expr),
+                            ast::Boundary::Last => self
+                                .aux_last
+                                .insert(column_group.start_col_idx() + idx, expr),
+                        },
+                        _ => unimplemented!(),
+                    }
```
grjte, line 124: You could condense this a bit by defining a variable before the match for `column_group.start_col_idx() + idx`

ir/src/boundary_constraints.rs
```diff
@@ -95,6 +95,40 @@ impl BoundaryConstraints {
                 },
                 _ => unimplemented!(),
             },
+            IdentifierType::TraceColumnGroup(column_group) => {
+                if let TraceColAccess::GroupAccess(name, idx) = constraint.column() {
+                    if *idx >= column_group.size() {
+                        return Err(SemanticError::vector_access_out_of_bounds(
+                            &VectorAccess::new(name.clone(), *idx),
+                            column_group.size(),
+                        ));
```
grjte, line 104: this should probably be `trace_access_out_of_bounds` instead of vector access

ir/src/lib.rs
```diff
@@ -27,16 +30,17 @@ pub type TraceSegment = u8;
...
 ///
 /// TODO: docs
 #[derive(Default, Debug)]
 pub struct AirIR {
     air_name: String,
-    //TODO: remove dead code attribute
-    #[allow(dead_code)]
     constants: Constants,
+    #[allow(dead_code)]
+    trace_column_groups: TraceColumnGroups,
```
grjte, line 43: We could probably skip adding this to the AirIR for now.

ir/src/symbol_table.rs
```diff
@@ -185,8 +173,26 @@ impl SymbolTable {
...
+            .identifiers
+            .iter()
+            .filter_map(|(name, ident_type)| {
+                if let IdentifierType::TraceColumnGroup(group) = ident_type {
+                    Some((name.clone(), group.clone()))
+                } else {
+                    None
+                }
+            })
+            .collect();
```
grjte, line 189: I think we can skip this for now and keep `into_declarations`as it was. I think anyone who wants a named trace column group for codegen may want the identifier info for the other columns too, so I think it makes sense to hold off on doing this until there's a use-case to define what's actually needed

grjte (changes requested):
This is just a partial review, since I think there should be some changes based on my comments in #101. I left a few comments inline related to #101 and a few other small things

Al-Kindi-0 (commented):
Looks good, thank you! I think the only thing missing is to translate the suggestions in #101 to this PR.

ir/src/symbol_table.rs
tohrnii: Makes sense. I was thinking the same. Removed.

ir/src/boundary_constraints.rs
tohrnii: Thanks, fixed.

ir/src/boundary_constraints.rs
tohrnii: The code has been refactored.

ir/src/error.rs
tohrnii: Thanks, removed.

ir/src/transition_constraints/graph.rs
tohrnii: Please let me know if you disagree with the current refactor.

ir/src/symbol_table.rs
```diff
@@ -140,8 +123,17 @@ impl SymbolTable {
...
+                let trace_column = TraceColumn::new(trace_segment, col_idx);
+                self.insert_symbol(trace_cols.name(), IdentifierType::TraceColumn(trace_column))?;
+            } else {
+                let trace_column_group =
+                    TraceColumnGroup::new(trace_segment, trace_cols.size() as usize, col_idx);
+                self.insert_symbol(
+                    trace_cols.name(),
+                    IdentifierType::TraceColumnGroup(trace_column_group),
+                )?;
+            }
```
grjte, line 136: I'm not sure we need 2 different types here. What's your reasoning for not just having a single type like we did in the parser where for single columns the size of the group is 1?

grjte (changes requested):
It looks good to me overall! I'm just wondering about managing the trace columns in a similar way to what we just did in the parser, in which case changes would be needed in a few files (although actually it wouldn't be a major refactor)

ir/src/symbol_table.rs
tohrnii: No good reason. Only that inserting access in graph is defined differently for symbol and vector so it seemed equivalent to me in terms of code readability/redundancy. If we go with defining one type, we will have to do checks in 2 places while inserting it in graph (one for symbol access and another for vector access). Please let me know if you prefer keeping one type here.

tohrnii: Although maybe that check is redundant. We should only get vector access when size > 1.

tohrnii: No good reason, I was mistaken. Fixed. Thanks.

ir/src/transition_stmts/graph.rs
```diff
@@ -287,6 +286,16 @@ impl AlgebraicGraph {
...
                 }
             }
+            IdentifierType::TraceColumns(columns) => {
+                let trace_segment = columns.trace_segment();
+                let col_idx = columns.offset() + vector_access.idx();
+                let node_index = self.insert_op(Operation::TraceElement(TraceAccess::new(
+                    trace_segment,
+                    col_idx,
+                    0,
+                )));
```
grjte, line 296: nit: Maybe we can use constants `CURRENT_ROW` and `NEXT_ROW` for the row offsets of 0 and 1 or at least add comments where these `TraceAccess` structs are created? I think it would be easier to read what's going on for anyone who is new to the codebase or has stepped away for a week and apparently forgotten everything ;)

grjte (approved):
Looks good to me, thanks @tohrnii! I've left 2 opinionated non-blocking comments inline.

ir/src/trace_columns.rs
```diff
@@ -1,39 +1,36 @@
...
+/// Describes columns in the execution trace by the trace segment to which it belongs, it's size
+/// and it's offset.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct TraceColumns {
-    /// A map of a set of trace columns using the declared identifier as the key and the column
-    /// index as the value.
-    columns: BTreeMap<String, usize>,
+    trace_segment: TraceSegment,
+    size: usize,
+    offset: usize,
```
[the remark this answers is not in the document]
tohrnii: Makes sense. My thinking was that the size is a more local property of the column group and should come before the offset which is meaningful in the context of the whole trace. But this argument probably doesn't make sense since `trace_segment` is also in the context of the whole trace. Will change it.

ir/src/transition_stmts/graph.rs
tohrnii: Changed it to use `CURRENT_ROW` and `NEXT_ROW`. Thanks for the suggestion, I used to get confused with this one as well.

Overcastan (approved):
Looks good! Thank you!

Al-Kindi-0 (approved):
Looks good, thank you!

Issue #56, opened 2022-11-02
Implement codegen option for generic constraint evaluation format


bobbinth:
The output of this codegen option should be a JSON file with the following structure:

```
{
    num_polys: integer,
    num_variables: integer,
    constants: [array of field elements in the Goldilocks field],
    expressions: [array of expression nodes],
    outputs: [array of node references],
}
```

In the above, node reference has the following structure:
```
{
  type: "POL | POL_NEXT | VAR | CONST | REF",
  index: integer,
}
```
Where:
* `POL` refers to the value in the column at the specified `index` in the current row.
* `POL_NEXT` refers to the value in the column at the specified `index` in the next row.
* `VAR` refers to a public input or a random value at the specified `index`.
* `CONST` refers to a constant at the specified `index`.
* `REF` refers to a previously defined expression at the specified `index`.

An expression has the following structure:
```
{
  op: "ADD | SUB | MUL",
  lhs: node_reference,
  rhs: node_reference, 
}
```
Where `ADD`, `SUB`, and `MUL` are the corresponding operations in the Goldilocks base field.

### Code generation
To generate the code in the above format, we would need to do the following:

1. Merge main and auxiliary trace columns together into a single list of columns. For example, if we have 5 main trace columns and 2 aux trace columns, we should end up with with a list of either 9 or 11 columns, depending on the extension degree specified (i.e., 9 columns for quadratic extension and 11 columns for cubic extension). This also means that field extension degree should be an input parameter for this codegen option.
2. Flatten public inputs into a single array of values and merge it with random values array. This array will represent the `variables` array for the expressions.
3. Reduce all operations to the 3 supported operations in the base field. This also means that all operations over the extension field must be transformed into equivalent operations in the base field.

In addition to the above, we also need to add constraint merging logic to the constraint evaluation description. There should be only 3 entry points in the `outputs` section:
- One entry point to compute merged boundary constraints against the first step.
- One entry point to compute merged boundary constraints against the last step.
- One entry point to compute merged transition constraints.

The logic for merging transition constraints is described [here](https://docs.rs/winter-air/0.4.1/winter_air/struct.TransitionConstraintGroup.html#method.merge_evaluations) and the logic for computing and merging boundary constraints is described [here](https://docs.rs/winter-air/0.4.1/winter_air/struct.BoundaryConstraintGroup.html#method.evaluate_at).

A few other things to consider:
[...]

PR #104, opened 2022-12-15
Codegen option for generic constraint evaluation format

Fumuran:
Related issue: #56 

Things left to implement:
- [x] boundary constraints (for `expressions` and `outputs` arrays)
- [x] EXP operation emulation 
- [x] Random Values support

Things to implement in later version: 
- Periodic Columns support
- Constraint merging logic


bobbinth:
One quick comment: this should go into its own crate (not winterfell codegen). So, instead of being under `codegen/winterfell` it should be under `codegen/gce` (for "generic constraint evaluator"). But I'm also open to other suggestions. @grjte - how do you think we should name it?

grjte:
> So, instead of being under `codegen/winterfell` it should be under `codegen/gce` (for "generic constraint evaluator"). But I'm also open to other suggestions. @grjte - how do you think we should name it?

I think `gce` works - I've thought a bit, but there's nothing else I prefer.

ir/src/lib.rs
```diff
@@ -69,6 +76,8 @@ impl AirIR {
                     symbol_table.insert_constants(constants)?;
                 }
                 ast::SourceSection::TraceCols(columns) => {
+                    // set the overall trace columns count
+                    num_polys = (columns.main_cols.len() as u8, columns.aux_cols.len() as u8);
```
grjte, line 80: This won't quite work, because we could have more than one `TraceCols` source section (e.g. once we support modules).

Instead of doing it here and initializing above, maybe we should collect this information inside the symbol table, the way that we collect the `num_trace_segments` information. Then you could update the `into_declarations` method to pass the information you need to initialize `num_polys`. I think this will encapsulate things a bit more cleanly.

ir/src/lib.rs
```diff
@@ -37,6 +37,10 @@ pub struct AirIR {
     //TODO: remove dead code attribute
     #[allow(dead_code)]
     constants: Constants,
+    // I don't think that adding a new field just for generic constraint evaluation worth it.
+    // TODO: get rid of this field
+    // amount of named columns: (main, aux)
+    num_polys: (u8, u8),
```
grjte, line 43: I like that you're considering whether or not we want this, but I actually do think we need to add something here, because I don't think there's any other way to get this information. 

One note though - currently we only have 2 trace segments, but we've actually been making the IR more generic to support more than 2 segments, so this should probably be a vector to be consistent with those changes.

grjte (changes requested):
This is a partial review. I left several preliminary comments inline related to refactoring. After that, it should be easier to assess some of the other design decisions here.

air-script/tests/helpers.rs
```diff
@@ -8,43 +9,54 @@ pub enum TestError {
...
+    ir: AirIR,
 }
 
 impl Test {
-    pub fn new(input_path: String) -> Self {
-        Test { input_path }
-    }
-
-    pub fn transpile(&self) -> Result<String, TestError> {
+    pub fn new(input_path: String) -> Result<Self, TestError> {
```
grjte, line 20: I'd prefer to keep these separate. The idea is that if they're separate then in the future we could offer different options - full transpilation via `transpile`, or only testing the `parsing` or `ir` generation steps

air-script/tests/helpers.rs
```diff
@@ -1,4 +1,5 @@
-use codegen_winter::CodeGenerator;
+use codegen_gce::GCECodeGenerator;
+use codegen_winter::WinterfellCodeGenerator;
```
grjte, line 2: Once we have both of these in the `air_script` library, you can import as:

```
use air_script::{GceCodeGenerator, WinterfellCodeGenerator};
```

I think we should actually do the same with the IR and parser inputs, so we would have:


```
use air_script::{parse, AirIR, GceCodeGenerator, WinterfellCodeGenerator};
```

air-script/tests/helpers.rs
```diff
@@ -8,43 +9,54 @@ pub enum TestError {
     IO(String),
     Parse(String),
     IR(String),
+    Gce(String),
```
grjte, line 12: We should probably make this more generic for`Codegen`, rather than just `Gce`.

air-script/tests/main.rs
```diff
@@ -1,86 +1,221 @@
...
 }
 
 #[test]
-fn binary() {
-    let generated_air = Test::new("tests/binary/binary.air".to_string())
-        .transpile()
+fn gce_aux_trace() {
+    Test::new("tests/aux_trace/aux_trace.air".to_string())
+        .unwrap()
+        .generate_gce(2, "tests/aux_trace/generated_aux_trace.json")
```
grjte, line 25: Let's not unwrap twice in the same chain here. Instead, we can do something like this.

```
    let test = Test::new("tests/aux_trace/aux_trace.air".to_string()).unwrap();
    assert!(test
        .generate_gce(2, "tests/aux_trace/generated_aux_trace.json")
        .is_ok());
```

codegen/gce/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+};
+
+mod boundary_constraints;
+use boundary_constraints::set_boundary_expressions_and_outputs;
+
+mod error;
+use error::ConstraintEvaluationError;
+
+mod helpers;
+use helpers::{acumulate_constants, count_boundary_rand_values, Expression, NodeType};
```
grjte, line 14: nit: there's a typo here - this should be `accumulate_constants`

codegen/gce/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+fn set_num_polys(ir: &AirIR, extension_degree: u8) -> u16 {
+    // TODO: Should all aux columns be extended to be quadratic or cubic?
+    let num_polys_vec = ir.num_polys();
+    num_polys_vec
+        .iter()
+        .skip(1)
+        .fold(num_polys_vec[0], |acc, &x| {
+            acc + x * extension_degree as u16
+        })
+}
```
grjte, line 117: Please add doc comments to this and all helper functions to explain the intended functionality (the meaning of the inputs & outputs)

grjte (changes requested):
Thanks @Overcastan! This refactor looks a lot cleaner. I've done a partial review. Can you please address these comments and add doc comments to functions where they're missing? It'll make it easier to review the rest of this.

air-script/tests/main.rs
```diff
@@ -1,86 +1,221 @@
-use expect_test::expect_file;
+use std::fs::File;
+use std::io::prelude::*;
```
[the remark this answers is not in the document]
Overcastan: Fixed. There was no reason, it looks like the auto import just arranged them like this, and I overlooked

air-script/tests/aux_trace/aux_trace.json
```diff
@@ -0,0 +1,28 @@
...
+		{"op": "MUL", "lhs": {"type": "POL_NEXT", "index": 4}, "rhs": {"type": "EXPR", "index": 11}},
+		{"op": "SUB", "lhs": {"type": "POL", "index": 4}, "rhs": {"type": "EXPR", "index": 12}},
+		{"op": "SUB", "lhs": {"type": "POL", "index": 0}, "rhs": {"type": "CONST", "index": 0}},
+		{"op": "SUB", "lhs": {"type": "POL", "index": 1}, "rhs": {"type": "CONST", "index": 0}},
+		{"op": "SUB", "lhs": {"type": "POL", "index": 3}, "rhs": {"type": "CONST", "index": 0}},
+		{"op": "SUB", "lhs": {"type": "POL", "index": 4}, "rhs": {"type": "VAR", "index": 16}},
+		{"op": "SUB", "lhs": {"type": "POL", "index": 3}, "rhs": {"type": "CONST", "index": 0}},
+		{"op": "SUB", "lhs": {"type": "POL", "index": 4}, "rhs": {"type": "CONST", "index": 0}}
+	],
+	"outputs": [1, 3, 5, 10, 13, 14, 15, 16, 17, 18, 19]
```
grjte, line 27: This output isn't correct, since the "outputs" should be an array of node references.

So for example, the first transition constraint `enf a' = b + c` would be expressed by the following:

```
	"expressions": [
		{"op": "ADD", "lhs": {"type": "POL", "index": 1}, "rhs": {"type": "POL", "index": 2}},
		{"op": "SUB", "lhs": {"type": "POL_NEXT", "index": 0}, "rhs": {"type": "REF", "index": 0}}
	],
	"outputs": [
		{"type": "POL_NEXT", "index": 0},
		{"type": "POL", "index": 1},
		{"type": "POL", "index": 2},
		{"type": "CONST", "index": 0},
		{"type": "REF", "index": 0}
	]
```

In this case, the first item in the expressions array represents `b+c`, which is then added to outputs as a `REF`. The second item in the expressions array represents `a' - (b+c)`

codegen/gce/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        let num_polys = set_num_polys(ir, extension_degree);
+        let num_variables =
+            set_num_variables(ir, &mut const_public_type_map, boundary_constraints_vec);
+
+        // TODO #1: get rid of the vector and push values directly into result string
+        // constants from Constants AirIR field
+        // TODO #2: currently I add all found constants in the vector. Should I add only unique ones,
+        // since I'll get constants by their value, not index?
+        let constants = set_constants(ir, &mut const_public_type_map, boundary_constraints_vec);
```
grjte, line 64: I'm not sure if I understood this question correctly, but we should be referencing constants by index, so yes, it would be best to only add unique ones.

In other words, if we had `a'= a + 1`, we would have something like the following:

```
...
        "constants": [1]
	"expressions": [
		{"op":"ADD", "lhs": {"type": "POL", "index": 0}, "rhs": {"type": "CONST", "index": 0}},
		{"op": "SUB", "lhs": {"type": "POL_NEXT", "index": 0}, "rhs": {"type": "REF", "index": 0}}
	],
	"outputs": [
		{"type": "POL_NEXT", "index": 0},
		{"type": "POL", "index": 0},
		{"type": "CONST", "index": 0},
		{"type": "REF", "index": 0}
	]
```

grjte (changes requested):
Another partial review :) The testing structure looks good now except for one nit inline. The test reference code is incorrect, though. I would start by thinking through what the reference json should be to make sure it's clear. I left a few examples inline, but please let me know if you have questions

codegen/gce/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+    num_variables: usize,
+    constants: Vec<u64>,
+    expressions: Vec<ExpressionJson>,
+    outputs: Vec<usize>,
+}
+
+impl CodeGenerator {
+    pub fn new(ir: &AirIR, extension_degree: u8) -> Result<Self, ConstraintEvaluationError> {
+        // maps indexes in Node vector in AlgebraicGraph and in `expressions` JSON array
+        let mut expressions_map = BTreeMap::new();
```
grjte, line 37: This isn't used at this level, so it can be removed here and created directly when the `ExpressionsHandler` is instantiated

codegen/gce/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        let num_polys = set_num_polys(ir, extension_degree);
+        let num_variables = set_num_variables(ir);
+        let constants = set_constants(ir);
+
+        let mut expressions_handler = ExpressionsHandler::new(ir, &constants, &mut expressions_map);
+
+        let mut expressions = expressions_handler.get_expressions()?;
+        // vector of `expressions` indexes
+        let outputs = expressions_handler.get_outputs(&mut expressions)?;
```
grjte, line 47: Instead of this pattern where you get the expressions and then pass them back in, I would keep them in the `ExpressionsHandler`. I suggest adjusting to a builder pattern. We could rename `ExpressionsHandler` to `GceBuilder`, add a method which handles building the `expressions` and `outputs`, and then add an `into` method that consumes the builder and yields the `expressions` and `outputs`. (This would also reflect the pattern that we'll be moving the IR to in #150)

Something like:

```
    pub fn new(ir: &AirIR, extension_degree: u8) -> Result<Self, ConstraintEvaluationError> {
        let num_polys = set_num_polys(ir, extension_degree);
        let num_variables = set_num_variables(ir);
        let constants = set_constants(ir);

        let builder = GceBuilder::new();
        builder.build(ir, &constants)?;
        let (expressions, outputs) = builder.into_gce();

        Ok(CodeGenerator {
            num_polys,
            num_variables,
            constants,
            expressions,
            outputs,
        })
    }
```

codegen/gce/src/expressions.rs
```diff
@@ -0,0 +1,364 @@
...
+use std::collections::BTreeMap;
+
+const MAIN_TRACE_SEGMENT_INDEX: u8 = 0;
+
+pub struct ExpressionsHandler<'a> {
+    ir: &'a AirIR,
+    constants: &'a [u64],
+    // maps indexes in Node vector in AlgebraicGraph and in `expressions` JSON array
+    expressions_map: &'a mut BTreeMap<usize, usize>,
+}
```
grjte, line 23: As noted, the `expressions_map` can be instantiated in the `new` method, so it no longer needs to be marked as a mut borrow with a lifetime.

I think we should also add the following. As mentioned, we can declare them in `new` and then populate them in a `build` method.

```
    expressions: Vec<ExpressionJson>,
    outputs: Vec<usize>,
```

I would also probably remove the `AirIR` and `constants` borrows here and just pass them to the `build` method.

For the `build` method, we could do something like the following:

```
    pub fn build(
        &mut self,
        ir: &AirIR,
        constants: &[u64],
    ) -> Result<(), ConstraintEvaluationError> {
        self.build_expressions(ir, constants)?;
        self.build_outputs(ir)?;
        Ok(())
    }
```

codegen/gce/src/expressions.rs
```diff
@@ -0,0 +1,364 @@
...
+        ExpressionsHandler {
+            ir,
+            constants,
+            expressions_map,
+        }
+    }
+
+    /// Parses expressions in transition graph's Node vector, creates [Expression] instances and pushes
+    /// them to the `expressions` vector.
+    pub fn get_expressions(&mut self) -> Result<Vec<ExpressionJson>, ConstraintEvaluationError> {
```
grjte, line 40: We can change the signature to the following to use the builder pattern I suggested:

```
fn build_expressions(
        &mut self,
        ir: &AirIR,
        constants: &[u64],
    ) -> Result<(), ConstraintEvaluationError> {
```

codegen/gce/src/expressions.rs
```diff
@@ -0,0 +1,364 @@
...
+            }
+        }
+        Ok(expressions)
+    }
+
+    /// Fills the `outputs` vector with indexes from `expressions` vector according to the `expressions_map`.
+    pub fn get_outputs(
+        &self,
+        expressions: &mut Vec<ExpressionJson>,
+    ) -> Result<Vec<usize>, ConstraintEvaluationError> {
```
grjte, line 115: We can change the signature to the following to implement the builder pattern I suggested:

```
fn build_outputs(&self, ir: &AirIR) -> Result<Vec<usize>, ConstraintEvaluationError> {
```

grjte (changes requested):
@Overcastan thank you! The refactor you've done looks good. It's much easier to follow. This is a partial review with a few nits and a small suggested adjustment to the pattern you're using to build the expressions & outputs

bobbinth:
Closing as obsolete.

Issue #84, opened 2022-11-28
Add validity constraints


tohrnii:
We should add support for defining validity constraints. These constraints would be similar to transition constraints, with the only exception that they would work only with the current row of the trace. For example:

```
trace_columns:
    main: [a, b, c]

integrity_constraints:
    enf a^2 - a = 0    # this is a validity constraint
```

These constraints should be in the same section as transition constraints however we should change the name of the combined section to `integrity_constraints` (or something similar). Also, in the IR we should still probably have a single algebraic DAG, but we could mark entry points as transition vs. validity constraints.

_Originally posted by @bobbinth in #53._

grjte:
I think `integrity_constraints` is fine as a name

grjte:
Our existing "transition_constraints" currently includes both validity and transition constraints, so the only change needed there is to distinguish between the constraint types in the IR.

I think the tasklist for this issue looks something like the following:

- [ ] rename `transition_constraints` to `integrity_constraints`
- [ ] in the IR, identify constraints as "validity" or "transition" while adding them to the graph, and label them somehow in the entry point lists

Question: Instead of tracking 1 array of entry nodes should we track 2 arrays of constraint entry nodes - one for validity constraints and one for transition constraints? We don't need this for Miden, but it would be needed for something like Polygon Zero.

@bobbinth what do you think?

bobbinth:
For what we need now, two separate arrays will probably suffice, but I wonder if we should make this even more general for what we might do in the future. Specifically, each entry point could contain a set of row offsets used for a given constraints. The offsets here would be similar to the ones mentioned in #54.

For example, if the set of offsets consists of a single offset `0`, then this is a validity constraint. If the set of offsets consists of offsets `0` and `1`, this is a transition constraint. In the future, we might have different types of transition constraints - e.g., a constraint which accesses rows `0`, `1`, and `5` or something like that.

grjte:
Sure, I think this makes sense to do. The more general approach is better for the future, and it's probably equal work, since making the change for #54 will probably help with identifying validity vs. transition constraints.

That means this issue is blocked by #54 and we should do that first.

Then for this issue, we would need to update the way constraints are tracked, so that instead of an array of `NodeIndex` (entry points to the constraint in the `AlgebraicGraph`), we would use something else that has the `NodeIndex` data and the row offset data?

bobbinth:
Agreed! And yes, instead of just `NodeIndex` it would be some struct which could look like this:
[...]

PR #105, opened 2022-12-20
Rename transition constraints to integrity constraints

tohrnii:
Partly addressing #84.

This PR renames transition constraints to integrity constraints to make the section name more generic to be able to differentiate between validity and transition constraints in the IR.


ir/src/lib.rs
```diff
@@ -179,21 +179,20 @@ impl AirIR {
...
-    pub fn transition_constraints(&self, trace_segment: TraceSegment) -> &[NodeIndex] {
-        self.transition_constraints.constraints(trace_segment)
+    pub fn integrity_constraints(&self, trace_segment: TraceSegment) -> &[NodeIndex] {
+        self.integrity_constraints.constraints(trace_segment)
     }
 
-    pub fn transition_graph(&self) -> &AlgebraicGraph {
-        self.transition_constraints.graph()
+    pub fn integrity_graph(&self) -> &AlgebraicGraph {
+        self.integrity_constraints.graph()
```
grjte, line 196: I'm not sure about the name `integrity_graph`. The rest of the renaming mostly makes sense and preserves meaning, but this feels like the meaning changes a bit. I think it might be better to call this `constraint_graph`. What do you think? I know it doesn't include the boundary constraints, but I think that's ok (and at some point in the future it might).

codegen/winterfell/src/air/mod.rs
```diff
@@ -13,8 +13,8 @@ use periodic_columns::add_fn_get_periodic_column_values;
 mod boundary_constraints;
 use boundary_constraints::{add_fn_get_assertions, add_fn_get_aux_assertions};
 
-mod transition_constraints;
-use transition_constraints::{add_fn_evaluate_aux_transition, add_fn_evaluate_transition};
+mod integrity_constraints;
+use integrity_constraints::{add_fn_evaluate_aux_transition, add_fn_evaluate_transition};
```
grjte, line 17: I think it might be better not to rename these, since Winterfell doesn't have an idea of `integrity constraints`, it only has `boundary` or `transition` constraints. I think that our codegen crates will probably have the most clarity if they generally use the same language as the target. This is just my opinion though. What do you think?

If we agree on this, we should restore the comments below and the text in the README as well

ir/src/lib.rs
```diff
@@ -10,9 +10,9 @@ use symbol_table::{IdentifierType, SymbolTable};
 pub mod boundary_constraints;
 use boundary_constraints::BoundaryConstraints;
 
-pub mod transition_constraints;
-use transition_constraints::{AlgebraicGraph, TransitionConstraints, MIN_CYCLE_LENGTH};
-pub use transition_constraints::{NodeIndex, TransitionConstraintDegree};
+pub mod integrity_constraints;
+use integrity_constraints::{AlgebraicGraph, IntegrityConstraints, MIN_CYCLE_LENGTH};
+pub use integrity_constraints::{NodeIndex, TransitionConstraintDegree};
```
grjte, line 15: I think we should probably rename `TransitionConstraintDegree` as well, to keep everything in the IR consistent. That would mean updating all of the related comments, etc. as well. It seems like a deliberate decision that you chose not to do this though, so it would be good to understand your reasoning & discuss

grjte (changes requested):
Thank you! This looks good to me overall, but I left a few comments inline for discussion.

ir/src/lib.rs
tohrnii: I wasn't sure if we wanted to change this naming since this struct is defined in winterfell so I wasn't sure if we should keep it the same in IR as well or not. But I think it makes sense to change this in IR. Thanks.

codegen/winterfell/src/air/mod.rs
tohrnii: Makes sense. Will revert. Thanks.

ir/src/lib.rs
grjte: I would do `constraint_graph`

ir/src/lib.rs
grjte: Cool, I figured that was your thinking (and it makes sense). Sounds good to change it for our IR though

ir/src/lib.rs
bobbinth: I like `constraint_graph`. I actually think that in the future we'll probably fold boundary constraints into this graph as well.

tohrnii:
Thank you @grjte for the review. Made suggested changes. I'll rebase this one after 102 is merged and we can merge this one after that.

ir/README.md
```diff
@@ -28,6 +28,6 @@ Although generation of an `AirIR` uses a symbol table while processing the sourc
 - **Public inputs**, represented by a vector that maps an identifier to a size for each public input that was declared. (Currently, public inputs can only be declared as fixed-size arrays.)
```
Al-Kindi-0, line 28: nit: is the sentence "...to a size..." complete?

Al-Kindi-0 (commented):
Looks good to me, thank you! Only one minor nit which might not be relevant.

ir/README.md
tohrnii: @Al-Kindi-0 It seems complete to me. Not sure if I'm missing something.

grjte: Technically it's not complete, because we have a comma instead of a verb (instead of "Public inputs are represented by a vector that maps..."). That's because it's really a continuation of the sentence above the set of bullet points. It might be clearer to say "Public inputs, which are represented by...", in which case the other bullet points should be similarly updated.

grjte: But it's technically correct as it is, as long as the whole chunk is read together (it's just not a sentence on its own)

grjte (approved):
Looks good, thanks @tohrnii!

Al-Kindi-0 (approved):
Looks good, thanks @tohrnii !

Overcastan (approved):
Looks good! Thank you!

Issue #106, opened 2022-12-22
merge boundary constraints into AlgebraicGraph and support variables


tohrnii:
Variable support to transition constraint was added in #100. We should also add support for variables to boundary constraints. It might be useful to also add boundary constraints to the constraints graph in the same PR we add the variables.

The original issue with syntax is described in #61.

grjte:
It now looks like we should do this task before completing #56, since having the boundary constraints in the constraints graph will simplify GCE generation a lot

grjte:
Closed by #116 #117 #126 and #160

PR #111, opened 2023-01-10
Differentiate between validity and transition constraints in constraint graph

tohrnii:
Partly addressing #84.

### DIFF
--- a/ir/src/integrity_stmts/graph.rs
+++ b/ir/src/integrity_stmts/graph.rs
@@ -1,7 +1,7 @@
 use super::{
     super::BTreeMap, degree::IntegrityConstraintDegree, SemanticError, SymbolTable, TraceSegment,
 };
-use crate::{symbol_table::IdentifierType, VariableRoots, CURRENT_ROW, NEXT_ROW};
+use crate::{symbol_table::IdentifierType, ExprDetails, VariableRoots, CURRENT_ROW, NEXT_ROW};
 use parser::ast::{
     self, constants::ConstantType, Identifier, IntegrityExpr, IntegrityVariableType, MatrixAccess,
     VectorAccess,
@@ -83,36 +83,31 @@ impl AlgebraicGraph {
 
     // --- MUTATORS -------------------------------------------------------------------------------
 
-    /// Adds the expression to the graph and returns the result index and a constraint type
-    /// indicating whether it is applied to the main execution trace or an auxiliary trace.
+    /// Adds the expression to the graph and returns the [ExprDetails] of the constraint.
     /// Expressions are added recursively to reuse existing matching nodes.
     pub(super) fn insert_expr(
         &mut self,
         symbol_table: &SymbolTable,
         expr: IntegrityExpr,
-        row_offsets: &mut Vec<usize>,
         variable_roots: &mut VariableRoots,
-    ) -> Result<(TraceSegment, NodeIndex), SemanticError> {
+    ) -> Result<ExprDetails, SemanticError> {
         match expr {
             IntegrityExpr::Const(value) => {
                 // constraint target defaults to Main trace.
                 let trace_segment = 0;
                 let node_index = self.insert_op(Operation::Constant(ConstantValue::Inline(value)));
-                Ok((trace_segment, node_index))
+                Ok((trace_segment, node_index, CURRENT_ROW))
             }
             IntegrityExpr::Elem(Identifier(ident)) => {
-                self.insert_symbol_access(symbol_table, &ident, row_offsets, variable_roots)
+                self.insert_symbol_access(symbol_table, &ident, variable_roots)
             }
             IntegrityExpr::VectorAccess(vector_access) => {
-                self.insert_vector_access(symbol_table, &vector_access, row_offsets, variable_roots)
+                self.insert_vector_access(symbol_table, &vector_access, variable_roots)
             }
             IntegrityExpr::MatrixAccess(matrix_access) => {
-                self.insert_matrix_access(symbol_table, &matrix_access, row_offsets, variable_roots)
+                self.insert_matrix_access(symbol_table, &matrix_access, variable_roots)
             }
-            IntegrityExpr::Next(trace_access) => {
-                row_offsets.push(NEXT_ROW);
-                self.insert_next(symbol_table, &trace_access)
-            }
+            IntegrityExpr::Next(trace_access) => self.insert_next(symbol_table, &trace_access),
             IntegrityExpr::Rand(index) => {
                 // constraint target for random values defaults to the second trace segment.
                 // TODO: make this more general, so random values from further trace segments can be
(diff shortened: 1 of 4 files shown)


ir/src/integrity_stmts/mod.rs
```diff
@@ -123,3 +126,28 @@ impl IntegrityStmts {
         Ok(())
     }
 }
+
+/// A [ConstraintRoot] represents the entry node of a subgraph representing an integrity constraint
+/// within the [AlgebraicGraph]. It also contains the row offsets for the constraint.
```
grjte, line 131: I think we need to define what is meant by "the row offsets for the constraint"

edit: also, as noted in my other comments,  I think this should change to "the row offset"

ir/src/integrity_stmts/mod.rs
```diff
@@ -98,10 +98,12 @@ impl IntegrityStmts {
             IntegrityStmt::Constraint(constraint) => {
                 let expr = constraint.expr();
 
+                let mut row_offsets = Vec::new();
                 // add it to the integrity constraints graph and get its entry index.
                 let (trace_segment, root_index) = self.constraints_graph.insert_expr(
                     symbol_table,
                     expr,
+                    &mut row_offsets,
```
grjte, line 106: I don't think that this should be added as a parameter. In the case of the `variable_roots` we have to pass it as a parameter, because we need to track them and modify the set of them across many constraints. In the case of the `row_offsets`, this is specific to each constraint, just like the `trace_segment` or `root_index`. Therefore, I think we should follow the same pattern as the `root_index` and have this be part of the return value of `insert_expr`

ir/src/integrity_stmts/graph.rs
```diff
@@ -90,6 +90,7 @@ impl AlgebraicGraph {
         &mut self,
         symbol_table: &SymbolTable,
         expr: IntegrityExpr,
+        row_offsets: &mut Vec<usize>,
```
grjte, line 93: As noted, I think we should change the output signature rather than the input. I also think we should  just track the highest row offset, since that defines the constraint. I don't see why we care about all of the row offsets at this point, and it will probably be simpler not to keep them all unless we actually have to. We should be able to just treat the row_offset the same way that we deal with the `trace_segment`, i.e. pass it up from the leaves and then take the max each time we do an operation with 2 sub-expressions such as `Add`

Let me know what you think and if there's something I overlooked.

Also, since we'd now be returning a tuple of 3 elements, it probably makes sense to have a convenience type alias for `(TraceSegment, NodeIndex, usize)`. It might also make sense to have an alias `DEFAULT` equal to `CURRENT_ROW`, so that we're not talking about rows when dealing with variables. That's just a stylistic thought though.

grjte (changes requested):
Thanks @tohrnii! I think we should take a slightly different approach. I left my thoughts inline

ir/src/integrity_stmts/graph.rs
tohrnii: That makes sense. I was going by the signature @bobbinth posted in the issue:
```
struct ConstraintRoot {
    index: NodeIndex,
    offsets: Vec<usize>,
}
```
I thought maybe Bobbin suggested that as keeping track of all the row offsets had utility for other projects that jump more than one rows but maybe I misunderstood something?

bobbinth: I think tracking all offsets could be useful in the future (when we allow constraints which work with more than 2 rows) because the constraints which access the same set of rows may be treated differently (though, I can't come up with concrete examples now).

In general, however, this structure will probably change once boundary constraints are integrated into the graph. Specifically, for each constraint we'll need to define a set of rows it works with and just offsets would not be enough (i.e., for boundary constraints, instead of offsets we'd need to specify whether it access the first or the last row). In Winterfell, there is a [ConstraintDivisor](https://github.com/facebook/winterfell/blob/main/air/src/air/divisor.rs#L13) struct which does something similar - but I think in our case we can probably use something simpler.

bobbinth: We could have something like `ConstraintDomain` enum:
```Rust
pub enum ConstraintDomain {
  FirstRow,           // for boundary constraints against the first row
  LastRow,            // for boundary constraints against the last row
  EveryRow,           // for validity constraints
  EveryWindow(usize), // for transition constraints
}
```
For `EveryWindow` the internal value specifies the size of the evaluation frame. E.g., for transition constraints over two consecutive rows the internal value would be `2`. I used the term `window` as this is what Rust uses for [similar concepts](https://doc.rust-lang.org/std/primitive.slice.html#method.windows) - but I'm not too happy with it. So, open to other suggestions.

If the need arises, we could add other variants to this enum.

grjte: hm follow-up, I guess you're talking about a case where we access e.g. 3 rows with offsets 1, 3, 5. So in this case we'd want to spot matching `ConstraintRoots` that also access the rows with these offsets, and that's why we'd want all of those offsets?

But then our window would be from row 0 to row 5, with `ConstraintDomain::EveryWindow(6)`, but that `ConstraintDomain` doesn't have the the information about the particular set either. Or are you thinking about capturing it by having multiple `ConstraintDomain` associated with the `ConstraintRoot`, e.g. `ConstraintDomain::EveryWindow(1)`, `ConstraintDomain::EveryWindow(3)`, `ConstraintDomain::EveryWindow(5)`

grjte: I guess it feels like unnecessary complexity for something for which we don't have a concrete use-case, and it wouldn't be hard to change this in the future if/when we do have a concrete use, at which point we'd have more information about what would work best.

bobbinth: Yes, the intent was to differentiate between constraints which require frame size 3, but one of them access all 3 rows, while the other one access only rows 0 and 2. But as you said, we don't need this for now - and probably makes sense to add this when/if we need it.

I like `Frame`! So, the enum would be:
```Rust
pub enum ConstraintDomain {
  FirstRow,           // for boundary constraints against the first row
  LastRow,            // for boundary constraints against the last row
  EveryRow,           // for validity constraints
  EveryFrame(usize),  // for transition constraints; internal value is size of the frame
}
```
We don't need to use this now - but would make sense to do when we merge boundary and transition constraints in the same graph. Then, `ConstraintRoot` struct would look something like:
```Rust
struct ConstraintRoot {
    index: NodeIndex,
    domain: ConstraintDomain,
}
```

tohrnii:

ir/src/integrity_stmts/mod.rs
```diff
@@ -123,3 +124,30 @@ impl IntegrityStmts {
...
+    /// Returns the index of the entry node of the subgraph representing the constraint.
+    pub fn node_index(&self) -> &NodeIndex {
+        &self.index
+    }
+
+    /// Returns the row offset for the constraint which is the maximum of all the row offsets
+    /// accessed by the constraint.
+    pub fn offset(&self) -> usize {
+        self.offset
+    }
```
grjte, line 152: I think this row offset terminology may not be entirely clear from these comments, since I don't think we explain it fully anywhere in the code. Maybe we should include an example like "For example, if a constraint only accesses the trace in the current row then the row offset will be 0, but if it accesses the trace in both the current and the next rows then the row offset will be 1."

grjte (changes requested):
Looks good now! I left a couple small things inline that I think got lost among the bigger things in the previous review.

ir/src/integrity_stmts/mod.rs
grjte: > I think we need to define what is meant by "the row offsets for the constraint"

I still think this could be addressed better. I left a new comment inline

ir/src/integrity_stmts/graph.rs
```diff
@@ -91,13 +91,13 @@ impl AlgebraicGraph {
         symbol_table: &SymbolTable,
         expr: IntegrityExpr,
         variable_roots: &mut VariableRoots,
-    ) -> Result<(TraceSegment, NodeIndex), SemanticError> {
+    ) -> Result<(TraceSegment, NodeIndex, usize), SemanticError> {
```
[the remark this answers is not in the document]
grjte: Sure, that works!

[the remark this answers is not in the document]
tohrnii: Thanks. Done.

ir/src/integrity_stmts/mod.rs
tohrnii: Thanks. Done.

grjte (approved):
Thank you! I left one nit inline, but it's ok to leave it. I can make the adjustment when I move the boundary constraints into the graph for #106

ir/src/integrity_stmts/graph.rs
grjte: We'll add this enum in #106

PR #112, opened 2023-01-13
Add support for $main and $aux built-ins

### LINKED ISSUE #26: add support for $main and $aux built-ins
Currently, we have the built-in `$rand` which allows accessing a random value by index, e.g. `$rand[0]`.

It would be convenient to do the same thing with the `main` and `auxiliary` execution traces and allow columns to be accessed by index rather than exclusively by their column identifier.

This requires:

- [x] add lexing/parsing for `$main[n]`, `$aux[n]`, `$main[n]'`, and `$aux[n]'` where `n` is a number. This should only be allowed in transition constraint expressions. The AST needs to be updated to allow the new column representation as well.
- [x] update the IR to process and validate columns accessed this way when building the transition constraints graph

Overcastan:
Closes #26
Add support for access main and aux segments by column index: `$main[n]`, `$aux[n]`, `$main[n]'`, and `$aux[n]'` where `n` is a number.

### DIFF
--- a/air-script/tests/main.rs
+++ b/air-script/tests/main.rs
@@ -97,11 +97,12 @@ fn trace_col_groups() {
 }
 
 #[test]
-fn index_col_access() {
-    let generated_air = Test::new("tests/index_col_access/index_col_access.air".to_string())
-        .transpile()
-        .unwrap();
+fn indexed_trace_access() {
+    let generated_air =
+        Test::new("tests/indexed_trace_access/indexed_trace_access.air".to_string())
+            .transpile()
+            .unwrap();
 
-    let expected = expect_file!["index_col_access/index_col_access.rs"];
+    let expected = expect_file!["indexed_trace_access/indexed_trace_access.rs"];
     expected.assert_eq(&generated_air);
 }
--- a/air-script/tests/main.rs
+++ b/air-script/tests/main.rs
@@ -95,3 +95,13 @@ fn trace_col_groups() {
     let expected = expect_file!["trace_col_groups/trace_col_groups.rs"];
     expected.assert_eq(&generated_air);
 }
+
+#[test]
+fn index_col_access() {
+    let generated_air = Test::new("tests/index_col_access/index_col_access.air".to_string())
+        .transpile()
+        .unwrap();
+
+    let expected = expect_file!["index_col_access/index_col_access.rs"];
+    expected.assert_eq(&generated_air);
+}
--- a/ir/src/integrity_stmts/graph.rs
+++ b/ir/src/integrity_stmts/graph.rs
@@ -118,7 +118,7 @@ impl AlgebraicGraph {
                 Ok((trace_segment, node_index, CURRENT_ROW))
             }
             IntegrityExpr::IndexedTraceAccess(column_access) => {
-                self.insert_column_access(symbol_table, column_access)
+                self.insert_trace_access(symbol_table, column_access)
             }
             IntegrityExpr::Add(lhs, rhs) => {
                 // add both subexpressions.
@@ -197,11 +197,13 @@ impl AlgebraicGraph {
         }
     }
 
-    /// Adds a trace column to the graph and returns the node index and trace segment.
+    /// Adds a trace element access to the graph and returns the node index and trace segment.
     ///
     /// # Errors
-    /// Returns an error if the identifier is bigger than overall number of columns in segment.
-    fn insert_column_access(
+    /// Returns an error if:
+    /// - The identifier is greater than overall number of columns in segment.
+    /// - The segment is greater than the maximum number of segments.
+    fn insert_trace_access(
         &mut self,
         symbol_table: &SymbolTable,
         column_access: IndexedTraceAccess,
@@ -219,7 +221,7 @@ impl AlgebraicGraph {
         let segment_idx = column_access.segment_idx();
         if segment_idx > u8::MAX as usize {
             return Err(SemanticError::IndexOutOfRange(format!(
-                "Segment index {} is greater than 255",
+                "Segment index {} is greater than the maximum number of segments (255)",
                 segment_idx
             )));
         }
--- a/ir/src/integrity_stmts/graph.rs
+++ b/ir/src/integrity_stmts/graph.rs
@@ -3,8 +3,8 @@ use super::{
 };
 use crate::{symbol_table::IdentifierType, ExprDetails, VariableRoots, CURRENT_ROW, NEXT_ROW};
 use parser::ast::{
-    self, constants::ConstantType, Identifier, IntegrityExpr, IntegrityVariableType, MatrixAccess,
-    VectorAccess,
+    self, constants::ConstantType, Identifier, IndexedTraceAccess, IntegrityExpr,
+    IntegrityVariableType, MatrixAccess, VectorAccess,
 };
 
 // ALGEBRAIC GRAPH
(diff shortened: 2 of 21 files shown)


grjte:
@Overcastan can you rebase this please before we review?

parser/src/lexer/mod.rs
```diff
@@ -83,6 +83,14 @@ pub enum Token {
...
     Rand,
 
+    /// A reserved keyword for accessing main columns by index
+    #[token("$main")]
+    MainIndex,
+
+    /// A reserved keyword for accessing aux columns by index
+    #[token("$aux")]
+    AuxIndex,
+
```
grjte, line 93: 1. I would move this up 2 sections and put it next to the existing main/aux keywords. I know that I specified in the task that these should only be accessible in transition constraint expressions, but I think that's no longer true. I think we should probably allow using them for boundary constraints as well now. 
2. I would rename these to `MainAccess` and `AuxAccess` to be more consistent with our internal terminology in the AST and IR.
3. I would rename the existing `Main` and `Aux` to `MainDecl` and `AuxDecl` so that there's more clarity between those keywords and these new keywords.

parser/src/ast/integrity_constraints.rs
```diff
@@ -41,6 +41,9 @@ pub enum IntegrityExpr {
     /// name of the matrix and indices of the element to access.
     MatrixAccess(MatrixAccess),
     Next(TraceAccess),
+    /// Represents a column in the execution trace, e.g. `$main[2]`. [ColumnAccess] contains the
+    /// segment index and column index to be accessed.
+    ColumnAccess(ColumnAccess),
```
grjte, line 46: We now have 2 different ways of representing a trace access in the AST which are very different from each other. I'd prefer to have them be more similar.

I propose the following changes:
- rename the `TraceAccess` struct to `NamedTraceAccess`
- rename the `ColumnAccess` struct to `IndexedTraceAccess`, and change the enum variant here to `TraceAccess`

This will end up with something like the following:
```
   /// TODO: please add a comment here when you update to `NamedTraceAccess`
   Next(NamedTraceAccess),
    /// Represents accessing a column in the execution trace at a particular row offset, e.g. `$main[2]`.
    /// [IndexedTraceAccess] contains the segment index, column index within the segment, and row 
    /// offset to be accessed.
    TraceAccess(IndexedTraceAccess),
```

parser/src/ast/mod.rs
```diff
@@ -180,3 +180,34 @@ impl MatrixAccess {
...
+    }
+
+    pub fn column_idx(&self) -> usize {
+        self.column_idx
+    }
+
+    pub fn row_offset(&self) -> usize {
+        self.row_offset
+    }
+}
```
grjte, line 213: 1. As noted, I think this should be renamed to `IndexedTraceAccess`.
2. I also think it should be moved up in the file so it's right below the `NamedTraceAccess`
3. The comment at the top should also be adjusted for accuracy, since it doesn't mention the row offset

ir/src/integrity_stmts/graph.rs
```diff
@@ -184,6 +187,34 @@ impl AlgebraicGraph {
...
+                column_access.column_idx(),
+                column_access.segment_idx(),
+                symbol_table.segment_width()[column_access.segment_idx()]
+            )));
+        }
+        let trace_access = TraceAccess::new(
+            column_access.segment_idx() as u8,
+            column_access.column_idx(),
+            column_access.row_offset(),
+        );
```
grjte, line 213: Since these contain all the same information, can you implement the `From` trait instead?

Also, we should probably check the size of segment_idx somewhere before casting it.

ir/src/lib.rs
```diff
@@ -146,6 +149,10 @@ impl AirIR {
         &self.constants
     }
 
+    pub fn segment_width(&self) -> &Vec<u16> {
+        &self.segment_width
+    }
+
```
grjte, line 155: nit: I think I'd prefer `segment_widths` instead, to be clearer that it has the width of all the segments.

ir/src/lib.rs
```diff
@@ -46,6 +46,7 @@ const NEXT_ROW: usize = 1;
 pub struct AirIR {
     air_name: String,
     constants: Constants,
+    segment_width: Vec<u16>,
```
grjte, line 49: nit: I think I'd prefer `segment_widths` instead, to be clearer that it has the width of all the segments.

ir/src/symbol_table.rs
```diff
@@ -65,7 +65,10 @@ impl SymbolTable {
     /// Sets the number of trace segments to the maximum of `num_trace_segments` and the provided
     /// trace segment identifier, which is indexed from zero.
     fn set_num_trace_segments(&mut self, trace_segment: TraceSegment) {
-        self.num_trace_segments = self.num_trace_segments.max((trace_segment + 1).into())
+        // self.num_trace_segments = self.num_trace_segments.max((trace_segment + 1).into())
+        if trace_segment >= self.segment_width.len() as u8 {
+            self.segment_width.resize(trace_segment as usize + 1, 0);
+        }
```
grjte, line 71: The method should be renamed if the functionality is changing. The comment should be similarly updated. Also, the commented-out code can be deleted.

It's not clear to me what's happening here. Once these things are updated, I may have other thoughts.

grjte (changes requested):
Thank you @Overcastan! I like your approach overall. I left a few thoughts and may have a bit more feedback once these changes are made.

More generally:
- can you please add some parsing tests?
- can you add a minimal integration test?

Also, I think we can allow accessing `$main` and `$aux` in boundary constraints, but this can be done in a separate PR.

tohrnii (approved):
Thank you @Overcastan. I don't have much to add other than the comments that @grjte pointed out. Just one clarification for parser tests, please add tests for both lexer and parser. Other than that one question is if we should allow named constants as indices but this should probably be another issue as we will probably need to do that for some other things as well.

ir/src/symbol_table.rs
Overcastan: The idea was to combine information about number of segments and number of columns in each segment. Since now we have a `segment_widths` vector, the number of segments can be obtained as length of this vector. And if we want to update the segments number we should resize `segment_widths` vector (but only if new segment is greater than current biggest one). I forgot to update the comment, but I didn't change the name of the function intentionally so that the operation with `SymbolTable` would remain the same, because the function continued to return the exact same value. But I can change that if you think it's better to update the name.

grjte: Yes, please update the name. I don't think the name reflects the functionality anymore.

air-script/tests/main.rs
```diff
@@ -95,3 +95,13 @@ fn trace_col_groups() {
...
+
+#[test]
+fn index_col_access() {
+    let generated_air = Test::new("tests/index_col_access/index_col_access.air".to_string())
+        .transpile()
+        .unwrap();
+
+    let expected = expect_file!["index_col_access/index_col_access.rs"];
+    expected.assert_eq(&generated_air);
+}
```
grjte, line 107: I think we should rename this test and the associated files/directories to `indexed_trace_access`

grjte (changes requested):
Looks good, thanks @Overcastan! Most of my comments are nits at this point. 

In addition to addressing the inline comments, can you please:
1. rebase from `next` to integrate the latest changes to the IR
2. clean up the commit log (rebase to squash the refactor, and create a "test" commit where the tests are added)

grjte (approved):
Looks good, thank you!

PR #115, opened 2023-01-19
Add parens based on operator precedence

tohrnii:
This PR adds a fix for the redundant parentheses in expressions. Here we check the precedence between the two operands and only add parentheses when the operand precedence is lower.


Overcastan (approved):
Looks good to me, thank you @tohrnii !

tohrnii:
One question here is how should we handle `Exp` operation. Should the base always have parentheses or only non-leaf expressions have them.

Al-Kindi-0 (approved):
Looks good, thank you!

grjte:
> One question here is how should we handle `Exp` operation. Should the base always have parentheses or only non-leaf expressions should have them.

Ideally, parentheses should only be there in cases where they're syntactically necessary for the correctness of the calculation

tohrnii:
> Ideally, parentheses should only be there in cases where they're syntactically necessary for the correctness of the calculation.

> However, it's not worth spending a lot of time on this right now, so the simplest solution is fine.

Makes sense. I've skipped it for now since it requires either introducing a little redundancy or making the helper function slightly more complicated so just went with keeping it simple for now.

tohrnii:
Superseded by #155

Issue #98, opened 2022-12-06
Consider putting commonly used structs and consts in a common crate like air-script-core


tohrnii:
> It does feel a bit weird that we use structs from the parser in the IR, but I guess we do it in a few other places already. Maybe there should be a air-script-core or air-script-common crate which defines such commonly used structs. (but this would be for another PR)

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/air-script/pull/94#discussion_r1040591555_

tohrnii:
Closed by #116

PR #116, opened 2023-01-19
move commonly used structs/types to common crate + refactor

grjte:
This PR addresses #98 in preparation for #106.

As part of moving the shared structs to the new `air-script-core` crate, I also made a the following changes to reduce complexity and duplication.

- added a `row_offset` to `NamedTraceAccess` to make it behave more like `IndexedTraceAccess`.
- replaced `Next` expression variant with `NamedTraceAccess`. Replacing this allowed for unification of `BoundaryExpr` and `TransitionExpr` into a single `Expression` type. The reasoning for why nothing is lost here in terms of safety is that the `BoundaryExpr` from the AST could already have included named trace columns referencing the "current" row, which need to be checked for in the IR.
- replaced `BoundaryVariable` and `TransitionVariable` types with a single `Variable` struct. The only difference previously was that they were referencing `BoundaryExpr` and `TransitionExpr`. With the unification of the `Expression` type, there is no need for these to be different.
- replaced `BoundaryVariableType` and `TransitionVariableType` with a single `VariableType` struct, with the same reasoning as above.
- merged the `IndexedTraceAccess` from the AST and the `TraceAccess` defined in the IR.


ir/src/boundary_stmts.rs
```diff
@@ -176,33 +176,30 @@ impl BoundaryStmts {
     }
 }
 
-/// Recursively validates the BoundaryExpression.
+/// Recursively validates the Expressionession.
```
tohrnii, line 179: nit: There's a typo.

parser/src/ast/boundary_constraints.rs
```diff
@@ -36,8 +36,8 @@ impl BoundaryConstraint {
     }
 
     /// Returns a clone of the constraint's value expression.
```
tohrnii, line 38: nit: This comment should be changed if we no longer return a clone of the value.

tohrnii (approved):
Thank you @grjte, the whole code structure seems much cleaner now. I just left a couple of very minor nits. One question: If a boundary expression contains a trace access, is the only place we check it is during codegen or did I miss something? If it's just during codegen then we will probably need to add a check sooner when we target other backends?

ir/src/integrity_stmts/graph.rs
```diff
@@ -117,10 +114,13 @@ impl AlgebraicGraph {
                 let node_index = self.insert_op(Operation::RandomValue(index));
                 Ok((trace_segment, node_index, CURRENT_ROW))
             }
-            IntegrityExpr::IndexedTraceAccess(column_access) => {
+            Expression::IndexedTraceAccess(column_access) => {
                 self.insert_trace_access(symbol_table, column_access)
```
Overcastan, line 118: If we have `insert_named_trace_access`, may be we should rename this function to `insert_indexed_trace_access` to maintain consistency?

Overcastan (approved):
Thank you @grjte Looks really good! Just one small question inline

ir/src/integrity_stmts/graph.rs
grjte: Agreed! I was going to do it in a different PR, to minimize the amount of changes in this one. (I was trying to keep this to strictly the changes required for the new structure)

grjte:
> One question: If a boundary expression contains a trace access, is the only place we check it is during codegen or did I miss something? If it's just during codegen then we will probably need to add a check sooner when we target other backends?

We'll need to add this check to the IR, but I wanted to keep this PR as minimal as possible given that the refactor already touched a lot of things. I'm planning to handle this validation after adding the boundary constraints to the graph

ir/src/boundary_stmts.rs
grjte: Fixed, thanks!

parser/src/ast/boundary_constraints.rs
grjte: Great catch, thanks!

Issue #87, opened 2022-11-28
List Comprehension


tohrnii:
It would be nice to have python style [list comprehension](https://www.geeksforgeeks.org/python-list-comprehension/) supported in AirScript. For eg.

```
trace_columns:
  main: [a, b, c[4]]

# raise value in the current row to power 7
let x = [col^7 for col in c]

# raise value in the next row to power 7
let y = [col'^7 for col in c]
```

In both cases, the result would be a vector of 4 elements.

We could also add support for iterating over two lists simultaneously like:
```
trace_columns:
  main: [a, b, c[4], d[4]]

let diff = [x - y for (x, y) in (c, d)]
```

We could also add support for iterating and enumerating like:

```
trace_columns:
  main: [a, b, c[4]]

let x = [2^i * c for (i, c) in (0..3, c)]
```

We could also support list folding. There are two possible syntax:

```
trace_columns:
  main: [a, b, c[4], d[4]]

# compute sum of products of values in c and d
let x += c * d for (c, d) in (c, d)

# compute product of sums of values in c and d
let y *= c + d for (c, d) in (c, d)
```
OR

```
trace_columns:
  main: [a, b, c[4], d[4]]

# compute sum of products of values in c and d
let x = sum([c * d for (c, d) in (c, d)])

# compute product of sums of values in c and d
let y = prod([c + d for (c, d) in (c, d)])
```
_The syntax was proposed by @bobbinth [here](https://github.com/maticnetwork/miden/discussions/254#discussioncomment-3293836)._

tohrnii:
Question:
Should we allow users to access the length of the array for iterators? So, for eg something like,
```
trace_columns:
  main: [a, b, c[4]]

let x = [2^i * c for (i, c) in (c.len(), c)]
```

instead of 
```
trace_columns:
  main: [a, b, c[4]]

let x = [2^i * c for (i, c) in (0..3, c)]
```

tohrnii:
Another question: Do we have consensus on using `0..n` for ranges instead of pythonic `range(0, n)`?

bobbinth:
> Should we allow users to access the length of the array for iterators/ enumeration or in general?

I can't think of cases where we'd need something like this - but I might not be thinking of everything here. In general, my preference would not to add it until we have a clear use-case. So, unless it is super-trivial to add, I'd probably skip it for now.

> Do we have consensus on using `0..n` for ranges instead of pythonic `range(0, n)`?

[...]

PR #119, opened 2023-01-24
Add support for list comprehensions to parser

tohrnii:
Partly addressing #87. This PR adds list comprehension support to the parser. This PR should be merged after #120.

### DIFF
--- a/air-script-core/src/variable.rs
+++ b/air-script-core/src/variable.rs
@@ -1,4 +1,4 @@
-use super::{Expression, Identifier};
+use super::{Expression, Identifier, Range};
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Variable {
@@ -25,4 +25,38 @@ pub enum VariableType {
     Scalar(Expression),
     Vector(Vec<Expression>),
     Matrix(Vec<Vec<Expression>>),
+    ListComprehension(ListComprehension),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ListComprehension {
+    expression: Expression,
+    context: Vec<(Identifier, Iterable)>,
+}
+
+impl ListComprehension {
+    /// Creates a new list comprehension.
+    pub fn new(expression: Expression, context: Vec<(Identifier, Iterable)>) -> Self {
+        Self {
+            expression,
+            context,
+        }
+    }
+
+    /// Returns the expression that is evaluated for each member of the list.
+    pub fn expression(&self) -> &Expression {
+        &self.expression
+    }
+
+    /// Returns the context of the list comprehension.
+    pub fn context(&self) -> &Vec<(Identifier, Iterable)> {
+        &self.context
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum Iterable {
+    Identifier(Identifier),
+    Range(Range),
+    Slice(Identifier, Range),
 }
--- a/air-script-core/src/variable.rs
+++ b/air-script-core/src/variable.rs
@@ -49,7 +49,7 @@ impl ListComprehension {
     }
 
     /// Returns the context of the list comprehension.
-    pub fn context(&self) -> &Vec<(Identifier, Iterable)> {
+    pub fn context(&self) -> &[(Identifier, Iterable)] {
         &self.context
     }
 }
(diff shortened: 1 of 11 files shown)


parser/src/lexer/tests/list_comprehension.rs
```diff
@@ -0,0 +1,70 @@
+use super::{expect_valid_tokenization, Token};
+
+// LIST COMPREHENSION VALID TOKENIZATION
+// ================================================================================================
+
+#[test]
+fn single_list_comprehension() {
+    let source = "let y = [x for x in x]";
+    let tokens = vec![
```
grjte, line 9: I'm not sure about the term "single". It might be clearer after we switch from "double" to "pair", but I think I'd prefer "item" or "individual". Let's see what others think

parser/src/parser/grammar.lalrpop
```diff
@@ -318,6 +322,45 @@ IndexedTraceAccess: IndexedTraceAccess = {
...
+        (Members::Single(..), Iterables::Double(..)) => Err(ParseError::User {
+            error: Error::ParseError(
+                InvalidListComprehension("For a single member, a single iterable is required".to_string())
+            )
+        }),
+        (Members::Single(..), Iterables::Range(..)) => Err(ParseError::User {
+            error: Error::ParseError(
+                InvalidListComprehension("For a single member, a single iterable is required".to_string())
+            )
+        }),
```
grjte, line 338: Can you combine these match arms?

grjte (changes requested):
Thanks @tohrnii! I left a few small things inline. Let's see what @bobbinth thinks of this approach as well, though

parser/src/parser/grammar.lalrpop
tohrnii: Done.

tohrnii:
One other clarification, I missed this earlier. @bobbinth you wrote a constraint in the evaluator functions example for rescue prime constraints:

```
  enf x = y for (x, y) in (s1, s2)
```

I'm not sure what this means. Is there a typo here?

tohrnii:
Based on an example by @bobbinth (in rescue prime constraints), we would need to be able to do something like:
```
  enf x = y for (x, y) in (s1, s2)
```
However, the PR in the current form doesn't support these kinds of expressions. I'll update this PR to allow such expressions.

parser/src/lexer/tests/list_comprehension.rs
Al-Kindi-0: I think "single" and "pair" work well together.

air-script-core/src/variable.rs
```diff
@@ -25,4 +25,51 @@ pub enum VariableType {
...
+    Single(Identifier),
+    Double(Identifier, Identifier),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum Iterables {
+    Single(Identifier),
+    Double(Identifier, Identifier),
+    Range(Range, Identifier),
 }
```
bobbinth, line 75: Is this a temporary simplification? Ideally, we should not put a limit on the number of iterable lists and be able to do something like:
```
for (x, y, z) in (a, b[1..4], 0..3)
```
And I'm not sure this model will support this.

bobbinth (commented):
Looks good to me! Thank you! I did leave one comment inline: basically, not sure if the current implementation is a simplification of where we want to be. Fine if so, but if not - I think we need to have a more general structure which could handle many iterable lists.

air-script-core/src/variable.rs
tohrnii: Makes sense. It wasn't meant to be a temporary simplification as I thought this was the extent of the kinds of expressions that we wanted to allow. However, if we want to allow more complicated expressions here, you are right that we should implement a more generic structure and it probably makes sense to do it now instead of coming back to it later so I'll modify this PR.

bobbinth: I think at the parsing level the extra complexity would still be pretty minimal. We could have an `Iterable` enum which would describe different variants of what an iterable could be (i.e., identifier, range, identifier + range). And then `ListComprehension` struct could look something like:
```Rust
pub struct ListComprehension {
    expression: Expression,
    context: Vec<(Identifier, Iterable)>,
}
```
Where tuples stored in `context` contain bindings of names to their respective interables. For example, in `for (x, y, z) in (a, b[1..4], 0..3)`, the context would be:
```
(x, Iterable(a)), (y, Iterable(a, 1..4)), (z, Iterable(0..3))
```

bobbinth (approved):
Looks good! Thank you! I left one small non-blocking comment inline.

Overcastan (approved):
Looks great to me, thank you @tohrnii!

grjte (approved):
Looks great, thanks @tohrnii!

PR #120, opened 2023-01-24
Allow expressions as exponents in exp expressions

tohrnii:
This PR adds support for expressions as exponents for `Exp` expressions at the parser level but doesn't yet handle them at the IR and codegen level. This is required to write some constraints more concisely.

### DIFF
--- a/codegen/winterfell/src/air/boundary_constraints.rs
+++ b/codegen/winterfell/src/air/boundary_constraints.rs
@@ -167,7 +167,11 @@ impl Codegen for Expression {
                 )
             }
             Self::Exp(lhs, rhs) => {
-                format!("({}).exp({})", lhs.to_string(ir, is_aux_constraint), rhs)
+                if let Self::Const(rhs) = **rhs {
+                    format!("({}).exp({})", lhs.to_string(ir, is_aux_constraint), rhs)
+                } else {
+                    todo!()
+                }
             }
             _ => panic!("boundary constraint expressions cannot reference the trace"),
         }
--- a/ir/src/integrity_stmts/graph.rs
+++ b/ir/src/integrity_stmts/graph.rs
@@ -171,7 +171,12 @@ impl AlgebraicGraph {
                 let (lhs, trace_segment, domain) =
                     self.insert_expr(symbol_table, *lhs, variable_roots)?;
                 // add exponent subexpression.
-                let node_index = self.insert_op(Operation::Exp(lhs, rhs as usize));
+                let node_index = if let Expression::Const(rhs) = *rhs {
+                    self.insert_op(Operation::Exp(lhs, rhs as usize))
+                } else {
+                    todo!()
+                };
+
                 Ok((node_index, trace_segment, domain))
             }
         }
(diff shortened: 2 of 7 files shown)


tohrnii:
Closing this since these changes have been included in #119.

parser/src/parser/tests/arithmetic_ops.rs
```diff
@@ -225,15 +225,6 @@ fn err_closing_paren_without_opening_paren() {
...
 
-#[test]
-fn non_const_exp_fail() {
-    // Should fail if the exponent is not a constant
-    let source = "
-    integrity_constraints:
-        enf clk'^a = 1";
-    build_parse_test!(source).expect_unrecognized_token();
-}
-
```
grjte, line 236: Instead of deleting this, let's replace it with a test for the non-const exponent

ir/src/integrity_stmts/graph.rs
```diff
@@ -171,7 +171,12 @@ impl AlgebraicGraph {
                 let (lhs, trace_segment, domain) =
                     self.insert_expr(symbol_table, *lhs, variable_roots)?;
                 // add exponent subexpression.
-                let node_index = self.insert_op(Operation::Exp(lhs, rhs as usize));
+                let node_index = if let Expression::Const(rhs) = *rhs {
+                    self.insert_op(Operation::Exp(lhs, rhs as usize))
+                } else {
+                    todo!()
+                };
+
```
grjte, line 179: what's the reasoning for not completing this now? do we need to do something other than call `insert_expr` again for the `rhs`? I'd prefer to implement it now unless there's something complicated that I'm overlooking

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -167,7 +167,11 @@ impl Codegen for Expression {
                 )
             }
             Self::Exp(lhs, rhs) => {
-                format!("({}).exp({})", lhs.to_string(ir, is_aux_constraint), rhs)
+                if let Self::Const(rhs) = **rhs {
+                    format!("({}).exp({})", lhs.to_string(ir, is_aux_constraint), rhs)
+                } else {
+                    todo!()
+                }
```
grjte, line 174: I requested implementing the IR change now, but leaving this `todo` is fine, as this whole section changes with my current PR #126, so we can update it during rebase

grjte (changes requested):
Thanks @tohrnii, just a few small things

ir/src/integrity_stmts/graph.rs
tohrnii: @grjte I haven't changed the signature of the `Exp` variant of the `Operation` enum yet because we will have to either evaluate exponent expressions or limit them to be able to calculate degree of the operation correctly. I was thinking to handle it with adding list comprehension to IR PR.

parser/src/parser/tests/arithmetic_ops.rs
tohrnii: Thanks. Done.

ir/src/integrity_stmts/graph.rs
grjte: Ok sure, makes sense.

grjte (approved):
Looks good, thanks @tohrnii! 

edit: As mentioned above, can you add a TODO in the `Operation` enum in the IR before merging as well?

ir/src/integrity_stmts/graph.rs
grjte: Actually, it can you add a TODO in the `Operation` enum as well before merging this?

PR #126, opened 2023-01-26
Add boundary constraints to the constraint graph

grjte:
This PR addresses part of #106, by moving boundary constraints into the `AlgebraicGraph` alongside the integrity constraints, and updating the codegen and tests accordingly.

It includes the following changes:

- [x] add the boundary statements to the graph
- [x] refactor to graph handling to simplify the combined handling of integrity + boundary constraints
- [x] update winter codegen to work with the new boundary constraints
- [x] update the tests to reflect the changes (primarily constraint re-ordering) and skip the tests which are failing due to the pending changes listed below

There are several things that yet to be completed, but I'd like to tackle these in a subsequent PR since this is already quite large:
 
- restore functionality of public inputs, by adding them to the graph, then restore the related tests
- restore boundary constraint validation in the IR, then restore the related tests
- add support for boundary variables
- verify that all cases are captured and add any additional validation/testing as required


air-script/tests/variables/variables.air
```diff
@@ -14,7 +14,7 @@ periodic_columns:
     k0: [1, 1, 1, 1, 1, 1, 1, 0]
 
 boundary_constraints:
-    enf a.first = stack_inputs[0]
+    enf a.first = 0
```
grjte, line 17: I changed this to keep the public inputs usage in the public inputs test. Once we add variables for the boundary constraints, we can test them here as well

ir/src/constraints/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+/// variables used in the integrity constraints, and a matrix for each constraint type (boundary,
+/// validity, transition), where each index contains a vector of the constraint roots for all the
+/// constraints of that type against the segment of the trace corresponding to that index. For
+/// example, transition constraints against the main execution trace, which is trace segment 0, will
+/// be specified by a vector in transition_constraints[0] containing a [ConstraintRoot] in the graph
+/// for each constraint against the main trace.
+#[derive(Default, Debug)]
+pub(super) struct Constraints {
+    /// Constraint roots for all boundary constraints against the execution trace, by trace segment,
+    /// where boundary constraints are any constraints that apply either the first or last row of
```
Al-Kindi-0, line 74: nit: to the first

ir/src/constraints/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+                unimplemented!("TODO: add support for boundary variables")
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Takes two expressions which are expected to be equal and merges them into a constraint (a
+    /// subtree in the graph that must be equal to zero for a particular domain). The constraint is
+    /// is then saved in the appropriate constraint list (boundary, validity, or transition).
```
Al-Kindi-0, line 282: nit: repeated "is"

Al-Kindi-0 (commented):
Looks great to me, thanks @grjte !
Just some minor nits from my side.

ir/src/constraints/graph.rs
```diff
@@ -58,130 +53,271 @@ impl AlgebraicGraph {
...
+                matrix_access,
+                variable_roots,
+                default_domain,
+            ),
             Expression::Rand(index) => {
                 // constraint target for random values defaults to the second trace segment.
                 // TODO: make this more general, so random values from further trace segments can be
                 // used. This requires having a way to describe different sets of randomness in
                 // the AirScript syntax.
                 let trace_segment = 1;
```
tohrnii, line 139: nit: Should we define an alias `AUX_TRACE_SEGMENT`?

tohrnii (approved):
Thank you @grjte, looks good to me. I left a few minor nits inline.

Overcastan (approved):
Looks great to me, thank you @grjte!

codegen/winterfell/src/air/mod.rs
```diff
@@ -1,10 +1,16 @@
...
+use air_script_core::{Constant, ConstantType};
+use ir::{
+    constraints::{ConstantValue, ConstraintDomain, Operation},
+    IntegrityConstraintDegree, NodeIndex, PeriodicColumns,
+};
 
 mod constants;
 use constants::add_constants;
 
+mod ir_types;
```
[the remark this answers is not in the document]
grjte: Yeah, I don't like this name either, I just didn't have any better ideas at the time

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -55,125 +31,51 @@ pub(super) fn add_fn_get_aux_assertions(impl_ref: &mut Impl, ir: &AirIR) {
...
+        ElemType::Felt
+    } else {
+        ElemType::E
+    };
+
     // declare the result vector to be returned.
-    get_aux_assertions.line("let mut result = Vec::new();");
+    let last_row = "last_step";
+    func_body.line("let mut result = Vec::new();");
+    func_body.line(format!("let {last_row} = self.last_step();"));
```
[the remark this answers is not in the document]
grjte: I thought about that - it's how we did it before. It seemed a bit more complicated and maybe unnecessary, but I'll take another look. I don't really have a strong opinion either way

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -55,125 +31,51 @@ pub(super) fn add_fn_get_aux_assertions(impl_ref: &mut Impl, ir: &AirIR) {
...
+
+    // add the boundary constraints
+    for constraint in ir.boundary_constraints(trace_segment) {
+        let (trace_access, expr_root) = ir
```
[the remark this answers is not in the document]
grjte: It's returning the root `NodeIndex` in the graph where the expression starts, which is why I called it `expr_root`. I don't have a strong opinion, but I think `expr` is less accurate

codegen/winterfell/src/air/mod.rs
```diff
@@ -106,38 +112,22 @@ fn add_fn_new(impl_ref: &mut Impl, ir: &AirIR) {
...
-        .constraint_degrees(1)
-        .iter()
-        .map(|degree| degree.to_string(ir, true))
-        .collect::<Vec<_>>();
-    new.line(format!(
-        "let aux_degrees = vec![{}];",
-        aux_degrees.join(", ")
-    ));
+    // define the integrity constraint degrees of the aux trace `aux_degrees`.
+    add_constraint_degrees(new, ir, 1, "aux_degrees");
```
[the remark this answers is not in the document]
grjte: I'd prefer not to. I like seeing each step for each trace segment. That's how the codegen is set up in general, so it feels clearer and more explicit to me

ir/src/constraints/graph.rs
grjte: Yeah, I think that makes sense. In the longer run we should make this more general, but that's very low priority, so we probably won't get there for a while

codegen/winterfell/src/air/boundary_constraints.rs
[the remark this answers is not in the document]
tohrnii: Makes sense. I was thinking that in this context it's clear that we are talking about the root and `expr` is more concise but I can see the argument to be explicit to make it clear.

codegen/winterfell/src/air/mod.rs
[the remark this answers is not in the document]
grjte: I moved `ElemType`, but otherwise I haven't changed this. Basically, the problem with `codegen_trait` is that we actually have different codegen traits in a couple places (for constants and for periodic columns).  The problem with naming it something codegen-related is that everything in this crate is codegen-related, so I'm not sure that's useful. The idea of this name is that this is where the codegen is defined for the core types of the `ir`, but that's not even fully true. What is true is that it's where there's codegen that's related to the graph/constraints, so maybe I could change it to something about those?

Other thoughts/suggestions would be great

grjte:
Thanks for your reviews @Al-Kindi-0 @Overcastan and @tohrnii! I made the suggested changes, except for one point of discussion raised by @tohrnii (which it would be great to get your thoughts on).

Since this is a very significant change, let's wait for the review from @bobbinth before we merge.

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -55,125 +31,57 @@ pub(super) fn add_fn_get_aux_assertions(impl_ref: &mut Impl, ir: &AirIR) {
...
+            .constraint_graph()
+            .split_boundary_constraint(constraint.node_index())
+            .expect("failed to split boundary constraint");
+        debug_assert!(trace_access.trace_segment() == trace_segment);
+
+        // declare the last row variable, if needed.
+        if !last_row_declared && constraint.domain() == ConstraintDomain::LastRow {
+            func_body.line(format!("let {last_row} = self.last_step();"));
+            last_row_declared = true;
```
bobbinth, line 66: Thought for the future: it seems like determining if the `last_row` variable was declared or not adds quite a bit of complexity. Should we just always use `self.last_step()` instead? If it has `inline(always)` attribute, should theoretically give the same result.

ir/src/constraints/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+    /// rows.
+    transition_constraints: Vec<Vec<ConstraintRoot>>,
+
+    /// A directed acyclic graph which represents all of the constraints and their subexpressions.
+    graph: AlgebraicGraph,
+
+    /// Variable roots for the variables used in integrity constraints. For each element in a
+    /// vector or a matrix, a new root is added with a key equal to the [VariableValue] of the
+    /// element.
+    variable_roots: VariableRoots,
```
bobbinth, line 93: As far as I can tell, this is a helper map which is used only during the construction of the graph and is not really needed after the graph is constructed. If so, it may be good to find a better way to handle it in the future so that once constructed `Constraints` does not need to retain any extra info.

ir/src/constraints/graph.rs
```diff
@@ -58,130 +55,271 @@ impl AlgebraicGraph {
...
-                lhs_base.max(rhs_base)
-            }
+    /// Given a node index that is expected to be the root index of a boundary constraint, returns
+    /// the [IndexedTraceAccess] representing the trace segment and column against which the
+    /// boundary constraint expression must hold, as well as the node index that represents the root
+    /// of the constraint expression that must equal zero during evaluation.
+    pub fn split_boundary_constraint(
+        &self,
+        index: &NodeIndex,
+    ) -> Result<(&IndexedTraceAccess, &NodeIndex), SemanticError> {
```
bobbinth, line 65: Seems like this is used only in Winterfell codegen, and I'm not sure if this will be relevant in other codgens (or if it will be relevant in the same way). Maybe the logic should go into Winterfell codegen rather than being here?

bobbinth (approved):
Looks great! Thank you! I reviewed everything except for tests.

I think the structure is much cleaner than before. My comments are mostly for the future except fro the `split_boundary_constraint` comment. But even that one is non-blocking.

codegen/winterfell/src/air/mod.rs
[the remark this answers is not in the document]
tohrnii: > What is true is that it's where there's codegen that's related to the graph/constraints, so maybe I could change it to something about those?

Makes sense. So, maybe just `graph` (seems to be inline with naming of other files)? I don't have a strong opinion though.

codegen/winterfell/src/air/mod.rs
```diff
@@ -17,6 +23,15 @@ use boundary_constraints::{add_fn_get_assertions, add_fn_get_aux_assertions};
 mod transition_constraints;
 use transition_constraints::{add_fn_evaluate_aux_transition, add_fn_evaluate_transition};
 
+// HELPER TYPES
+// ================================================================================================
+
+#[derive(Debug, Clone, Copy)]
+pub enum ElemType {
+    Felt,
+    E,
```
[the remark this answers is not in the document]
grjte: I was wondering about something along these lines as well, and I like this better, thanks!

codegen/winterfell/src/air/ir_types.rs
```diff
@@ -0,0 +1,116 @@
...
+            Operation::Mul(l_idx, r_idx) => {
+                let lhs = l_idx.to_string(ir, elem_type);
+                let rhs = r_idx.to_string(ir, elem_type);
+                format!("({lhs}) * ({rhs})")
+            }
+            Operation::Exp(l_idx, r_idx) => {
+                let lhs = l_idx.to_string(ir, elem_type);
+                match elem_type {
+                    ElemType::Felt => format!("({lhs}).exp(Felt::new({r_idx}))"),
+                    ElemType::E => format!("({lhs}).exp(E::PositiveInteger::from({r_idx}_u64))"),
```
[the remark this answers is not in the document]
grjte: Agreed. I was trying to minimize changes in this PR. I've opened #130 to track this.

Are you thinking this unrolling should be done in the IR or in the winterfell codegen? (It seems to make sense as a future optimization that could be done in the IR)

codegen/winterfell/src/air/boundary_constraints.rs
grjte: I'm happy to update this now. I actually originally declared `last_row` in all cases specifically to avoid this complexity, but this solution seems like the best option

ir/src/constraints/mod.rs
grjte: Yes, this is true. I think there are still a few refactors that could be done, but I wanted to keep this PR as minimal as possible. Once the remaining issues related to this work are addressed this is worth revisiting (or possibly while addressing those)

codegen/winterfell/src/air/mod.rs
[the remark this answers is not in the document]
grjte: Yeah, I think there might not be a great answer, but `graph` does seem better.

ir/src/constraints/graph.rs
grjte: Yeah, I think you're right that it's better there. I've moved it.

Issue #86, opened 2022-11-28
Add section for random values and allow declaring names for accessing random values


tohrnii:
For some backends it may be useful to have a `random_values` section. This section would enable naming/grouping of random values similar to how trace columns section allows for trace columns. The syntax for this section could look as follows:

```
random_values:
    rand: [15]
```

The above specifies that there should be 15 random available for use in auxiliary trace constraints, and that these random values will be available under variable $rand.

Similar to trace columns, we could name various values as follows:

```
random_values:
    rand: [a, b]
```

The above specifies that a and b names could be used to refer to specific random values, and also that $rand variable could be used to refer to the entire vector.

_Originally posted by @bobbinth in #53._

grjte:
A couple of questions here:

1. Do we still want to have a default $rand, or are we replacing that by this section so that they must always be declared? (if we're replacing it, we'll probably want to add some checks in the IR to make sure all sizes and accesses are declared / legal)
2. Do we still want to use the '$' now that we're naming these? If so, I think we should also access the named array values that way, i.e., by `$rand[0]` and `$a` `$b`.

@bobbinth what did you have in mind when you wrote this comment?

bobbinth:
I was thinking that this section would be optional - but I can also see an argument for making it mandatory. If we do keep it optional, I think we should assume `rand` as the default value. In the future, we could allow people to specify their own names (e.g., `alphas` instead of `rand`).

Regarding `$`, I was thinking this could be our general way of addressing variables declared in such sections. So, for example we have `$main` and `$aux`. `$rand` would be treated the same way.

So, in this example:
```
random_values:
    rand: [a, b]
```
We could access the first random value simply as `a` or as `$rand[0]`.

Overcastan:
I have a couple of questions: 
1. Do we want to allow more than one array in the section? 
2. Do we want to create array(s) according to the trace segments, or we want to make random values independent from trace segments?

grjte:
@Overcastan great question. This comes down to how "random values" work. The proof protocol works in rounds. The prover sends a commitment to the verifier, then the verifier sends some randomness, then the prover sends another commitment, then the verifier could se

Provers sends a commitment to the main trace (segment 0)
Verifier sends some randomness
Prover uses the randomness to commit to the aux trace (segment 1) & sends it to the verifier

In MidenVM and Winterfell, this is where we stop. Therefore at the moment, there's only one set of randomness, and that's the randomness which is used when building the "segment 1" trace.

[...]

PR #128, opened 2023-01-27
Individual section and named access for random values 

Overcastan:
This PR addresses #86.
It includes the following changes:
- Added individual section for random values declaration. Random values array could be created in forms of:
   - Constant length list, e.g. `rand: [16]`.
   - Array of identifiers. Vector type identifiers are also supported, e.g. `rand: [a, b[12], c]`.
- Added possibility to access random values by declared name. For now it is two main approaches to accessing a random value:
   - Indexed: `$rand[i]` — could be used with both variants of random values array declarations. If an index is referencing to the array type named random value (e.g. `b[12]`), this accessor would be equivalent to its 0'th index (`b[0]`).
   - Named: `a` — could be used only with array of identifiers. If `a` is vector type identifier, `a` call would be equivalent to `a[0]`.
   
What would be done in the future PRs (maybe we (I) should create new issues):
- Access random value with `$` and custom array name. For now it is possible to create random values array with custom name, but accessing its values could be done only with `$rand` call.
- Add internal index of array type random value (`i` in `b[i]`) to the code generation. For now we store this index, but it is not used in the code generation.

Problems:
- For now boundary constraints store random values as `Expression::Rand` only if we use indexed `$rand[i]` accessor. If we use `a` or `b[i]` variants they are stored as `Expression::Elem` and `Expression::VectorAccess` respectively, so in generated code they are not presented as random values (that's why there is no integration test so far). The idea was to change this expressions to `Expression::Rand`, but this replacement turned out to be overcomplicated due to the recursiveness of the boundary constraints. I think this problem could be easily fixed after adding boundary constraints to the constraint graph.


ir/src/symbol_table.rs
```diff
@@ -43,6 +47,10 @@ pub(super) struct SymbolTable {
     /// Vector in which index is trace segment and value is number of columns in this segment
     segment_widths: Vec<u16>,
 
+    /// Number of random values. For array initialized in `rand: [n]` form it will be `n`, and for
+    /// `rand: [a, b[n], c, ...]` it will be length of the array.
+    random_values_num: u16,
```
bobbinth, line 52: The convention we usually use is to put `num` in front. So, this would be `num_random_values`.

ir/src/symbol_table.rs
```diff
@@ -199,6 +224,10 @@ impl SymbolTable {
         &self.segment_widths
     }
 
+    pub(super) fn random_values_num(&self) -> u16 {
```
bobbinth, line 227: Similar to the other comment: should probably be `num_random_values()`

ir/src/symbol_table.rs
```diff
@@ -151,6 +159,23 @@ impl SymbolTable {
         Ok(())
     }
 
+    pub(super) fn insert_random_values(
+        &mut self,
+        values: &RandomValues,
+    ) -> Result<(), SemanticError> {
+        // TODO: add name of random values vector to the identifiers. For now it can be called only
+        // by `$rand`
+        if values.size() > 0 {
```
bobbinth, line 168: Instead of relying on `values.size()`, I'd probably add an accessor method - something like `has_bindings()`.

ir/src/symbol_table.rs
```diff
@@ -151,6 +159,23 @@ impl SymbolTable {
...
+            self.random_values_num = values.size() as u16;
+        } else {
+            self.random_values_num = values.values().len() as u16;
```
bobbinth, line 171: The logic for figuring out the number of random values should probably be encapsulated in the `RandomValues` struct.

bobbinth (commented):
Thank you! Not a full review yet - just a few assorted comments.

parser/src/ast/random_values.rs
```diff
@@ -0,0 +1,82 @@
...
+/// ```
+///
+/// created [RandomValues] instance will look like
+///
+/// `RandomValues{name: "rand", size: 0, values: [RandomValue{name: "a", size: 1}, RandomValue{name: "b", size: 12}]}`
+#[derive(Debug, Eq, PartialEq)]
+pub struct RandomValues {
+    name: Identifier,
+    size: usize,
+    values: Vec<RandomValue>,
```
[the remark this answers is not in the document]
Overcastan: Using `bindings` might be better because technically we just bind name stored here to some random value, right? 

For now `name` is not using anywhere, I added it so that in the future it would be easier to implement a random value call by this name. So we allow users to create random values arrays with custom name, but `$` access could be done only with `$rand` name (so this feature is not actually working for now). As I mentioned in PR description 
> For now it is possible to create random values array with custom name, but accessing its values could be done only with $rand call.

I thought about it, my idea was to make `size = 0` as some kind of flag, showing that we are working with named random values. But otherwise this `values` array will be empty, so, indeed, I can simplify inserting random values into the symbol table by always storing size. I'll improve this part, thank you!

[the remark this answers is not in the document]
Al-Kindi-0: I have a question related to this discussion: how would one access the random value `b[3]` if we have the following declaration `rand: [a, b[12], c]`?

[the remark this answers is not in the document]
Overcastan: Great question. I think I have confused myself and made a mistake. In this "named" style of declaration I assume that the length of random values array is equal to the number of named declarations (e.g. `[a, b[12], c]` has length 3), and then I was working with elements as integers or arrays. But maybe it's incorrect and I should work with flattened array (e.g. `[a, b[12], c]` should have length 14). It will be better in all aspects.

Answering your question: for now it's possible only with `b[3]` named access, because current implementation only supports indexes to the each named variable (so `a` is `$rand[0]`, `b` is `$rand[1]` and `c` is `$rand[2]`), and that is my fault, I didn't understand well what I had to implement. I'll fix it now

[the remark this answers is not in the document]
grjte: > Answering your question: for now it's possible only with b[3] named access, because current implementation only supports indexes to the each named variable (so a is $rand[0], b is $rand[1] and c is $rand[2]), and that is my fault, I didn't understand well what I had to implement. I'll fix it now

Yes, I think this is not the correct approach. When accessing these using `$rand` we should flatten the array as you've mentioned

air-script/tests/aux_trace/aux_trace.air
```diff
@@ -7,6 +7,9 @@ trace_columns:
 public_inputs:
     stack_inputs: [16]
 
+random_values:
+    rand: [2]
+
```
grjte, line 12: Let's remove this from here and add a specific `random_values.air` test that tests both types of random values accessors (`$rand` and named random values)

air-script/tests/variables/variables.air
```diff
@@ -10,6 +10,9 @@ public_inputs:
     stack_inputs: [16]
     stack_outputs: [16]
 
+random_values:
+    rand: [1]
+
```
grjte, line 15: Let's remove this from here and add a specific `random_values.air` test

ir/src/integrity_stmts/graph.rs
```diff
@@ -467,9 +490,10 @@ pub enum Operation {
     /// value is the length of the column's periodic cycle. The periodic value made available from
     /// the specified column is based on the current row of the trace.
     PeriodicColumn(usize, usize),
-    /// A random value provided by the verifier. The inner value is the index of this random value
-    /// in the array of all random values.
-    RandomValue(usize),
+    /// A random value provided by the verifier. The first inner value is the index of this random
+    /// value in the array of all random values and the second is value index of random value in
+    /// case it represented by array, e.g. `rand: [a, b[4]]`.
+    RandomValue(usize, usize),
```
grjte, line 496: Let's keep this the way it was. The new `RandomValue` syntax should just be semantic sugar. We should handle most of the changes during parsing, and the final part (identifying `Expression::Elem` and `Expression::VectorAccess` as random values) can be done here in the IR before getting the index in the full flattened array and saving it as an `Operation::RandomValue(usize)` type

codegen/winterfell/src/air/transition_constraints.rs
```diff
@@ -104,7 +104,7 @@ impl Codegen for Operation {
             Operation::PeriodicColumn(col_idx, _) => {
                 format!("periodic_values[{col_idx}]")
             }
-            Operation::RandomValue(idx) => {
+            Operation::RandomValue(idx, _size) => {
```
grjte, line 107: We should be able to get rid of this `_size`, as commented elsewhere

parser/src/ast/random_values.rs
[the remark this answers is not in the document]
Al-Kindi-0: Thank you guys, it makes more sense to me now.

grjte (changes requested):
Thanks @Overcastan! This is a partial review, because I think we should split this PR in half and review just the parser changes first. Some rebasing and other changes will be needed before the IR updates, since the boundary constraints PR #126  should be merged first

Please open the new PR with just the parser changes and I'll do my review of those files there

Overcastan:
> If random values are named, then we should be able to access them by name in AirScript

Yes, but in this comment I was talking about name of the hole random array, not the random values. The point is that we can declare random values array with custom name, for example `rand: [14]` or `alphas: [14]`, and I support parsing of the custom array name (so `rand` name is not mandatory). But for now we can access random values by index only with `$rand[i]`, and not with `$alphas[i]`. I left this functionality for the future PR because It requires additional changes to the hole random call, I was afraid to make this PR too big. 

> I don't think we need to name random values during code generation.

Yes, I also came to this idea after Al's question, I'll fix it

tohrnii:
Superseded by #129 and #134

PR #129, opened 2023-01-30
Parsing of Individual section of random values

Overcastan:
Partly addressing #86. 
Implementing parsing of random values, declared in individual section.

### DIFF
--- a/parser/src/ast/mod.rs
+++ b/parser/src/ast/mod.rs
@@ -16,6 +16,9 @@ pub use boundary_constraints::*;
 pub mod integrity_constraints;
 pub use integrity_constraints::*;
 
+pub mod random_values;
+pub use random_values::*;
+
 // AST
 // ================================================================================================
 
@@ -32,6 +35,7 @@ pub struct Source(pub Vec<SourceSection>);
 /// - PeriodicColumns: Periodic columns are each represented by a fixed-size array with all of its
 ///   elements specified. The array length is expected to be a power of 2, but this is not checked
 ///   during parsing.
+/// - RandomValues: Random Values represent the randomness sent by the Verifier
 /// - BoundaryConstraints: Boundary Constraints to be enforced on the boundaries of columns defined
 ///   in the TraceCols section. Currently there are two types of boundaries, First and Last
 ///   representing the first and last rows of the column.
@@ -44,6 +48,7 @@ pub enum SourceSection {
     Trace(Trace),
     PublicInputs(Vec<PublicInput>),
     PeriodicColumns(Vec<PeriodicColumn>),
+    RandomValues(RandomValues),
     BoundaryConstraints(Vec<BoundaryStmt>),
     IntegrityConstraints(Vec<IntegrityStmt>),
 }
--- a/parser/src/ast/mod.rs
+++ b/parser/src/ast/mod.rs
@@ -35,7 +35,7 @@ pub struct Source(pub Vec<SourceSection>);
 /// - PeriodicColumns: Periodic columns are each represented by a fixed-size array with all of its
 ///   elements specified. The array length is expected to be a power of 2, but this is not checked
 ///   during parsing.
-/// - RandomValues: Random Values represent the randomness sent by the Verifier
+/// - RandomValues: Random Values represent the randomness sent by the Verifier.
 /// - BoundaryConstraints: Boundary Constraints to be enforced on the boundaries of columns defined
 ///   in the TraceCols section. Currently there are two types of boundaries, First and Last
 ///   representing the first and last rows of the column.
--- a/parser/src/ast/random_values.rs
+++ b/parser/src/ast/random_values.rs
@@ -1,0 +1,85 @@
+use super::Identifier;
+
+// RANDOM VALUES
+// ================================================================================================
+
+/// Declaration of random values for an AIR. Random values could be represented by a named
+/// identifier `name` which is used to identify a fixed size array of length `size` and an empty
+/// `bindings` vector or by a named identifier `name` which is used to identify a `bindings`
+/// [RandBinding] vector and its `size` field.
+///
+/// # Examples
+///
+/// If random values are declared in form
+///
+/// ```airscript
+/// random_values:
+///     rand: [15]
+/// ```
+///
+/// created [RandomValues] instance will look like
+///
+/// `RandomValues { name: "rand", size: 15, bindings: [] }`
+///
+/// If random values are declared in form
+///
+/// ```airscript
+/// random_values:
+///     rand: [a, b[12]]
+/// ```
+///
+/// created [RandomValues] instance will look like
+///
+/// `RandomValues { name: "rand", size: 13, bindings: [ RandBinding { name: "a", size: 1 }, RandBinding { name: "b", size: 12 } ] }`
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub struct RandomValues {
+    name: Identifier,
+    size: u64,
+    bindings: Vec<RandBinding>,
+}
+
+impl RandomValues {
+    pub(crate) fn new(name: Identifier, size: u64, bindings: Vec<RandBinding>) -> Self {
+        Self {
+            name,
+            size,
+            bindings,
+        }
+    }
+
+    pub fn name(&self) -> &str {
+        let Identifier(name) = &self.name;
+        name
+    }
+
+    pub fn size(&self) -> u64 {
+        self.size
+    }
+
+    pub fn bindings(&self) -> &Vec<RandBinding> {
+        &self.bindings
+    }
+}
+
+/// Declaration of a random value binding used in [RandomValues]. It is represented by a named
+/// identifier and its size.
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub struct RandBinding {
+    name: Identifier,
+    size: u64,
+}
+
+impl RandBinding {
+    pub(crate) fn new(name: Identifier, size: u64) -> Self {
+        Self { name, size }
+    }
+
+    pub fn name(&self) -> &str {
+        let Identifier(name) = &self.name;
+        name
+    }
+
+    pub fn size(&self) -> u64 {
+        self.size
+    }
+}
--- a/parser/src/error.rs
+++ b/parser/src/error.rs
@@ -8,11 +8,12 @@ pub enum Error {
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum ParseError {
-    InvalidInt(String),
-    InvalidTraceCols(String),
-    MissingMainTraceCols(String),
     InvalidConst(String),
+    InvalidInt(String),
     InvalidListComprehension(String),
+    InvalidRandomValues(String),
+    InvalidTraceCols(String),
     MissingBoundaryConstraint(String),
     MissingIntegrityConstraint(String),
+    MissingMainTraceCols(String),
 }
--- a/parser/src/lexer/tests/random_values.rs
+++ b/parser/src/lexer/tests/random_values.rs
@@ -1,0 +1,61 @@
+use super::{expect_valid_tokenization, Token};
+
+#[test]
+fn random_values_empty_list() {
+    let source = "
+random_values:
+    rand: []";
+
+    let tokens = vec![
+        Token::RandomValues,
+        Token::Colon,
+        Token::Ident("rand".to_string()),
+        Token::Colon,
+        Token::Lsqb,
+        Token::Rsqb,
+    ];
+    expect_valid_tokenization(source, tokens);
+}
+
+#[test]
+fn random_values_fixed_list() {
+    let source = "
+random_values:
+    rand: [15]";
+
+    let tokens = vec![
+        Token::RandomValues,
+        Token::Colon,
+        Token::Ident("rand".to_string()),
+        Token::Colon,
+        Token::Lsqb,
+        Token::Num("15".to_string()),
+        Token::Rsqb,
+    ];
+    expect_valid_tokenization(source, tokens);
+}
+
+#[test]
+fn random_values_ident_vector() {
+    let source = "
+random_values:
+    rand: [a, b[12], c]";
+
+    let tokens = vec![
+        Token::RandomValues,
+        Token::Colon,
+        Token::Ident("rand".to_string()),
+        Token::Colon,
+        Token::Lsqb,
+        Token::Ident("a".to_string()),
+        Token::Comma,
+        Token::Ident("b".to_string()),
+        Token::Lsqb,
+        Token::Num("12".to_string()),
+        Token::Rsqb,
+        Token::Comma,
+        Token::Ident("c".to_string()),
+        Token::Rsqb,
+    ];
+    expect_valid_tokenization(source, tokens);
+}
--- a/parser/src/parser/grammar.lalrpop
+++ b/parser/src/parser/grammar.lalrpop
@@ -138,7 +138,7 @@ RandomValues: RandomValues = {
 
 RandValuesDecl: RandomValues = {
     <name: Identifier> ":" <size: Size> => RandomValues::new(name, size, vec![]),
-    <_name: Identifier> ":" "[" "]" =>? Err(
+    <Identifier> ":" "[" "]" =>? Err(
         ParseError::User{
             error: Error::ParseError(
                 InvalidRandomValues(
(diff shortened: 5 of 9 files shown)


air-script/tests/aux_trace/aux_trace.air
```diff
@@ -23,6 +23,6 @@ integrity_constraints:
     enf b' = c + a'
     enf c = a + b
 
-    # integrity constraints against the auxiliary trace with random values
-    enf p0' = p0 * (a + $rand[0] + b + $rand[1])
-    enf p1 = p1' * (c + $rand[0])
\ No newline at end of file
+    # integrity constraints against the auxiliary trace
+    enf p0' = p0 * (a + b)
+    enf p1 = p1' * c
```
grjte: Let's remove this from this PR

air-script/tests/aux_trace/aux_trace.rs
```diff
@@ -92,7 +92,7 @@ impl Air for AuxiliaryAir {
     {
         let current = aux_frame.current();
         let next = aux_frame.next();
-        result[0] = next[0] - ((current[0]) * (current[0] + aux_rand_elements.get_segment_elements(0)[0] + current[1] + aux_rand_elements.get_segment_elements(0)[1]));
-        result[1] = current[1] - ((next[1]) * (current[2] + aux_rand_elements.get_segment_elements(0)[0]));
+        result[0] = next[0] - ((current[0]) * (current[0] + current[1]));
+        result[1] = current[1] - ((next[1]) * (current[2]));
```
grjte, line 96: Let's remove this from this PR

air-script/tests/variables/variables.air
```diff
@@ -35,5 +35,5 @@ integrity_constraints:
     # c = a * b when s = 1.
     enf s * (c - a * b) = o[0][0] - o[0][1] - o[1][0]
 
-    # the auxiliary column contains the product of values of c offset by a random value.
-    enf p' = p * (c + $rand[0])
+    # the auxiliary column contains the product of values of c.
+    enf p' = p * c
```
grjte, line 39: Let's remove this from this PR

air-script/tests/variables/variables.rs
```diff
@@ -92,6 +92,6 @@ impl Air for VariablesAir {
     {
         let current = aux_frame.current();
         let next = aux_frame.next();
-        result[0] = next[0] - ((current[0]) * (current[3] + aux_rand_elements.get_segment_elements(0)[0]));
+        result[0] = next[0] - ((current[0]) * (current[3]));
```
grjte, line 95: Let's remove this from this PR

parser/src/lexer/tests/random_values.rs
```diff
@@ -0,0 +1,86 @@
+use super::{expect_valid_tokenization, Token};
+
+#[test]
+fn random_values_section() {
+    let source = "random_values";
+    let tokens = vec![Token::RandomValues];
+    expect_valid_tokenization(source, tokens);
+}
```
grjte, line 8: I think we can probably get rid of this test

parser/src/lexer/tests/random_values.rs
```diff
@@ -0,0 +1,86 @@
...
+        Token::Lsqb,
+        Token::Num("15".to_string()),
+        Token::Rsqb,
+    ];
+    expect_valid_tokenization(source, tokens);
+}
+
+#[test]
+fn random_values_ident_vector_default_name() {
+    let source = "
```
grjte, line 65: I would rename this test, since the relevant thing that it's testing is the group of values `b` (for the lexer `rand` vs `alphas` above is equivalent)

parser/src/parser/grammar.lalrpop
```diff
@@ -118,6 +119,24 @@ PeriodicColumn: PeriodicColumn = {
...
+
+RandomValues: RandomValues = {
+    "random_values" ":" <rand_values: RandValues> => rand_values
+}
+
+RandValues: RandomValues = {
+    <name: Identifier> ":" <index: Index> => RandomValues::new(name, index, vec![]),
+    <name: Identifier> ":" "[" "]" => RandomValues::new(name, 0, vec![]),
+    <name: Identifier> ":" "[" <rand_vec: CommaElems<RandElem>> "]" => RandomValues::new(name, rand_vec.iter().map(|v| v.size()).sum::<usize>(), rand_vec)
+}
```
grjte, line 133: I think these could be named more clearly. One of them is the section, while the other one is the declaration. I know it fits with our existing convention to call the source section one `RandomValues`, but maybe the second could be `RandValuesDecl` or something like  that?

parser/src/parser/grammar.lalrpop
```diff
@@ -118,6 +119,24 @@ PeriodicColumn: PeriodicColumn = {
...
+
+RandElem: RandBinding = {
+    <name: Identifier> => RandBinding::new(name, 1),
+    <name: Identifier> <index: Index> => RandBinding::new(name, index)
```
grjte, line 137: We should probably use `Size` instead of index here, to be consistent with the `TraceCols` and the `PublicInput` declarations

parser/src/ast/random_values.rs
```diff
@@ -0,0 +1,84 @@
...
+        &self.bindings
+    }
+}
+
+/// Declaration of a random value used in [RandomValues]. It is represented by a named identifier and its size.
+#[derive(Debug, Eq, PartialEq)]
+pub struct RandBinding {
+    name: Identifier,
+    size: usize,
+}
```
grjte, line 69: Although binding is a good word choice, this feels inconsistent with the other ways we've implemented and named things. I wonder if we should either change this to `NamedRandomValues` or plan to use the word binding more consistently in other places.

Curious what others think

parser/src/parser/tests/random_values.rs
```diff
@@ -0,0 +1,47 @@
...
+        Identifier("alphas".to_string()),
+        14,
+        vec![
+            RandBinding::new(Identifier("a".to_string()), 1),
+            RandBinding::new(Identifier("b".to_string()), 12),
+            RandBinding::new(Identifier("c".to_string()), 1),
+        ],
+    ))]);
+    build_parse_test!(source).expect_ast(expected);
+}
```
grjte, line 34: Let's separate the vector test and the custom name test

parser/src/parser/tests/random_values.rs
```diff
@@ -0,0 +1,47 @@
...
+    let source = "
+    random_values:
+        rand: []";
+    let expected = Source(vec![RandomValues(RandomValues::new(
+        Identifier("rand".to_string()),
+        0,
+        vec![],
+    ))]);
+    build_parse_test!(source).expect_ast(expected);
+}
```
grjte, line 47: This test is good, but we should probably add a few error cases as well. (e.g. empty `random_values` section or more than one set of random values being declared)

grjte (changes requested):
Thanks @Overcastan, this looks good! I left a few thoughts inline

parser/src/ast/random_values.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub fn size(&self) -> u64 {
+        self.size
+    }
+
+    pub fn bindings(&self) -> &Vec<RandBinding> {
+        &self.bindings
+    }
+}
+
+/// Declaration of a random value used in [RandomValues]. It is represented by a named identifier and its size.
```
tohrnii, line 64: nit: The doc comment should be updated.

parser/src/ast/random_values.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    pub fn name(&self) -> &str {
+        let Identifier(name) = &self.name;
+        name
+    }
+
+    pub fn size(&self) -> u64 {
+        self.size
+    }
+}
```
tohrnii, line 84: I'm wondering if we can create a structure with more generic name that we can use for other similar structs (like TraceCols and PublicInput), and then define aliases for those instead. This will make things a little less redundant. cc: @grjte

parser/src/error.rs
```diff
@@ -10,8 +10,9 @@ pub enum Error {
 pub enum ParseError {
     InvalidInt(String),
     InvalidTraceCols(String),
-    MissingMainTraceCols(String),
     InvalidConst(String),
+    InvalidRandomValues(String),
+    MissingMainTraceCols(String),
     MissingBoundaryConstraint(String),
     MissingIntegrityConstraint(String),
```
tohrnii, line 17: nit: If we are keeping the names in alphabetical order, then `InvalidInt` and `InvalidTraceCols` should probably be moved as well.

parser/src/parser/grammar.lalrpop
```diff
@@ -118,6 +120,40 @@ PeriodicColumn: PeriodicColumn = {
...
+    <_name: Identifier> ":" "[" "]" =>? Err(
+        ParseError::User{
+            error: Error::ParseError(
+                InvalidRandomValues(
+                    "Random Values section cannot be empty".to_string()
+                )
+            )
+        }
+    ),
+    <name: Identifier> ":" "[" <rand_vec: CommaElems<RandElem>> "]" => RandomValues::new(name, rand_vec.iter().map(|v| v.size()).sum(), rand_vec)
```
tohrnii, line 148: Can these two be combined to return an error if `rand_vec` is empty or does that not work?

parser/src/parser/grammar.lalrpop
```diff
@@ -118,6 +120,40 @@ PeriodicColumn: PeriodicColumn = {
...
+
+RandomValues: RandomValues = {
+    "random_values" ":" <rand_values: ListElems<RandValuesDecl>> =>? match rand_values.len() {
+        1 => Ok(rand_values[0].clone()),
+        _ => Err(ParseError::User{
+            error: Error::ParseError(
+                InvalidRandomValues("No more than one set of random values can be declared".to_string())
+            )
+        })
+    }
```
tohrnii, line 134: Can this not be replaced with just
`"random_values" ":" <rand_values: RandValuesDecl> => rand_values`?
Or was your idea to do better error reporting here? If you change it, then `ListElems` definition below can also be removed.

parser/src/ast/random_values.rs
```diff
@@ -0,0 +1,84 @@
...
+/// If random values declared in form
+///
+/// ```airscript
+/// random_values:
+///     rand: [a, b[12]]
+/// ```
+///
+/// created [RandomValues] instance will look like
+///
+/// `RandomValues{name: "rand", size: 2, bindings: [RandBinding{name: "a", size: 1}, RandBinding{name: "b", size: 12}]}`
```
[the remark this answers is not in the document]
Overcastan: Yep, it should be 13, I forgot to update the docs. Thank you!

parser/src/ast/random_values.rs
Overcastan: I've thought about it too, but I haven't figured out how to do it well. It will be great if we can implement this

parser/src/parser/grammar.lalrpop
Overcastan: Yes, I was trying to implement check of the case when more than one set of random values has been declared. If I leave it as it was (and this is exactly what you suggested) when it will panic with `UnrecognizedToken` error, and I wanted to handle this error with specific message. I don't really like the implementation, but I haven't been able to come up with anything better.

tohrnii: I think we should be able to do something like:
```
"random_values" ":" <rand_values: RandValuesDecl+>
```
and then return an error if length is not one.
Alternatively, we could do:
```
"random_values" ":" <rand_values: RandValuesDecl*>
```
so that we can report better error for empty random values section as well.

tohrnii: Although we should do better error reporting for other things in the grammar as well.

parser/src/parser/grammar.lalrpop
Overcastan: Unfortunately, I couldn't get it to work like this

parser/src/parser/grammar.lalrpop
Overcastan: Indeed, that could be done much easier, by bad. I think that `+` macro should be enough, because `rand: *nothing*` is just an incorrect declaration in general and will cause parsing error, and `rand: []` is handling in `RandValuesDecl` parsing

tohrnii: Yeah, makes sense.

parser/src/ast/random_values.rs
grjte: I agree that this would be ideal. Let me take a look and see if I think of anything. Otherwise, we can make a separate issue to revisit this in a future PR. It would be great to do it now, though.

tohrnii: Can we call this just `Array` or `ArrayDecl`?

parser/src/ast/random_values.rs
```diff
@@ -0,0 +1,85 @@
...
+/// If random values are declared in form
+///
+/// ```airscript
+/// random_values:
+///     rand: [15]
+/// ```
+///
+/// created [RandomValues] instance will look like
+///
+/// `RandomValues{name: "rand", size: 15, bindings: []}`
```
tohrnii, line 22: nit: Should there be spaces between curly braces (before and after) for clarity? Not a strong opinion, wdyt?

parser/src/error.rs
```diff
@@ -8,11 +8,12 @@ pub enum Error {
...
+    InvalidConst(String),
     InvalidInt(String),
+    InvalidListComprehension(String),
+    InvalidRandomValues(String),
     InvalidTraceCols(String),
     MissingMainTraceCols(String),
-    InvalidConst(String),
-    InvalidListComprehension(String),
     MissingBoundaryConstraint(String),
     MissingIntegrityConstraint(String),
```
tohrnii, line 18: minor nit: should probably be re-arranged to keep them alphabetical.

tohrnii (approved):
Thank you @Overcastan, looks good to me. I just left a few very minor nits.

parser/src/parser/grammar.lalrpop
```diff
@@ -120,6 +122,39 @@ PeriodicColumn: PeriodicColumn = {
...
+}
+
+RandValuesDecl: RandomValues = {
+    <name: Identifier> ":" <size: Size> => RandomValues::new(name, size, vec![]),
+    <_name: Identifier> ":" "[" "]" =>? Err(
```
grjte, line 141: nit: we can replace `<_name: Identifier>` with `<Identifier>`

parser/src/parser/tests/random_values.rs
grjte: Can you separate this into 2 tests please? One that tests the vector identifier and one that tests the custom name?

parser/src/parser/tests/random_values.rs
```diff
@@ -0,0 +1,78 @@
...
+    let source = "
+    random_values:
+        rand: [12]
+        alphas: [a, b[2]]
+        betas: [c[12], d]";
+    let error = Error::ParseError(ParseError::InvalidRandomValues(
+        "No more than one set of random values can be declared".to_string(),
+    ));
+    build_parse_test!(source).expect_error(error)
+}
```
grjte, line 78: This is good, but actually I'd do just 2 lines instead of 3, that way we have the "minimal failing test" here. Otherwise, our tests suggest that 2 sets of declarations should be ok.

parser/src/ast/random_values.rs
grjte: Let's revisit this in a future PR

parser/src/ast/random_values.rs
grjte: We can revisit this in a future PR, alongside @tohrnii's comment https://github.com/0xPolygonMiden/air-script/pull/129#discussion_r1091155024

grjte (changes requested):
Thanks @Overcastan, looks good! Just a couple very small final things

parser/src/ast/mod.rs
```diff
@@ -32,6 +35,7 @@ pub struct Source(pub Vec<SourceSection>);
 /// - PeriodicColumns: Periodic columns are each represented by a fixed-size array with all of its
 ///   elements specified. The array length is expected to be a power of 2, but this is not checked
 ///   during parsing.
+/// - RandomValues: Random Values represent the randomness sent by the Verifier
```
Al-Kindi-0, line 38: nit: a dot at the end is missing

Al-Kindi-0 (approved):

Al-Kindi-0 (approved):
Looks great, thank you @Overcastan !

grjte (approved):
LGTM! Thank you!

Issue #46, opened 2022-10-28
Throw error for boundary constraints using `$rand` against main trace


grjte:
The use of `$rand` random values by boundary constraints should be limited to auxiliary constraints and therefore to boundary constraints which are against columns from the auxiliary trace.

We need to throw an error if `$rand` is used in boundary constraints defined against main trace columns. This should be handled in the IR.

For example, this should be an error:

```
...
trace_columns:
    main: [a, b]
...
boundary_constraints:
    enf b.first = a + $rand[0]
...
```

See:
https://github.com/0xPolygonMiden/air-dsl/pull/36#discussion_r1006186112
https://github.com/0xPolygonMiden/air-dsl/pull/36#discussion_r1006188516

tohrnii:
Closed by #137.

PR #134, opened 2023-02-02
Adding random values to IR

Overcastan:
Partly addressing #86.
- Added support for random values in IR.
- Added indexed access to random values by custom name.
- Added integration test for random values.

### DIFF
--- a/air-script/tests/aux_trace/aux_trace.air
+++ b/air-script/tests/aux_trace/aux_trace.air
@@ -7,9 +7,6 @@ trace_columns:
 public_inputs:
     stack_inputs: [16]
 
-random_values:
-    rand: [2]
-
 boundary_constraints:
     enf a.first = 1
     enf b.first = 1
@@ -17,8 +14,8 @@ boundary_constraints:
     enf p0.first = 1
     enf p0.last = 1
 
-    # auxiliary boundary constraint with a random value
-    enf p1.first = $rand[0]
+    # auxiliary boundary constraint
+    enf p1.first = 1
     enf p1.last = 1
   
 integrity_constraints:
@@ -26,6 +23,6 @@ integrity_constraints:
     enf b' = c + a'
     enf c = a + b
 
-    # integrity constraints against the auxiliary trace with random values
-    enf p0' = p0 * (a + $rand[0] + b + $rand[1])
-    enf p1 = p1' * (c + $rand[0])
+    # integrity constraints against the auxiliary trace
+    enf p0' = p0 * (a + b)
+    enf p1 = p1' * c
--- a/air-script/tests/aux_trace/aux_trace.air
+++ b/air-script/tests/aux_trace/aux_trace.air
@@ -7,6 +7,9 @@ trace_columns:
 public_inputs:
     stack_inputs: [16]
 
+random_values:
+    rand: [2]
+
 boundary_constraints:
     enf a.first = 1
     enf b.first = 1
--- a/air-script/tests/aux_trace/aux_trace.air
+++ b/air-script/tests/aux_trace/aux_trace.air
@@ -7,6 +7,9 @@ trace_columns:
 public_inputs:
     stack_inputs: [16]
 
+random_values:
+    rand: [2]
+
 boundary_constraints:
     enf a.first = 1
     enf b.first = 1
@@ -14,8 +17,8 @@ boundary_constraints:
     enf p0.first = 1
     enf p0.last = 1
 
-    # auxiliary boundary constraint
-    enf p1.first = 1
+    # auxiliary boundary constraint with a random value
+    enf p1.first = $rand[0]
     enf p1.last = 1
   
 integrity_constraints:
@@ -23,6 +26,6 @@ integrity_constraints:
     enf b' = c + a'
     enf c = a + b
 
-    # integrity constraints against the auxiliary trace
-    enf p0' = p0 * (a + b)
-    enf p1 = p1' * c
+    # integrity constraints against the auxiliary trace with random values
+    enf p0' = p0 * (a + $rand[0] + b + $rand[1])
+    enf p1 = p1' * (c + $rand[0])
--- a/air-script/tests/main.rs
+++ b/air-script/tests/main.rs
@@ -112,14 +112,14 @@ fn indexed_trace_access() {
 
 #[test]
 fn random_values() {
-    let generated_air = Test::new("tests/random_values/random_values.air".to_string())
+    let generated_air = Test::new("tests/random_values/random_values_simple.air".to_string())
         .transpile()
         .unwrap();
 
     let expected = expect_file!["random_values/random_values.rs"];
     expected.assert_eq(&generated_air);
 
-    let generated_air = Test::new("tests/random_values/random_values_simple.air".to_string())
+    let generated_air = Test::new("tests/random_values/random_values_bindings.air".to_string())
         .transpile()
         .unwrap();
 
--- a/air-script/tests/main.rs
+++ b/air-script/tests/main.rs
@@ -109,3 +109,20 @@ fn indexed_trace_access() {
     let expected = expect_file!["indexed_trace_access/indexed_trace_access.rs"];
     expected.assert_eq(&generated_air);
 }
+
+#[test]
+fn random_values() {
+    let generated_air = Test::new("tests/random_values/random_values.air".to_string())
+        .transpile()
+        .unwrap();
+
+    let expected = expect_file!["random_values/random_values.rs"];
+    expected.assert_eq(&generated_air);
+
+    let generated_air = Test::new("tests/random_values/random_values_simple.air".to_string())
+        .transpile()
+        .unwrap();
+
+    let expected = expect_file!["random_values/random_values.rs"];
+    expected.assert_eq(&generated_air);
+}
--- a/air-script/tests/random_values/random_values.air
+++ b/air-script/tests/random_values/random_values.air
@@ -1,0 +1,18 @@
+def RandomValuesAir
+
+trace_columns:
+    main: [a, b]
+    aux: [c, d]
+
+public_inputs:
+    stack_inputs: [16]
+
+random_values:
+    rand: [x, y[14], z]
+
+integrity_constraints:
+    enf c' = z - $rand[0] + y[2]
+
+boundary_constraints:
+    enf c.first = $rand[5] + y[2] + z
+    enf c.last = x + $rand[15] + y[10]
--- a/air-script/tests/random_values/random_values_simple.air
+++ b/air-script/tests/random_values/random_values_simple.air
@@ -1,0 +1,18 @@
+def RandomValuesAir
+
+trace_columns:
+    main: [a, b]
+    aux: [c, d]
+
+public_inputs:
+    stack_inputs: [16]
+
+random_values:
+    rand: [16]
+
+integrity_constraints:
+    enf c' = $rand[15] - $rand[0] + $rand[3]
+
+boundary_constraints:
+    enf c.first = $rand[5] + $rand[3] + $rand[15]
+    enf c.last = $rand[0] + $rand[15] + $rand[11]
--- a/air-script/tests/random_values/random_values_simple.air
+++ b/air-script/tests/random_values/random_values_simple.air
@@ -10,9 +10,9 @@ public_inputs:
 random_values:
     rand: [16]
 
-integrity_constraints:
-    enf c' = $rand[15] - $rand[0] + $rand[3]
-
 boundary_constraints:
     enf c.first = $rand[5] + $rand[3] + $rand[15]
     enf c.last = $rand[0] + $rand[15] + $rand[11]
+
+integrity_constraints:
+    enf c' = $rand[15] - $rand[0] + $rand[3]
--- a/ir/src/constraints/graph.rs
+++ b/ir/src/constraints/graph.rs
@@ -108,13 +108,18 @@ impl AlgebraicGraph {
                 default_domain,
             ),
             Expression::Rand(index) => {
-                // The constraint target for random values defaults to the second (auxiliary) trace
-                // segment.
+                // constraint target for random values defaults to the second trace segment.
                 // TODO: make this more general, so random values from further trace segments can be
                 // used. This requires having a way to describe different sets of randomness in
                 // the AirScript syntax.
-                let node_index = self.insert_op(Operation::RandomValue(*index));
-                Ok(ExprDetails::new(node_index, AUX_SEGMENT, default_domain))
+                if *index >= symbol_table.num_random_values() as usize {
+                    return Err(SemanticError::IndexOutOfRange(format!(
+                        "Random value index {} is greater than or equal to the total number of random values ({}).", 
+                        index,
+                        symbol_table.num_random_values()
+                    )));
+                }
+                self.insert_random_value(*index, default_domain, AUX_SEGMENT)
             }
             Expression::IndexedTraceAccess(column_access) => {
                 self.insert_indexed_trace_access(symbol_table, column_access)
@@ -348,6 +353,9 @@ impl AlgebraicGraph {
             IdentifierType::Constant(ConstantType::Scalar(_)) => {
                 self.insert_constant(ConstantValue::Scalar(ident.to_string()), domain)
             }
+            IdentifierType::RandomValue(offset, _size) => {
+                self.insert_random_value(*offset, domain, AUX_SEGMENT)
+            }
             IdentifierType::IntegrityVariable(integrity_variable) => {
                 if let VariableType::Scalar(variable_expr) = integrity_variable.value() {
                     self.insert_variable(
@@ -426,6 +434,9 @@ impl AlgebraicGraph {
             IdentifierType::PublicInput(_) => {
                 unimplemented!("TODO: add support for public inputs.")
             }
+            IdentifierType::RandomValue(offset, _size) => {
+                self.insert_random_value(*offset + vector_access.idx(), domain, AUX_SEGMENT)
+            }
             _ => Err(SemanticError::invalid_vector_access(
                 vector_access,
                 symbol_type,
@@ -471,6 +482,16 @@ impl AlgebraicGraph {
                 symbol_type,
             )),
         }
+    }
+
+    fn insert_random_value(
+        &mut self,
+        index: usize,
+        domain: ConstraintDomain,
+        trace_segment: u8,
+    ) -> Result<ExprDetails, SemanticError> {
+        let node_index = self.insert_op(Operation::RandomValue(index));
+        Ok(ExprDetails::new(node_index, trace_segment, domain))
     }
 }
 
--- a/ir/src/constraints/graph.rs
+++ b/ir/src/constraints/graph.rs
@@ -108,18 +108,12 @@ impl AlgebraicGraph {
                 default_domain,
             ),
             Expression::Rand(index) => {
-                // constraint target for random values defaults to the second trace segment.
+                // The constraint target for random values defaults to the second (auxiliary) trace
+                // segment.
                 // TODO: make this more general, so random values from further trace segments can be
                 // used. This requires having a way to describe different sets of randomness in
                 // the AirScript syntax.
-                if *index >= symbol_table.num_random_values() as usize {
-                    return Err(SemanticError::IndexOutOfRange(format!(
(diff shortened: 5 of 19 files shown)


air-script/tests/aux_trace/aux_trace.air
```diff
@@ -14,15 +14,15 @@ boundary_constraints:
     enf p0.first = 1
     enf p0.last = 1
 
-    # auxiliary boundary constraint with a random value
-    enf p1.first = $rand[0]
+    # auxiliary boundary constraint
+    enf p1.first = 1
```
grjte, line 18: I said the wrong thing in my previous review, sorry. I had misunderstood that we're removing the default `$rand` option and making the declaration required.

Given that, I'd like to keep the random values in this test, so instead of removing these, we should add the `random_values` declaration (as you did the first time). Sorry for the mistake!

air-script/tests/random_values/random_values.air
```diff
@@ -0,0 +1,18 @@
...
+
+random_values:
+    rand: [x, y[14], z]
+
+integrity_constraints:
+    enf c' = z - $rand[0] + y[2]
+
+boundary_constraints:
+    enf c.first = $rand[5] + y[2] + z
+    enf c.last = x + $rand[15] + y[10]
```
grjte, line 18: This test looks great.

nit: let's put the `boundary_constraints` before the `integrity_constraints`, since we usually use that order
nit 2: maybe we should use a different name instead of `rand`, like `alphas`, since these tests also help serve as examples/documentation of functionality.

air-script/tests/main.rs
```diff
@@ -109,3 +109,20 @@ fn indexed_trace_access() {
...
+    let expected = expect_file!["random_values/random_values.rs"];
+    expected.assert_eq(&generated_air);
+
+    let generated_air = Test::new("tests/random_values/random_values_simple.air".to_string())
+        .transpile()
+        .unwrap();
+
+    let expected = expect_file!["random_values/random_values.rs"];
+    expected.assert_eq(&generated_air);
+}
```
grjte, line 128: small nit: I would probably prefer to test the simpler case first (`random_values_simple.air`), and then to give the more complex test a more specific name, like `random_values_bindings.air` or `random_values_named.air` and put it second.

air-script/tests/random_values/random_values_simple.air
```diff
@@ -0,0 +1,18 @@
...
+
+random_values:
+    rand: [16]
+
+integrity_constraints:
+    enf c' = $rand[15] - $rand[0] + $rand[3]
+
+boundary_constraints:
+    enf c.first = $rand[5] + $rand[3] + $rand[15]
+    enf c.last = $rand[0] + $rand[15] + $rand[11]
```
grjte, line 18: nit: let's swap the order of the `boundary_constraints` and `integrity_constraints`

ir/src/symbol_table.rs
```diff
@@ -151,6 +159,25 @@ impl SymbolTable {
         Ok(())
     }
 
+    /// Adds all random values by their identifier names and array length.
+    pub(super) fn insert_random_values(
+        &mut self,
+        values: &RandomValues,
+    ) -> Result<(), SemanticError> {
+        // TODO: add name of random values vector to the identifiers. For now it can be called only
+        // by `$rand`
```
grjte, line 168: Is this left out of this PR because it's quite complicated to do? If we're not including it here, can you create a sub-issue for it so we don't lose track of it?

ir/src/symbol_table.rs
```diff
@@ -249,6 +280,16 @@ impl SymbolTable {
                     ))
                 }
             }
+            IdentifierType::RandomValue(_offset, size) => {
```
grjte, line 283: nit: it's more idiomatic to use `_` instead of `_offset`

ir/src/constraints/graph.rs
```diff
@@ -108,13 +108,18 @@ impl AlgebraicGraph {
                 default_domain,
             ),
             Expression::Rand(index) => {
-                // The constraint target for random values defaults to the second (auxiliary) trace
-                // segment.
+                // constraint target for random values defaults to the second trace segment.
```
grjte, line 111: nit: I would probably leave this comment as it was, since we still have lots of references to "auxiliary" throughout the code

ir/src/constraints/graph.rs
```diff
@@ -108,13 +108,18 @@ impl AlgebraicGraph {
...
-                let node_index = self.insert_op(Operation::RandomValue(*index));
-                Ok(ExprDetails::new(node_index, AUX_SEGMENT, default_domain))
+                if *index >= symbol_table.num_random_values() as usize {
+                    return Err(SemanticError::IndexOutOfRange(format!(
+                        "Random value index {} is greater than or equal to the total number of random values ({}).", 
+                        index,
+                        symbol_table.num_random_values()
+                    )));
+                }
+                self.insert_random_value(*index, default_domain, AUX_SEGMENT)
```
grjte, line 122: nit: I'd prefer to pass the symbol table to the `insert_random_values` method as well so that the validity check and error can be moved inside that method  (`if *index >= symbol_table.num_random_values()`). That will keep this `insert_expr` method cleaner & match the pattern of the other branches

ir/src/constraints/graph.rs
```diff
@@ -472,6 +483,16 @@ impl AlgebraicGraph {
...
+
+    fn insert_random_value(
+        &mut self,
+        index: usize,
+        domain: ConstraintDomain,
+        trace_segment: u8,
+    ) -> Result<ExprDetails, SemanticError> {
+        let node_index = self.insert_op(Operation::RandomValue(index));
+        Ok(ExprDetails::new(node_index, trace_segment, domain))
+    }
```
grjte, line 495: nit: I'd move this method to just below `insert_constant`. The ordering of the methods has "access" methods after the others.

grjte (changes requested):
Looks great, thank you @Overcastan! I left several comments inline, but it's all small stuff

air-script/tests/random_values/random_values.air
grjte: I noticed in the inline `TODO` below that `alphas` won't be supported yet in the IR, so feel free to ignore "nit 2" for now. Maybe we can update this when the other change is made

ir/src/constraints/graph.rs
```diff
@@ -348,6 +353,9 @@ impl AlgebraicGraph {
             IdentifierType::Constant(ConstantType::Scalar(_)) => {
```
tohrnii, line 353: nit: The doc comment of this method should be updated to include random values.

ir/src/symbol_table.rs
```diff
@@ -249,6 +280,16 @@ impl SymbolTable {
...
+                if vector_access.idx() >= *size {
+                    Err(SemanticError::vector_access_out_of_bounds(
+                        vector_access,
+                        *size,
+                    ))
+                } else {
+                    Ok(symbol_type)
+                }
```
tohrnii, line 291: nit: We should probably keep the order of comparison similar to other match arms just to be consistent, ie,
```
if vector_access.idx() < *size {
    Ok(symbol_type)
} else {
    Err(SemanticError::vector_access_out_of_bounds(
        vector_access,
        *size,
    ))
}
```

or change the order for the others if you have preference.

ir/src/symbol_table.rs
```diff
@@ -20,6 +20,9 @@ pub(super) enum IdentifierType {
     PeriodicColumn(usize, usize),
     /// an identifier for an integrity variable, containing its name and value
     IntegrityVariable(Variable),
+    /// an identifier for random value, containing its index in the random values array and its
+    /// length if this value is an array. For non-array random values second parameter is always 1.
+    RandomValue(usize, usize),
```
tohrnii, line 25: nit: If this can represent an array, should this be called RandomValues instead?

ir/src/constraints/graph.rs
```diff
@@ -348,6 +353,9 @@ impl AlgebraicGraph {
...
                 self.insert_constant(ConstantValue::Scalar(ident.to_string()), domain)
             }
+            IdentifierType::RandomValue(offset, _size) => {
```
tohrnii, line 356: minor nit: @grjte pointed this out elsewhere that it would be better to do RandomValue(offset, _)

ir/src/constraints/graph.rs
```diff
@@ -426,6 +434,9 @@ impl AlgebraicGraph {
             IdentifierType::PublicInput(_) => {
                 unimplemented!("TODO: add support for public inputs.")
             }
+            IdentifierType::RandomValue(offset, _size) => {
```
tohrnii, line 437: minor nit: @grjte pointed this out elsewhere that it would be better to do RandomValue(offset, _)

tohrnii (changes requested):
Thank you @Overcastan, looks good to me. I just left a few small nits inline.

ir/src/symbol_table.rs
Overcastan: The problem is that we have a main array of random values (`rand: [a, b, c, ... ]`), so calling it `RandomValues` can be confusing, it will be unclear whether we mean the main array or its element, which is represented by an array. I agree with you that it would be better to rename it somehow, but I don't know yet which name would be better. Perhaps `RandomValuesElement', but this name also does not reflect that this element can be an array.

ir/src/symbol_table.rs
Overcastan: Initially, I decided to leave the implementation of this for later, because the PR turned out to be quite large, and I did not want to complicate it with additional optional functionality. But we split it into two, so I don't see any more problems not including it in the current PR. So I'll do it now!

ir/src/symbol_table.rs
grjte: It could be `RandomValuesBinding` again, to be somewhat consistent with the AST?

tohrnii: What about `RandomValuesGroup`?

ir/src/tests/mod.rs
```diff
@@ -379,6 +379,120 @@ fn transition_constraints_using_parens() {
...
+        enf a' = c + 1
+    boundary_constraints:
+        enf a.first = $rand[10] * 2
+        enf a.last = 1";
+
+    let parsed = parse(source_ident_vector_index).expect("Parsing failed");
+    let result = AirIR::from_source(&parsed);
+    assert!(result.is_err());
+}
+
```
tohrnii, line 495: I think we should change these so that random values are only used against auxiliary columns. We don't have a check for it yet but we should still have valid tests here for now. Also, do you think we should address #46 here as well or should we do that in another PR?

ir/src/symbol_table.rs
```diff
@@ -151,6 +159,31 @@ impl SymbolTable {
         Ok(())
     }
 
+    /// Adds all random values by their identifier names and array length.
+    pub(super) fn insert_random_values(
+        &mut self,
+        values: &RandomValues,
+    ) -> Result<(), SemanticError> {
+        // TODO: add name of random values vector to the identifiers. For now it can be called only
+        // by `$rand`
```
tohrnii, line 168: nit: I think this comment can be removed now?

tohrnii (changes requested):
Thank you @Overcastan, looks good to me. I just left very minor things inline. One question is if we should handle #46 in this PR and the check that you suggested regarding returning an error if there are no aux trace columns but random values section is declared or if we should leave it for the next PR.

ir/src/tests/mod.rs
Overcastan: I think I can do it in this PR, looks like it wouldn't involve a lot of changes. Just one question: should we also throw an error if random values are used against the main trace in integrity constraints?

Overcastan:
I really don't like how I implemented additional checks, but my other ideas didn't work (it's too hard to implement them). Does anyone have suggestions how it can be done better?

ir/src/tests/mod.rs
tohrnii: I don't think so. I think we can have integrity constraints with random values and main trace column access but maybe it should be an error if there are random values but no auxiliary column access in a constraint? I'm not sure, maybe @grjte or @bobbinth could confirm?

bobbinth: Ideally, something like this should be a warning - but since we don't have warnings, I'm not quite sure what's the best approach. Probably having it be an error is better - but if it creates a lot of extra work now, we can do it later too.

ir/src/constraints/graph.rs
```diff
@@ -107,14 +107,19 @@ impl AlgebraicGraph {
...
                 // used. This requires having a way to describe different sets of randomness in
                 // the AirScript syntax.
-                let node_index = self.insert_op(Operation::RandomValue(*index));
-                Ok(ExprDetails::new(node_index, AUX_SEGMENT, default_domain))
+                self.insert_random_access(
+                    symbol_table,
+                    name.name(),
+                    *index,
+                    default_domain,
+                    AUX_SEGMENT,
```
grjte, line 121: nit: Let's keep the order of the domain and segment consistent everywhere. Since we do the segment first when initializing `ExprDetails`, we should do it first here as well

ir/src/constraints/graph.rs
```diff
@@ -273,6 +278,27 @@ impl AlgebraicGraph {
...
     }
 
+    /// Inserts random value with specified index into the graph and returns the resulting
+    /// expression details.
+    fn insert_random_value(
+        &mut self,
+        symbol_table: &SymbolTable,
+        index: usize,
+        domain: ConstraintDomain,
+        trace_segment: u8,
```
grjte, line 288: nit: as noted, let's reverse the order of these

ir/src/constraints/graph.rs
```diff
@@ -472,6 +507,35 @@ impl AlgebraicGraph {
...
+    /// # Example
+    /// This function inserts values like `$alphas[3]`, having `name` as `alphas` and `index` as
+    /// `3`
+    fn insert_random_access(
+        &mut self,
+        symbol_table: &SymbolTable,
+        name: &str,
+        index: usize,
+        domain: ConstraintDomain,
+        trace_segment: u8,
```
grjte, line 527: nit: let's reverse the order of these

ir/src/lib.rs
```diff
@@ -191,4 +196,21 @@ impl AirIR {
...
+                if trace.aux_cols.is_empty() {
+                    return Err(SemanticError::InvalidUsage(
+                        "Random Values section cannot be declared without aux trace columns"
+                            .to_string(),
+                    ));
+                }
+            }
+        }
+        Ok(())
+    }
```
grjte, line 215: We probably want to use the source `validator` for this, so lets leave it for a future PR.

ir/src/lib.rs
```diff
@@ -2,6 +2,7 @@ pub use air_script_core::{
     Constant, ConstantType, Expression, Identifier, IndexedTraceAccess, MatrixAccess,
     NamedTraceAccess, TraceSegment, Variable, VariableType, VectorAccess,
 };
+use parser::ast::SourceSection;
```
grjte, line 5: It would be better to access `SourceSection` consistently everywhere, so it should either always be `ast::SourceSection` or `SourceSection`. Since we've been using the former one so far, I would stick with that, and then we don't need this import.

ir/src/constraints/mod.rs
```diff
@@ -264,6 +266,11 @@ impl Constraints {
                     domain,
                 )?;
 
+                let contains_ranom_value = self.contains_rand_value(&rhs.root_idx());
+                if contains_ranom_value && lhs.trace_segment() == DEFAULT_SEGMENT {
+                    return Err(SemanticError::InvalidUsage("Random Values cannot be used in boundary constraints defined against main trace columns".to_string()));
+                }
+
```
grjte, line 273: Let's remove this for now and deal with it in a future PR. Then the `DEFAULT_SEGMENT` doesn't need to be `pub` anymore

ir/src/symbol_table.rs
```diff
@@ -151,6 +159,33 @@ impl SymbolTable {
...
+    /// Adds all random values by their identifier names and array length.
+    pub(super) fn insert_random_values(
+        &mut self,
+        values: &RandomValues,
+    ) -> Result<(), SemanticError> {
+        self.num_random_values = values.size() as u16;
+        let mut offset = 0;
+
+        let a = Some(10);
+        if a.is_some() {}
```
grjte, line 171: It looks like we can get rid of this

ir/src/symbol_table.rs
```diff
@@ -151,6 +159,33 @@ impl SymbolTable {
...
+
+        // add array name to the symbol table
+        self.insert_symbol(
+            values.name(),
+            IdentifierType::RandomValuesBinding(offset, values.size() as usize),
+        )?;
+        // add array values to the symbol table
```
grjte, line 178: nit: I think these comments could be slightly clearer. I'd probably do the following:

```
...
        // add the name of the random values array to the symbol table
        self.insert_symbol(
            values.name(),
            IdentifierType::RandomValuesBinding(offset, values.size() as usize),
        )?;
        // add the named random value bindings to the symbol table
...
```

ir/src/tests/mod.rs
```diff
@@ -379,6 +379,167 @@ fn transition_constraints_using_parens() {
...
+    integrity_constraints:
+        enf c' = $rand[3] + 1
+    boundary_constraints:
+        enf c.first = $rand[10] * 2
+        enf c.last = 1";
+
+    let parsed = parse(source).expect("Parsing failed");
+    let result = AirIR::from_source(&parsed);
+    assert!(result.is_ok());
+}
```
grjte, line 401: nit: here and on all tests, I'd prefer to put the `boundary_constraints` before the `integrity_constraints`, since that's generally what we do, and consistency tends to be helpful

ir/src/tests/mod.rs
```diff
@@ -379,6 +379,167 @@ fn transition_constraints_using_parens() {
...
+    integrity_constraints:
+        enf c' = $rand[3] + 1
+    boundary_constraints:
+        enf a.first = $rand[10] * 2
+        enf b[2].last = 1";
+
+    let parsed = parse(source).expect("Parsing failed");
+    let result = AirIR::from_source(&parsed);
+    assert!(result.is_err());
+}
```
grjte, line 541: Since we're not going to add this check until later, you could add the `ignore` attribute to this test for now and a `TODO` that we'll turn it on later (or it's fine to just remove the test and we can write it again later)

grjte (changes requested):
Thanks @Overcastan! I left a few more comments inline. I think we can simplify and leave the validity checks for a future PR, since this one is good overall and I'd rather merge it sooner

ir/src/symbol_table.rs
```diff
@@ -32,6 +35,7 @@ impl Display for IdentifierType {
                 write!(f, "TraceColumns in segment {}", columns.trace_segment())
             }
             Self::IntegrityVariable(_) => write!(f, "IntegrityVariable"),
+            Self::RandomValuesBinding(_, _) => write!(f, "RandomValue"),
```
tohrnii, line 38: minor nit: Should this be RandomValuesBinding?

ir/src/tests/mod.rs
```diff
@@ -379,6 +379,126 @@ fn transition_constraints_using_parens() {
     assert!(result.is_ok());
 }
 
+#[test]
+fn random_values_fixed_list() {
```
tohrnii, line 383: nit: Should this be renamed to `random_values_indexed_access` or something like that?

ir/src/tests/mod.rs
```diff
@@ -379,6 +379,126 @@ fn transition_constraints_using_parens() {
...
+    integrity_constraints:
+        enf c' = $alphas[3] + 1";
+
+    let parsed = parse(source).expect("Parsing failed");
+    let result = AirIR::from_source(&parsed);
+    assert!(result.is_ok());
+}
+
+#[test]
+fn random_values_ident_vector() {
```
tohrnii, line 425: nit: Should this be renamed to `random_values_named_access` or something similar?

tohrnii (approved):
Thank you @Overcastan, looks good to me. I just left a few very minor naming related nits but don't have strong opinion on those.

ir/src/symbol_table.rs
Overcastan: My bad, I overlooked it. Thank you!

grjte (approved):
LGTM! Thanks @Overcastan!

PR #136, opened 2023-02-07
feat(ir): add list comprehension to IR

tohrnii:
Partly addressing #87. This PR adds list comprehension and list folding support to the IR.

### DIFF
--- a/air-script/tests/list_comprehension/list_comprehension.air
+++ b/air-script/tests/list_comprehension/list_comprehension.air
@@ -1,0 +1,35 @@
+def ListComprehensionAir
+
+trace_columns:
+    main: [clk, fmp[2], ctx]
+    aux: [a, b[3], c[4], d[4]]
+
+public_inputs:
+    stack_inputs: [16]
+
+boundary_constraints:
+    enf c[2].first = 0
+
+integrity_constraints:
+    let x = [fmp for fmp in fmp]
+    enf clk = x[1]
+
+    let enumerate = [2^i * c for (i, c) in (0..4, c)]
+    enf a = clk * enumerate[3]
+    
+    let diff_ident_iterables = [x' - y' for (x, y) in (c, d)]
+    enf a = clk * diff_ident_iterables[0]
+    
+    let diff_slice_iterables = [x - y for (x, y) in (c[0..2], d[1..3])]
+    enf b[1] = clk * diff_slice_iterables[1]
+    
+    let x0 = [w + x - y - z for (w, x, y, z) in (0..3, b, c[0..3], d[0..3])]
+    enf a = x0[0] + x0[1] + x0[2]
+    
+    let x1 = sum([c * d for (c, d) in (c, d)])
+    let y1 = prod([c + d for (c, d) in (c, d)])    
+    enf c[3] = x1 + y1
+    
+    let x2 = sum([c * d for (c, d) in (c, d)])
+    let y2 = [m + x2 for m in fmp]
+    enf a * clk = y2[0] + x2
--- a/air-script/tests/list_comprehension/list_comprehension.rs
+++ b/air-script/tests/list_comprehension/list_comprehension.rs
@@ -86,11 +86,11 @@ impl Air for ListComprehensionAir {
         let main_next = main_frame.next();
         let aux_current = aux_frame.current();
         let aux_next = aux_frame.next();
-        result[0] = aux_current[0] - main_current[0] * E::from(2_u64).exp(E::PositiveInteger::from(3_u64)) * aux_current[7];
-        result[1] = aux_current[0] - main_current[0] * (aux_next[4] - aux_next[8]);
-        result[2] = aux_current[2] - main_current[0] * (aux_current[5] - aux_current[10]);
+        result[0] = aux_current[0] - E::from(main_current[0]) * E::from(2_u64).exp(E::PositiveInteger::from(3_u64)) * aux_current[7];
+        result[1] = aux_current[0] - E::from(main_current[0]) * (aux_next[4] - aux_next[8]);
+        result[2] = aux_current[2] - E::from(main_current[0]) * (aux_current[5] - aux_current[10]);
         result[3] = aux_current[0] - (E::from(0_u64) + aux_current[1] - aux_current[4] - aux_current[8] + E::from(1_u64) + aux_current[2] - aux_current[5] - aux_current[9] + E::from(2_u64) + aux_current[3] - aux_current[6] - aux_current[10]);
         result[4] = aux_current[7] - (aux_current[4] * aux_current[8] + aux_current[5] * aux_current[9] + aux_current[6] * aux_current[10] + aux_current[7] * aux_current[11] + (aux_current[4] + aux_current[8]) * (aux_current[5] + aux_current[9]) * (aux_current[6] + aux_current[10]) * (aux_current[7] + aux_current[11]));
-        result[5] = aux_current[0] * main_current[0] - (main_current[1] + aux_current[4] * aux_current[8] + aux_current[5] * aux_current[9] + aux_current[6] * aux_current[10] + aux_current[7] * aux_current[11] + aux_current[4] * aux_current[8] + aux_current[5] * aux_current[9] + aux_current[6] * aux_current[10] + aux_current[7] * aux_current[11]);
+        result[5] = aux_current[0] * E::from(main_current[0]) - (E::from(main_current[1]) + aux_current[4] * aux_current[8] + aux_current[5] * aux_current[9] + aux_current[6] * aux_current[10] + aux_current[7] * aux_current[11] + aux_current[4] * aux_current[8] + aux_current[5] * aux_current[9] + aux_current[6] * aux_current[10] + aux_current[7] * aux_current[11]);
     }
 }
--- a/air-script/tests/list_comprehension/list_comprehension.rs
+++ b/air-script/tests/list_comprehension/list_comprehension.rs
@@ -1,0 +1,96 @@
+use winter_air::{Air, AirContext, Assertion, AuxTraceRandElements, EvaluationFrame, ProofOptions as WinterProofOptions, TransitionConstraintDegree, TraceInfo};
+use winter_math::fields::f64::BaseElement as Felt;
+use winter_math::{ExtensionOf, FieldElement};
+use winter_utils::collections::Vec;
+use winter_utils::{ByteWriter, Serializable};
+
+pub struct PublicInputs {
+    stack_inputs: [Felt; 16],
+}
+
+impl PublicInputs {
+    pub fn new(stack_inputs: [Felt; 16]) -> Self {
+        Self { stack_inputs }
+    }
+}
+
+impl Serializable for PublicInputs {
+    fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        target.write(self.stack_inputs.as_slice());
+    }
+}
+
+pub struct ListComprehensionAir {
+    context: AirContext<Felt>,
+    stack_inputs: [Felt; 16],
+}
+
+impl ListComprehensionAir {
+    pub fn last_step(&self) -> usize {
+        self.trace_length() - self.context().num_transition_exemptions()
+    }
+}
+
+impl Air for ListComprehensionAir {
+    type BaseField = Felt;
+    type PublicInputs = PublicInputs;
+
+    fn context(&self) -> &AirContext<Felt> {
+        &self.context
+    }
+
+    fn new(trace_info: TraceInfo, public_inputs: PublicInputs, options: WinterProofOptions) -> Self {
+        let main_degrees = vec![TransitionConstraintDegree::new(1)];
+        let aux_degrees = vec![TransitionConstraintDegree::new(2), TransitionConstraintDegree::new(2), TransitionConstraintDegree::new(2), TransitionConstraintDegree::new(1), TransitionConstraintDegree::new(4), TransitionConstraintDegree::new(2)];
+        let num_main_assertions = 0;
+        let num_aux_assertions = 1;
+
+        let context = AirContext::new_multi_segment(
+            trace_info,
+            main_degrees,
+            aux_degrees,
+            num_main_assertions,
+            num_aux_assertions,
+            options,
+        )
+        .set_num_transition_exemptions(2);
+        Self { context, stack_inputs: public_inputs.stack_inputs }
+    }
+
+    fn get_periodic_column_values(&self) -> Vec<Vec<Felt>> {
+        vec![]
+    }
+
+    fn get_assertions(&self) -> Vec<Assertion<Felt>> {
+        let mut result = Vec::new();
+        result
+    }
+
+    fn get_aux_assertions<E: FieldElement<BaseField = Felt>>(&self, aux_rand_elements: &AuxTraceRandElements<E>) -> Vec<Assertion<E>> {
+        let mut result = Vec::new();
+        result.push(Assertion::single(6, 0, E::from(0_u64)));
+        result
+    }
+
+    fn evaluate_transition<E: FieldElement<BaseField = Felt>>(&self, frame: &EvaluationFrame<E>, periodic_values: &[E], result: &mut [E]) {
+        let main_current = frame.current();
+        let main_next = frame.next();
+        result[0] = main_current[0] - main_current[2];
+    }
+
+    fn evaluate_aux_transition<F, E>(&self, main_frame: &EvaluationFrame<F>, aux_frame: &EvaluationFrame<E>, _periodic_values: &[F], aux_rand_elements: &AuxTraceRandElements<E>, result: &mut [E])
+    where F: FieldElement<BaseField = Felt>,
+          E: FieldElement<BaseField = Felt> + ExtensionOf<F>,
+    {
+        let main_current = main_frame.current();
+        let main_next = main_frame.next();
+        let aux_current = aux_frame.current();
+        let aux_next = aux_frame.next();
+        result[0] = aux_current[0] - main_current[0] * E::from(2_u64).exp(E::PositiveInteger::from(3_u64)) * aux_current[7];
+        result[1] = aux_current[0] - main_current[0] * (aux_next[4] - aux_next[8]);
+        result[2] = aux_current[2] - main_current[0] * (aux_current[5] - aux_current[10]);
+        result[3] = aux_current[0] - (E::from(0_u64) + aux_current[1] - aux_current[4] - aux_current[8] + E::from(1_u64) + aux_current[2] - aux_current[5] - aux_current[9] + E::from(2_u64) + aux_current[3] - aux_current[6] - aux_current[10]);
+        result[4] = aux_current[7] - (aux_current[4] * aux_current[8] + aux_current[5] * aux_current[9] + aux_current[6] * aux_current[10] + aux_current[7] * aux_current[11] + (aux_current[4] + aux_current[8]) * (aux_current[5] + aux_current[9]) * (aux_current[6] + aux_current[10]) * (aux_current[7] + aux_current[11]));
+        result[5] = aux_current[0] * main_current[0] - (main_current[1] + aux_current[4] * aux_current[8] + aux_current[5] * aux_current[9] + aux_current[6] * aux_current[10] + aux_current[7] * aux_current[11] + aux_current[4] * aux_current[8] + aux_current[5] * aux_current[9] + aux_current[6] * aux_current[10] + aux_current[7] * aux_current[11]);
+    }
+}
--- a/ir/src/constraints/graph.rs
+++ b/ir/src/constraints/graph.rs
@@ -1,14 +1,15 @@
 use super::{
-    BTreeMap, ConstantType, ConstraintDomain, ExprDetails, Expression, Identifier, IdentifierType,
-    IndexedTraceAccess, IntegrityConstraintDegree, MatrixAccess, Scope, SemanticError, SymbolTable,
-    TraceSegment, VariableRoots, VariableType, VectorAccess,
+    list_comprehension::unfold_lc, BTreeMap, ConstantType, ConstraintDomain, ExprDetails,
+    Expression, Identifier, IdentifierType, IndexedTraceAccess, IntegrityConstraintDegree,
+    ListFoldingType, MatrixAccess, Scope, SemanticError, SymbolTable, TraceSegment, VariableRoots,
+    VariableType, VectorAccess,
 };
 
 // CONSTANTS
 // ================================================================================================
 
 /// The offset of the "current" row during constraint evaluation.
-const CURRENT_ROW: usize = 0;
+pub const CURRENT_ROW: usize = 0;
 /// The default segment against which a constraint is applied is the main trace segment.
 pub const DEFAULT_SEGMENT: TraceSegment = 0;
 /// The auxiliary trace segment.
@@ -156,7 +157,10 @@ impl AlgebraicGraph {
                 let node_index = if let Expression::Const(rhs) = **rhs {
                     self.insert_op(Operation::Exp(lhs.root_idx(), rhs as usize))
                 } else {
-                    todo!()
+                    Err(SemanticError::InvalidUsage(
+                        "Non const exponents are only allowed inside list comprehensions"
+                            .to_string(),
+                    ))?
                 };
 
                 Ok(ExprDetails::new(
@@ -165,7 +169,9 @@ impl AlgebraicGraph {
                     lhs.domain(),
                 ))
             }
-            Expression::ListFolding(_) => todo!(),
+            Expression::ListFolding(lf_type) => {
+                self.insert_list_folding(symbol_table, lf_type, variable_roots, default_domain)
+            }
         }
     }
 
@@ -427,6 +433,11 @@ impl AlgebraicGraph {
                 Ok(ExprDetails::new(node_index, trace_segment, domain))
             }
             IdentifierType::PublicInput(_) => {
+                if !domain.is_boundary() {
+                    return Err(SemanticError::InvalidUsage(
+                        "Public inputs cannot be accessed in integrity constraints.".to_string(),
+                    ));
+                }
                 let node_index = self.insert_op(Operation::PublicInput(
                     vector_access.name().to_string(),
                     vector_access.idx(),
@@ -513,6 +524,42 @@ impl AlgebraicGraph {
             _ => Err(SemanticError::InvalidUsage(format!(
                 "Identifier {name} was declared as a {elem_type} not as a random values"
             ))),
+        }
+    }
+
+    /// Inserts a list folding expression into the graph and returns the resulting expression
+    /// details.
+    ///
+    /// # Errors
+    /// - Panics if the list is empty.
+    /// - Returns an error if the list cannot be unfolded properly.
+    fn insert_list_folding(
+        &mut self,
+        symbol_table: &SymbolTable,
+        lf_type: &ListFoldingType,
+        variable_roots: &mut VariableRoots,
+        domain: ConstraintDomain,
+    ) -> Result<ExprDetails, SemanticError> {
+        match lf_type {
+            ListFoldingType::Sum(lc) | ListFoldingType::Prod(lc) => {
+                let list = unfold_lc(lc, symbol_table)?;
+                assert!(
+                    !list.is_empty(),
+                    "List on which list folding is applied is empty."
+                );
+
+                let mut acc = self.insert_expr(symbol_table, &list[0], variable_roots, domain)?;
+                for elem in list.iter().skip(1) {
+                    let expr = self.insert_expr(symbol_table, elem, variable_roots, domain)?;
+                    let op = match lf_type {
+                        ListFoldingType::Sum(_) => Operation::Add(acc.root_idx(), expr.root_idx()),
+                        ListFoldingType::Prod(_) => Operation::Mul(acc.root_idx(), expr.root_idx()),
+                    };
+                    acc = self.insert_bin_op(&acc, &expr, op)?;
+                }
+
+                Ok(acc)
+            }
         }
     }
 }
--- a/ir/src/constraints/list_comprehension.rs
+++ b/ir/src/constraints/list_comprehension.rs
@@ -145,14 +145,15 @@ fn parse_named_trace_access(
                             named_trace_access.row_offset(),
                         )))
                     }
-                    _ => Err(SemanticError::InvalidListComprehension(
-                        "Iterable should be a trace column".to_string(),
-                    ))?,
+                    _ => Err(SemanticError::InvalidListComprehension(format!(
+                        "Iterable {ident} should contain trace columns"
+                    )))?,
                 }
             }
-            Iterable::Range(_) => Err(SemanticError::InvalidListComprehension(
-                "Iterable cannot be of range type here".to_string(),
-            )),
+            Iterable::Range(_) => Err(SemanticError::InvalidListComprehension(format!(
+                "Iterable cannot be of type Range for named trace access {}",
+                named_trace_access.name()
+            ))),
             Iterable::Slice(ident, range) => {
                 let ident_type = symbol_table.get_type(ident.name())?;
                 match ident_type {
@@ -164,9 +165,9 @@ fn parse_named_trace_access(
                             named_trace_access.row_offset(),
                         )))
                     }
-                    _ => Err(SemanticError::InvalidListComprehension(
-                        "Iterable should be a trace column".to_string(),
-                    ))?,
+                    _ => Err(SemanticError::InvalidListComprehension(format!(
+                        "Iterable {ident} should contain trace columns"
+                    )))?,
                 }
             }
         },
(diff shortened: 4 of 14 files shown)


ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,328 @@
...
+    Ok(vector)
+}
+
+/// Parses a list comprehension expression and creates an expression based on the index of the
+/// expression in the list comprehension.
+///
+/// Returns an error if there is an error while parsing the sub-expression.
+fn parse_lc_expr(
+    expression: &Expression,
+    lc: &ListComprehension,
```
bobbinth, line 31: It seems to me that we are passing `ListComprehension` around to get access to its context and then use loops to figure out which member corresponds to which iterable. If so, a better solution could be to build a map from identifier to iterable upfront (maybe put this map into something a struct - something like `IterableContext`), and then pass this map around instead `ListComprehension`.

This way, instead of doing something like:
```Rust
for (member, iterable) in lc.context() {
    if ident == member {
        match iterable {
```
You'd be able to do something like:
```Rust
let iterable = context.get_iterable(ident)?;
match iterable {
```

You may be able to take it even a step further and put `IdentifierType` for each member into the context. This way, you won't need to pass `symbol_table` around either.

air-script/tests/list_comprehension/list_comprehension.air
```diff
@@ -0,0 +1,29 @@
...
+    let diff1 = [x - y for (x, y) in (c[0..2], d[1..3])]
+    enf clk = diff1[1]
+    let x0 = [w + x - y - z for (w, x, y, z) in (0..3, b, c[0..3], d[0..3])]
+    enf a = x0[0] + x0[1] + x0[2]
+    let x1 = sum([c * d for (c, d) in (c, d)])
+    let y1 = prod([c + d for (c, d) in (c, d)])    
+    enf clk = x1
+    let x2 = sum([c * d for (c, d) in (c, d)])
+    let y2 = [m + x2 for m in fmp]
+    enf clk = y2[0]
```
bobbinth, line 29: minor nit: I'd probably add blank lines after each `enf` - would make it a bit easier to read.

Also, could we add a test for using "next" value from a column?

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,328 @@
...
+                                ))?,
+                            }
+                        }
+                        IdentifierType::PublicInput(size) => {
+                            validate_access(i, *size)?;
+                            return Ok(Expression::VectorAccess(VectorAccess::new(
+                                ident.clone(),
+                                i,
+                            )));
+                        }
```
bobbinth, line 117: Public inputs should not be accessible in integrity constraints. Is this checked somewhere later?

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,328 @@
...
+            match iterable {
+                Iterable::Identifier(ident) => {
+                    let ident_type = symbol_table.get_type(ident.name())?;
+                    match ident_type {
+                        IdentifierType::TraceColumns(trace_columns) => {
+                            validate_access(i, trace_columns.size())?;
+                            return Ok(Expression::NamedTraceAccess(NamedTraceAccess::new(
+                                ident.clone(),
+                                i,
+                                CURRENT_ROW,
```
bobbinth, line 96: Why do we hard-code `CURRENT_ROW` here? Shouldn't the offset come from `trace_columns`?

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,328 @@
...
+                Iterable::Slice(ident, range) => {
+                    let ident_type = symbol_table.get_type(ident.name())?;
+                    match ident_type {
+                        IdentifierType::TraceColumns(trace_columns) => {
+                            validate_access(i, trace_columns.size())?;
+                            return Ok(Expression::NamedTraceAccess(NamedTraceAccess::new(
+                                ident.clone(),
+                                range.start() + i,
+                                CURRENT_ROW,
+                            )));
```
bobbinth, line 142: Why do we hard-code `CURRENT_ROW` here? Shouldn't the offset come from `trace_columns`?

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,328 @@
...
+                                ))?,
+                            }
+                        }
+                        IdentifierType::PublicInput(size) => {
+                            validate_access(i, *size)?;
+                            return Ok(Expression::VectorAccess(VectorAccess::new(
+                                ident.clone(),
+                                range.start() + i,
+                            )));
+                        }
```
bobbinth, line 162: Public inputs should not be accessible in integrity constraints. Is this checked somewhere later?

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,328 @@
...
+                Iterable::Identifier(ident) => {
+                    let ident_type = symbol_table.get_type(ident.name())?;
+                    match ident_type {
+                        IdentifierType::TraceColumns(size) => {
+                            validate_access(i, size.size())?;
+                            return Ok(Expression::NamedTraceAccess(NamedTraceAccess::new(
+                                ident.clone(),
+                                i,
+                                NEXT_ROW,
+                            )));
```
bobbinth, line 208: Why do we hard-code `NEXT_ROW` here? Shouldn't the offset come from `trace_columns`?

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,328 @@
...
+                Iterable::Slice(ident, range) => {
+                    let ident_type = symbol_table.get_type(ident.name())?;
+                    match ident_type {
+                        IdentifierType::TraceColumns(trace_columns) => {
+                            validate_access(i, trace_columns.size())?;
+                            return Ok(Expression::NamedTraceAccess(NamedTraceAccess::new(
+                                ident.clone(),
+                                range.start() + i,
+                                NEXT_ROW,
+                            )));
```
bobbinth, line 229: Why do we hard-code `NEXT_ROW` here? Shouldn't the offset come from `trace_columns`?

ir/src/constraints/graph.rs
```diff
@@ -166,7 +173,38 @@ impl AlgebraicGraph {
...
+                    for elem in list.iter().skip(1) {
+                        let expr =
+                            self.insert_expr(symbol_table, elem, variable_roots, default_domain)?;
+                        sum = self.insert_bin_op(
+                            &sum,
+                            &expr,
+                            Operation::Add(sum.root_idx(), expr.root_idx()),
+                        )?;
+                    }
+                    Ok(sum)
```
bobbinth, line 190: It may be possible to create a function which does this for both `Sum` and `Prod` - the only thing which is different between the two is which binary operation we use in the end.

bobbinth (changes requested):
Thank you! Looks good! This is not a full review yet - but I left some comments inline.

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,328 @@
...
+        Iterable::Identifier(ident) => {
+            let ident_type = symbol_table.get_type(ident.name())?;
+            match ident_type {
+                IdentifierType::IntegrityVariable(var_type) => match var_type.value() {
+                    VariableType::Vector(vector) => Ok(vector.len()),
+                    _ => Err(SemanticError::InvalidListComprehension(
+                        "List comprehensions only allowed for vector variables.".to_string(),
+                    )),
+                },
+                IdentifierType::PublicInput(size) => Ok(*size),
```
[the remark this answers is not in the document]
tohrnii: Thanks for catching this. It's not checked currently. I've added a check later on to throw an error if public inputs are accessed in integrity constraints.

ir/src/constraints/list_comprehension.rs
tohrnii: Thanks for catching this. It's not checked currently. I've added a check later on to throw an error if public inputs are accessed in integrity constraints.

ir/src/constraints/list_comprehension.rs
tohrnii: The AST has current row trace columns stored as general element since there's no way to disambiguate between whether it's a trace column or any other symbol table like an intermediate variable. So, if it's stored as a trace column in symbol table, we know that it should have 0 row offset here.

ir/src/constraints/list_comprehension.rs
tohrnii: The AST has current row trace columns stored as general element since there's no way to disambiguate between whether it's a trace column or any other symbol table like an intermediate variable. So, if it's stored as a trace column in symbol table, we know that it should have 0 row offset here.

ir/src/constraints/list_comprehension.rs
tohrnii: You are right, it should come from `named_trace_access` here. Thanks.

ir/src/constraints/list_comprehension.rs
tohrnii: You are right, it should come from `named_trace_access` here. Thanks.

ir/src/constraints/list_comprehension.rs
tohrnii: Thanks, done. I moved it to just a `BTreeMap<Identifier, Iterable>`. I didn't remove symbol table as it is required to build context for list folding inside list comprehension.

ir/src/constraints/graph.rs
tohrnii: Done

ir/src/constraints/list_comprehension.rs
bobbinth: I might be missing something here, but how is something like this stored in the symbol table:
```
trace_columns:
    main: [a, b]

integrity_constraints:
    let x = a'
    enf x = b + 2
```

bobbinth (changes requested):
Thank you! Still not a full review - but I left some comments inline.

I think the biggest issue is that the constraint test file (`list_comprehension.air`) seems to be compiled incorrectly into Rust code. I'm not sure if this is due to this PR or is a prior issue - but I think it would be difficult to check the behavior until this is corrected.

air-script/tests/list_comprehension/list_comprehension.rs
```diff
@@ -0,0 +1,94 @@
...
+        result[0] = current[0] - (current[2]);
+    }
+
+    fn evaluate_aux_transition<F, E>(&self, main_frame: &EvaluationFrame<F>, aux_frame: &EvaluationFrame<E>, _periodic_values: &[F], aux_rand_elements: &AuxTraceRandElements<E>, result: &mut [E])
+    where F: FieldElement<BaseField = Felt>,
+          E: FieldElement<BaseField = Felt> + ExtensionOf<F>,
+    {
+        let current = aux_frame.current();
+        let next = aux_frame.next();
+        result[0] = current[0] - (((E::from(2_u64)).exp(E::PositiveInteger::from(3_u64))) * (current[7]));
```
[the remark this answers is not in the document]
bobbinth: Actually, I would even say that the expression should have been just `E::from(8_u64)` but maybe this is an optimization step.

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,350 @@
...
+fn parse_elem(
+    ident: &Identifier,
+    expression: &Expression,
+    iterable_context: &IterableContext,
+    symbol_table: &SymbolTable,
+    i: usize,
+) -> Result<Expression, SemanticError> {
+    let iterable = iterable_context.get(ident);
+    if let Some(iterable_type) = iterable {
+        match iterable_type {
```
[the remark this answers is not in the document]
tohrnii: @bobbinth I think we can have symbols in a list comprehension expression that are not in the iterables of the list comprehension. For example, the following should be valid:
```
trace_columns:
    main: [c[3]]
let x = 1
let a = [b + x for b in c]
```
Here, when we are unrolling the list comprehension, we just leave x alone and return it if we don't find a corresponding iterable. So the above list comprehension unrolls into:
```
[c[0] + x, c[1] + x, c[2] + x]
```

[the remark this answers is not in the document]
bobbinth: Ah - makes sense! I'd probably add some comments and maybe restructure the function to make it more clear.

For example, for `parse_elem()` do we really need to pass in `expression`? If I understood it correctly, passing both `ident` and `expression` is redundant. And the only reason we pass it in is to clone it at the end. But we could have just as easily done something like `Expression::Elem(ident)` right?

And the logic of the whole function could be:
1. If the `ident` is not in the `iterable_context`, then just return `Expression::Elem(ident)` (and I guess any invalid references will be caught later).
2. Otherwise, process the `ident` as we do now.

ir/src/constraints/mod.rs
```diff
@@ -232,7 +236,15 @@ impl Constraints {
                 self.insert_constraint(symbol_table, lhs, rhs)?
             }
             IntegrityStmt::Variable(variable) => {
-                symbol_table.insert_integrity_variable(variable)?
+                if let VariableType::ListComprehension(list_comprehension) = variable.value() {
+                    let vector = unfold_lc(list_comprehension, symbol_table)?;
```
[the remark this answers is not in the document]
tohrnii: `expressions` seems a little vague to me since we are returning a vector of expressions to be used for a vector type variable in this context. Do you think `variable_value` is more accurate? Not a strong opinion though, I'm happy to go with `expressions` if that seems better to you.

air-script/tests/list_comprehension/list_comprehension.rs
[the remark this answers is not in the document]
tohrnii: Thanks. Yeah, I think the second one should probably be an optimization step. I'll add an issue to track these.

air-script/tests/list_comprehension/list_comprehension.air
```diff
@@ -0,0 +1,35 @@
...
+    let x0 = [w + x - y - z for (w, x, y, z) in (0..3, b, c[0..3], d[0..3])]
+    enf a = x0[0] + x0[1] + x0[2]
+    
+    let x1 = sum([c * d for (c, d) in (c, d)])
+    let y1 = prod([c + d for (c, d) in (c, d)])    
+    enf clk = x1
+    
+    let x2 = sum([c * d for (c, d) in (c, d)])
+    let y2 = [m + x2 for m in fmp]
+    enf clk = y2[0]
```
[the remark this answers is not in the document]
tohrnii: Makes sense. I was a little sloppy in writing these examples (just wanted to write down different kinds of expressions possible. I'll improve on them to only include valid expressions and will add some useful ones.

ir/src/constraints/list_comprehension.rs
tohrnii: The trace columns are stored as:
```
TraceColumns(TraceColumns)
```
where
```
pub struct TraceColumns {
    trace_segment: TraceSegment,
    offset: usize,
    size: usize,
}
```
where offset is the column offset of the trace column group instead of row offset.

When we insert the intermediate variable `let x = a'`, `a'` would be stored in the symbol table as:
```
TraceElement(IndexedTraceAccess { trace_segment: 0, col_idx: 0, row_offset: 1 })
```

I think the confusion might be that `a` and `a'` are stored differently in the AST. When we encounter `a` at the parser level we don't know if it's a trace column or any other identifier so we store it as `Expression::Elem`, however when we encounter `a'` at the parser level, we know that it should be referring it to a trace column so we store it as `Expression::NamedTraceAccess(NamedTraceAccess)` where
```
pub struct NamedTraceAccess {
    name: Identifier,
    idx: usize,
    row_offset: usize,
}
```

Maybe there's a better way to do this since this is certainly a little confusing. Please let me know if you have other suggestion.

So, when we parse the AST expression while unfolding list comprehension, if the expression is of type `Elem`, we know that it should be referring to current row.

air-script/tests/list_comprehension/list_comprehension.rs
```diff
@@ -0,0 +1,94 @@
...
+        result[1] = current[0] - (next[4] - (next[8]));
+        result[2] = current[0] - (current[5] - (current[10]));
+        result[3] = current[0] - (E::from(0_u64) + current[1] - (current[4]) - (current[8]) + E::from(1_u64) + current[2] - (current[5]) - (current[9]) + E::from(2_u64) + current[3] - (current[6]) - (current[10]));
+        result[4] = current[0] - ((current[4]) * (current[8]) + (current[5]) * (current[9]) + (current[6]) * (current[10]) + (current[7]) * (current[11]));
+        result[5] = current[0] - (current[1] + (current[4]) * (current[8]) + (current[5]) * (current[9]) + (current[6]) * (current[10]) + (current[7]) * (current[11]));
+    }
```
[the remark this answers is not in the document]
tohrnii: Thanks for catching this. This seems like it was an existing issue. I'll open a PR today with a fix.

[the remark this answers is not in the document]
tohrnii: I've opened #148.

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,342 @@
+use std::collections::BTreeMap;
+
+use super::{graph::CURRENT_ROW, IdentifierType, SemanticError, SymbolTable};
```
bobbinth, line 3: minor nit: I'd remove the blank line.

ir/src/constraints/graph.rs
```diff
@@ -398,6 +423,14 @@ impl AlgebraicGraph {
                 Ok(ExprDetails::new(node_index, trace_segment, domain))
             }
             IdentifierType::PublicInput(_) => {
+                if !matches!(
+                    domain,
+                    ConstraintDomain::FirstRow | ConstraintDomain::LastRow
```
bobbinth, line 428: Should we add `is_boundary()` to `ConstraintDomain`?

ir/src/constraints/graph.rs
```diff
@@ -165,7 +170,27 @@ impl AlgebraicGraph {
...
+                                Operation::Add(acc.root_idx(), expr.root_idx())
+                            }
+                            ListFoldingType::Prod(_) => {
+                                Operation::Mul(acc.root_idx(), expr.root_idx())
+                            }
+                        };
+                        acc = self.insert_bin_op(&acc, &expr, op)?;
+                    }
+                    Ok(acc)
+                }
```
bobbinth, line 192: Let's move this into a separate function - something like `insert_list_folding()` (similar to how we have separate functions `insert_vector_access()` etc.).

bobbinth (changes requested):
Looks good! Thank you! This is still not a full review - but I left a few comments inline.

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,342 @@
...
+
+/// Unfolds a list comprehension into a vector of expressions.
+///
+/// Returns an error if there is an error while parsing any of the expressions in the expanded
+/// vector from the list comprehension.
+pub fn unfold_lc(
+    lc: &ListComprehension,
+    symbol_table: &SymbolTable,
+) -> Result<Vec<Expression>, SemanticError> {
+    let num_iterations = get_num_iterations(lc, symbol_table)?;
```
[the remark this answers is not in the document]
tohrnii: Thanks. I've gone with enforcing number of iterations greater than 1 but maybe even 1 should be syntactically correct even if it's not very meaningful so I'm not sure what the right approach is here.

ir/src/constraints/graph.rs
tohrnii: Done

ir/src/constraints/mod.rs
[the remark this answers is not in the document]
grjte: I probably prefer `expressions` as well, but you could use `expr_vector` to capture both ideas

ir/src/constraints/mod.rs
```diff
@@ -6,6 +6,7 @@ use super::{
 use std::collections::{BTreeMap, BTreeSet};
 
 mod constraint;
+use air_script_core::Variable;
```
grjte, line 9: nit: this can be imported in `super` above

ir/src/constraints/list_comprehension.rs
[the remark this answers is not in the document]
grjte: In cases like this, I generally prefer the syntactically correct option, because it's more flexible for the user. I could imagine a case where we have a list comprehension as part of a evaluator which is reused over a variety of sets of imported trace column groups, most of which consist of multiple columns, but one of which consists of 1 column. In this case, things would break for the programmer and mean that they need to define a separate function for that one case, whereas it's something that we could probably handle cleanly (especially once there are more optimizations in the IR), allowing them to have a clearer code pattern

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,376 @@
...
+    if num_iterations <= 1 {
+        return Err(SemanticError::InvalidListComprehension(
+            "List comprehension must have more than one iteration.".to_string(),
```
grjte, line 21: nit: I would change this to "List comprehensions must have"

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,376 @@
...
+    let vector = (0..num_iterations)
+        .map(|i| parse_lc_expr(lc.expression(), &iterable_context, symbol_table, i))
+        .collect::<Result<Vec<_>, _>>()?;
+    Ok(vector)
+}
+
+/// Parses a list comprehension expression and creates an expression based on the index of the
+/// expression in the list comprehension.
+///
+/// Returns an error if there is an error while parsing the sub-expression.
```
grjte, line 35: nit: This should have an `# Errors` header

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,376 @@
...
+    VariableType, VectorAccess,
+};
+use std::collections::BTreeMap;
+
+type IterableContext = BTreeMap<Identifier, Iterable>;
+
+/// Unfolds a list comprehension into a vector of expressions.
+///
+/// Returns an error if there is an error while parsing any of the expressions in the expanded
+/// vector from the list comprehension.
```
grjte, line 13: nit: this should have an `# Errors` header

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,376 @@
...
+                    ListFoldingType::Prod(_) => Expression::Mul(Box::new(acc), Box::new(expr)),
+                };
+            }
+            Ok(acc)
+        }
+    }
+}
+
+/// Validates and returns the length of a list comprehension. Checks that the length of all iterables
+/// in the list comprehension is the same and returns an error if it's not.
```
grjte, line 194: nit: the error should be separated into an `# Errors` section. There are several other places with the same problem, so I won't comment them anymore, but please check the rest of the doc comments

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,376 @@
...
+/// the size of the vector.
+fn validate_access(i: usize, size: usize) -> Result<(), SemanticError> {
+    if i < size {
+        Ok(())
+    } else {
+        Err(SemanticError::InvalidListComprehension(
+            "Invalid access index".to_string(),
+        ))
+    }
+}
```
grjte, line 253: I think this should be an `IndexOutOfRange` instead, and then the error message could state that it happens in a list comprehension and provide useful details. Wdyt?

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,376 @@
...
+
+/// Builds an [IterableContext] from a given list comprehension.
+fn build_iterable_context(lc: &ListComprehension) -> Result<IterableContext, SemanticError> {
```
grjte, line 256: There's no documentation of `IterableContext` or of `Iterable`, and I think this is starting to get a bit hard to understand. Can you add brief docs for those so that this doc comment will be more helpful?

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,376 @@
...
+    let mut iterable_context = IterableContext::new();
+    for (member, iterable) in lc.context() {
+        if iterable_context
+            .insert(member.clone(), iterable.clone())
+            .is_some()
+        {
+            return Err(SemanticError::InvalidListComprehension(
+                "Duplicate member in list comprehension".to_string(),
+            ));
```
grjte, line 265: I think maybe this error should be more descriptive so it will more be helpful to anyone unfamiliar with the IR internals

ir/src/constraints/graph.rs
```diff
@@ -1,14 +1,16 @@
 use super::{
-    BTreeMap, ConstantType, ConstraintDomain, ExprDetails, Expression, Identifier, IdentifierType,
-    IndexedTraceAccess, IntegrityConstraintDegree, MatrixAccess, SemanticError, SymbolTable,
-    TraceSegment, VariableRoots, VariableType, VectorAccess,
+    list_comprehension::unfold_lc, BTreeMap, ConstantType, ConstraintDomain, ExprDetails,
+    Expression, Identifier, IdentifierType, IndexedTraceAccess, IntegrityConstraintDegree,
+    MatrixAccess, SemanticError, SymbolTable, TraceSegment, VariableRoots, VariableType,
+    VectorAccess,
 };
+use air_script_core::ListFoldingType;
```
grjte, line 7: nit: I'd prefer to import this at the top level of the crate and then import from `super` here.

ir/src/constraints/graph.rs
```diff
@@ -398,6 +405,11 @@ impl AlgebraicGraph {
                 Ok(ExprDetails::new(node_index, trace_segment, domain))
             }
             IdentifierType::PublicInput(_) => {
+                if !domain.is_boundary() {
```
grjte, line 408: now that `is_boundary` is defined, it can be updated in `ConstrainedBoundary` as well. I think it makes sense to do this at the same time as defining the method, unless you object

ir/src/constraints/graph.rs
```diff
@@ -485,6 +497,42 @@ impl AlgebraicGraph {
             ))),
         }
     }
+
+    /// Inserts a list folding expression into the graph and returns the resulting expression
+    /// details.
+    ///
+    /// # Errors
+    /// Panics if the list is empty.
+    /// Returns an error if the list cannot be unfolded properly.
```
grjte, line 506: should we give these bullet points? I think we usually do when there are multiple error cases, right?

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,376 @@
...
+) -> Result<Expression, SemanticError> {
+    match ident_type {
+        IdentifierType::TraceColumns(trace_columns) => {
+            validate_access(i, trace_columns.size())?;
+            Ok(Expression::NamedTraceAccess(NamedTraceAccess::new(
+                ident.clone(),
+                i,
+                CURRENT_ROW,
+            )))
+        }
```
grjte, line 291: I know it would mean passing the symbol table, but it might be preferable to create an `IndexedTraceAccess` at this point. Since it's the IR we have all the information, and if we create a `NamedTraceAccess` we just have to convert it later.

grjte (changes requested):
This is a partial review. I need to spend more time looking at the `list_comprehension` module. 

As a general comment, I think it would be easier to read this file if the doc comments were more detailed and/or provided examples - in most cases, the doc comments explain what's happening in terms of internal terminology that we've chosen or refer to undocumented structs, so it's a bit hard to understand one thing without already understanding everything. Ideally, everything that's happening could be understood by just reading the function doc comments, and I think that's not really the case right now (or not easily)

ir/src/constraints/list_comprehension.rs
tohrnii: Makes sense. I generally take the approach of having a separate section if there are more than one error cases but I don't think I've been consistent with that either. I'll add an errors section everywhere.

ir/src/constraints/list_comprehension.rs
tohrnii: Makes sense. Fixed.

ir/src/constraints/list_comprehension.rs
tohrnii: I've changed it to also print the member name in the error message.

ir/src/constraints/list_comprehension.rs
tohrnii: Makes sense. Done.

tohrnii:
@grjte I've addressed your comments except adding more detailed docs. I'm working on them now.

One question is if we should rename `unfold_lc` to something like `expand_lc` (or something else) because unfold might be confusing since we also have list folding?

grjte (changes requested):
Thanks @tohrnii I think it's looking really good. The comments you've added are very helpful. It feels quite clear to read through it at this point. I left several things inline, but they're all minor.

One final note is that I feel like we might be able to improve the pattern for the `list_comprehension` module using a struct + methods instead of all of the independent functions. I think it would be hard to do this now, but that it might be worth a look after #150 (or maybe as part of it)

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,397 @@
...
+    iterable: &Iterable,
+) -> Result<usize, SemanticError> {
+    match iterable {
+        Iterable::Identifier(ident) => {
+            let ident_type = symbol_table.get_type(ident.name())?;
+            match ident_type {
+                IdentifierType::Variable(_, var_type) => match var_type.value() {
+                    VariableType::Vector(vector) => Ok(vector.len()),
+                    _ => Err(SemanticError::InvalidListComprehension(
+                        "List comprehensions only allowed for vector variables.".to_string(),
```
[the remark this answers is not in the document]
tohrnii: Thanks. Changed it to 
```
format!("Variable {} should be a vector for a valid list comprehension.", ident.name()),
```
instead.

ir/src/tests/list_comprehension.rs
```diff
@@ -0,0 +1,306 @@
...
+        enf c[2].first = 0
+    
+    integrity_constraints:
+        let y = [col'^7 - col for col in c]
+        enf clk = y[1]";
+
+    let parsed = parse(source).expect("Parsing failed");
+    let result = AirIR::new(&parsed);
+    assert!(result.is_ok());
+}
```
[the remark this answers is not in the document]
tohrnii: Thanks. I think my thinking at the time was that they are treated differently at the parser level so named accordingly. I have now merged both the test cases.

tohrnii:
Thank you @bobbinth and @grjte for the reviews. Made suggested changes.

ir/src/tests/list_comprehension.rs
```diff
@@ -0,0 +1,306 @@
...
+    boundary_constraints:
+        enf c[2].first = 0
+    
+    integrity_constraints:
+        enf clk = 2^ctx";
+
+    let parsed = parse(source).expect("Parsing failed");
+    let result = AirIR::new(&parsed);
+    assert!(result.is_err());
+}
```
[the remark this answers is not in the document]
tohrnii: moved it to `integrity_constraints` below an exp related test.

air-script/tests/list_comprehension/list_comprehension.rs
```diff
@@ -0,0 +1,96 @@
+use winter_air::{Air, AirContext, Assertion, AuxTraceRandElements, EvaluationFrame, ProofOptions as WinterProofOptions, TransitionConstraintDegree, TraceInfo};
+use winter_math::fields::f64::BaseElement as Felt;
+use winter_math::{ExtensionOf, FieldElement};
+use winter_utils::collections::Vec;
+use winter_utils::{ByteWriter, Serializable};
```
grjte, line 5: nit: these could be condensed into 2 lines

ir/src/constraints/list_comprehension.rs
```diff
@@ -0,0 +1,399 @@
...
+                        validate_access(i, trace_columns.size())?;
+                        Ok(Expression::NamedTraceAccess(NamedTraceAccess::new(
+                            ident.clone(),
+                            range.start() + i,
+                            named_trace_access.row_offset(),
+                        )))
+                    }
+                    _ => Err(SemanticError::InvalidListComprehension(format!(
+                        "Iterable {ident} should be trace columns"
+                    )))?,
```
grjte, line 170: nit: this grammar doesn't sound quite right. Should this be "Iterable {ident} should be a trace column" or "Iterable {ident} should contain trace columns"?

grjte (approved):
LGTM, great work @tohrnii, thanks! I left 2 final non-blocking nits inline

air-script/tests/list_comprehension/list_comprehension.rs
tohrnii: I'm not sure how to do that. The import function takes a module path and type. So, currently we do:
```
scope.import("winter_utils::collections", "Vec")
```
If I move `collections` to the type argument:
```
scope.import("winter_utils",  "collections::Vec")
```
this generates output:
```
use winter_utils::{collections, ByteWriter, Serializable};
```
This is what I had done earlier to combine the imports before Bobbin pointed out the problem.

Looking at the import constructor it just joins the two strings:
```
pub fn new(path: &str, ty: &str) -> Self {
        Import {
            line: format!("{}::{}", path, ty),
            vis: None,
        }
    }
```
So, will have to look into it more. If you have any ideas on how to achieve this, please let me know otherwise I'll dig into this more later.

bobbinth (approved):
All looks good! Thank you!

air-script/tests/list_comprehension/list_comprehension.rs
grjte: Ah yeah, makes sense. It's not worth any time before the next version, so let's just leave it.

PR #144, opened 2023-02-13
Restore boundary constraint validation

grjte:
This PR restores validation of boundary constraint rules in the IR, addressing one of the outstanding items from #126


ir/src/lib.rs
```diff
@@ -164,7 +164,7 @@ impl AirIR {
         self.constraints.num_boundary_constraints(trace_segment)
     }
 
-    pub fn boundary_constraints(&self, trace_segment: TraceSegment) -> &[ConstraintRoot] {
+    pub fn boundary_constraints(&self, trace_segment: TraceSegment) -> Vec<&ConstraintRoot> {
         self.constraints.boundary_constraints(trace_segment)
     }
```
bobbinth, line 169: Why do we need to return a `Vec` from here instead of a slice?

bobbinth: Oh - n/m.

ir/src/constraints/mod.rs
```diff
@@ -72,10 +73,11 @@ impl ExprDetails {
...
 pub(super) struct Constraints {
-    /// Constraint roots for all boundary constraints against the execution trace, by trace segment,
-    /// where boundary constraints are any constraints that apply to either the first or the last
-    /// row of the trace.
-    boundary_constraints: Vec<Vec<ConstraintRoot>>,
+    /// Constraint roots for all boundary constraints against the execution trace, grouped by trace
+    /// segment. They map the constrained boundary within the trace to the [ConstraintRoot] that
+    /// contains its [ConstraintDomain] and entry [NodeIndex] in the graph. Boundary constraints are
+    /// any constraints that apply to either the first or the last row of the trace.
+    boundary_constraints: Vec<BTreeMap<ConstrainedBoundary, ConstraintRoot>>,
```
bobbinth, line 80: The information here seems somewhat redundant - both `ConstrainedBoundary` and `ConstraintRoot` contain domain info. Moreover, `ConstrainedBoundary` also contains `trace_segment` which seems to be the index into top-level vector here.

It seems to me that putting `trace_segment` into `ConstrainedBoundary` may not be necessary? But more generally, it seems to me that the `BTreeMap` and `ConstrainedBoundary` struct are used primarily for error-checking purposes (to make sure we don't have multiple boundary constraints defined against the same boundary of the same column). If so, I'd rather use something else for error checking and keep this as `Vec<Vec<ConstraintRoot>>`. For example, we could introduce another instance variable similar to `variable_roots` but to keep track of which boundary constraints are already "taken".

Separately, looking at the structure below, I wonder if we need to keep two separate member variables for validity and transition constraints. Could we not track them in a single vector as:
```Rust
pub(super) struct Constraints {
    ...
    integrity_constraints: Vec<Vec<ConstraintRoot>>,
    ...
}
```
The information contained in `ConstraintRoot` should be sufficient to differentiate between integrity and transition constraints. Or am I missing something?

Anyways, this is probably a separate issue not related to this PR.

bobbinth (changes requested):
Looks good! Thank you! I left a couple of small questions inline, and also one comment about potentially restructuring how we track boundary constraints.

ir/src/symbol_table.rs
```diff
@@ -212,6 +212,16 @@ impl SymbolTable {
...
 
+    /// Returns a vector containing the widths of all trace segments.
+    pub(super) fn segment_widths(&self) -> &Vec<u16> {
+        &self.segment_widths
+    }
+
+    /// Returns the number of random values that were specified for this AIR.
+    pub(super) fn num_random_values(&self) -> u16 {
+        self.num_random_values
+    }
```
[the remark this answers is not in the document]
grjte: Yes, it should, but random values for subsequent trace segments aren't yet handled. This was done as part of the recent random values work, and I think making that change can be done as part of supporting more trace segments, unless you disagree

ir/src/constraints/mod.rs
grjte: > It seems to me that putting trace_segment into ConstrainedBoundary may not be necessary?

Yeah, I originally implemented it this way and then got confused & changed it at the last minute, which I have been thinking was a mistake. I'll remove it from there.

> But more generally, it seems to me that the BTreeMap and ConstrainedBoundary struct are used primarily for error-checking purposes (to make sure we don't have multiple boundary constraints defined against the same boundary of the same column). If so, I'd rather use something else for error checking

I was on the fence about this and considered this approach too. The reason I did it this way was to avoid tracking the same thing twice, but I'm happy to try something else. In fact, if we keep the `ConstrainedBoundary` as it is (with the `trace_segment`) then this could be tracked more cleanly than my original thought here.

> Separately, looking at the structure below, I wonder if we need to keep two separate member variables for validity and transition constraints. Could we not track them in a single vector

Yes, we could track them as a single vector, but it seemed like it gave a bit more flexibility to have them separated if we want to access them separately without having to filter the vectors. I don't feel very strongly, but there's no cost to separating them as we process them, and there's a small cost to doing so later, which is why I had previously done it this way.

grjte: I've updated this - please let me know what you think of the adjusted approach for tracking constrained boundaries

tohrnii (approved):
Thank you @grjte. Looks good to me. I just left minor one nit inline.

ir/src/constraints/mod.rs
```diff
@@ -93,6 +94,9 @@ pub(super) struct Constraints {
     /// vector or a matrix, a new root is added with a key equal to the [VariableValue] of the
     /// element.
     variable_roots: VariableRoots,
+
+    /// A set of all boundaries which have been constrained.
```
bobbinth, line 98: nit: I'd add a sentence explaining what it is being used for (otherwise it is not immediately clear why we need this in addition to `boundary_constraints`).

bobbinth (approved):
All looks good! Thank you!

ir/src/constraints/mod.rs
grjte: Updated!

ir/src/constraints/graph.rs
```diff
@@ -170,36 +169,24 @@ impl AlgebraicGraph {
...
-    /// Converts a [NamedTraceAccess] element into an [IndexedTraceAccess] by its identifier name,
-    /// trace segment, and row offset, then adds it to the graph.
+    /// Adds a trace element access to the graph and returns the node index, trace segment, and row
+    /// offset.
     ///
     /// # Errors
     /// Returns an error if:
-    /// - The identifier was not declared as a trace column type.
-    pub(super) fn insert_named_trace_access(
+    /// - The identifier is greater than overall number of columns in segment.
```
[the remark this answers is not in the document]
grjte: Yes, thank you! Updated!

Issue #132, opened 2023-02-01
Allow to access a vector from a matrix


tohrnii:
Trying to access a vector from a matrix results in some problems currently. The parser recognizes it incorrectly and the error thrown at the IR level is incorrect. We should be able to support statements like these:

```
let a = [[1, 2], [3, 4]]
let b = a[0]
```

This should also be handled for list comprehension.

@grjte raised this point in an earlier PR review but it was left unaddressed.

grjte:
Closed by #147

PR #147, opened 2023-02-15
Allow access to vector in matrix variable

Overcastan:
Addressing: #132 
This PR makes possible to:
- create vector variables as row of the matrix variable
- create matrix variables as vector of matrix rows and vector variables

This PR fixes all bugs which do not require parsing modification.

Cases that were fixed:
1. Creating vector variable as row of the matrix and copying this variable:
```
integrity_constraints:
    let a = [[1, 2], [3, 4]]
    let b = a[0] 
    let c = b
    enf clk' = b[1] + c[0]
```
2. Creating matrix variable as vector of matrix rows and vector variables, and copying this variable:
```
integrity_constraints:
    let a = [[1, 2], [3, 4]]
    let b = a[0] 
    let c = [a[1], b]
    let d = c
    enf clk' = d[0][0] + c[1][1]
```

### DIFF
--- a/docs/src/description/variables.md
+++ b/docs/src/description/variables.md
@@ -44,8 +44,10 @@ Currently, it is not possible to:
     ```
     ...
     boundary_constraints:
-        let a = [1, 2]
-        let b = [a, [3, 4]]  <-- `b` consists of array `[3, 4]` and reference to array `a`
+        let a = [[1,2], [3,4]]  # <-- this is allowed
+        let b = [1, 2]
+        let c = [a[1], b]  # <-- this is allowed
+        let d = [b, [3, 4]]  # <-- this is not allowed, because `d` consists of array `[3, 4]` and reference to array `b`
         enf ...
     ...
     ```
@@ -59,9 +61,9 @@ Currently, it is not possible to:
         let a = [[1, 2], [3, 4]]
         let b = [5, 6]
         let c = 7
-        let d = [e for e in [8, c, b[0], a[0][1]]]  <-- source array is a plain array
-        let f = [g for g in a[1]]  <-- source is a matirx row
-        let h = [i for i in a[0][0..2]]  <-- source is a range in matrix row
+        let d = [e for e in [8, 9, 10]]  # <-- source array is an inlined vector
+        let f = [g for g in a[1]]  # <-- source is a matrix row
+        let h = [i for i in a[0][0..2]]  # <-- source is a range in matrix row
         enf ...
     ...
     ```
--- a/docs/src/description/variables.md
+++ b/docs/src/description/variables.md
@@ -45,11 +45,11 @@ Currently, it is not possible to:
     ...
     boundary_constraints:
         let a = [1, 2]
-        let b = [a, [3, 4]]  <-- b consists of array `[3, 4]` and reference to array `a`
+        let b = [a, [3, 4]]  <-- `b` consists of array `[3, 4]` and reference to array `a`
         enf ...
     ...
     ```
-2. Create variables with list comprehension for which the source array is a plain array, a matrix row, or a range in matrix row.
+2. Create variables with list comprehension for which the source array is a inlined vector, a matrix row, or a range in matrix row.
 
     Example: 
 
--- a/docs/src/description/variables.md
+++ b/docs/src/description/variables.md
@@ -34,6 +34,38 @@ integrity_constraints:
     enf a' = z[0][0] + z[0][1] + z[1][0] + z[1][1]
 ```
 
+### Syntax restriction for local variables
+Currently, it is not possible to:
+
+1. Create matrices containing both arrays and references to arrays.
+
+    Example:
+
+    ```
+    ...
+    boundary_constraints:
+        let a = [1, 2]
+        let b = [a, [3, 4]]  <-- b consists of array `[3, 4]` and reference to array `a`
+        enf ...
+    ...
+    ```
+2. Create variables with list comprehension for which the source array is a plain array, a matrix row, or a range in matrix row.
+
+    Example: 
+
+    ```
+    ...
+    integrity_constraints:
+        let a = [[1, 2], [3, 4]]
+        let b = [5, 6]
+        let c = 7
+        let d = [e for e in [8, c, b[0], a[0][1]]]  <-- source array is a plain array
+        let f = [g for g in a[1]]  <-- source is a matirx row
+        let h = [i for i in a[0][0..2]]  <-- source is a range in matrix row
+        enf ...
+    ...
+    ```
+
 ## Built-in variables
 
 Built-in variables are identified by the starting character `$`. There are two built-in variables:
--- a/ir/src/constraints/graph.rs
+++ b/ir/src/constraints/graph.rs
@@ -405,23 +405,57 @@ impl AlgebraicGraph {
             IdentifierType::Constant(ConstantType::Vector(_)) => {
                 self.insert_constant(ConstantValue::Vector(vector_access.clone()), domain)
             }
-            IdentifierType::Variable(scope, variable) => {
-                if let VariableType::Vector(vector) = variable.value() {
-                    self.insert_variable(
-                        symbol_table,
-                        variable_roots,
-                        domain,
-                        *scope,
-                        VariableValue::Vector(vector_access.clone()),
-                        &vector[vector_access.idx()],
-                    )
-                } else {
-                    Err(SemanticError::invalid_vector_access(
-                        vector_access,
-                        symbol_type,
-                    ))
+            IdentifierType::Variable(scope, variable) => match variable.value() {
+                VariableType::Vector(vector) => self.insert_variable(
+                    symbol_table,
+                    variable_roots,
+                    domain,
+                    *scope,
+                    VariableValue::Vector(vector_access.clone()),
+                    &vector[vector_access.idx()],
+                ),
+                VariableType::Scalar(matrix_vector_access_expr) => {
+                    match matrix_vector_access_expr {
+                        Expression::VectorAccess(matrix_row_access) => {
+                            let matrix_access = Expression::MatrixAccess(MatrixAccess::new(
+                                Identifier(matrix_row_access.name().to_string()),
+                                matrix_row_access.idx(),
+                                vector_access.idx(),
+                            ));
+                            self.insert_variable(
+                                symbol_table,
+                                variable_roots,
+                                domain,
+                                *scope,
+                                VariableValue::Vector(vector_access.clone()),
+                                &matrix_access,
+                            )
+                        }
+                        Expression::Elem(elem) => {
+                            let equal_vector_access = Expression::VectorAccess(VectorAccess::new(
+                                elem.clone(),
+                                vector_access.idx(),
+                            ));
+                            self.insert_variable(
+                                symbol_table,
+                                variable_roots,
+                                domain,
+                                *scope,
+                                VariableValue::Vector(vector_access.clone()),
+                                &equal_vector_access,
+                            )
+                        }
+                        _ => Err(SemanticError::invalid_vector_access(
+                            vector_access,
+                            symbol_type,
+                        )),
+                    }
                 }
-            }
+                _ => Err(SemanticError::invalid_vector_access(
+                    vector_access,
+                    symbol_type,
+                )),
+            },
             IdentifierType::TraceColumns(columns) => {
                 let trace_segment = columns.trace_segment();
                 let trace_access = IndexedTraceAccess::new(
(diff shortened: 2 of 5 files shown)


tohrnii (changes requested):
@Overcastan Thank you. Looks good. I had one preliminary question. It seems to me that this doesn't work if we want to create a new matrix from rows of another matrix for example. Like:
```
let a = [[1, 2], [3, 4]]
let b = [a[1], a[0]]
enf clk' = b[1][0]
```

Can you please check if it's possible to support these kinds of statements?

tohrnii:
I think one other way to handle this might be to just store it as a `Scalar(vector_access)` in the symbol table and then when it is accessed while inserting the expression, handle this case differently in the `insert_vector_access` function. This would remove the necessity to clone the vector but I don't think cloning is a big problem here (but I could be wrong) so maybe we should go with whichever solution makes it easy for us to handle all the possible cases.

tohrnii:
To list down some possible statements that we need to handle for now:

```
let a = [[1, 2], [3, 4]]
let b = a[0]
let c = [a[1], a[0]]
let d = [e for e in a[1]]
let f = sum([g for g in a[1])
```

Please let me know if I missed any.

tohrnii:
@Overcastan I think there is an issue with this approach. This doesn't work for statements like:

```
let a = [[1, 2], [3, 4]]
let c = [a[1], a[0], [5, 6]]
```

Sorry, I didn't think about this case earlier. I think we will have to make changes to the parser.

Overcastan:
Should we manage Vectors with different types, like `[5, 6, a[1][0]]`? Or, in other words, should we handle statements like 
```
let a = [[1, 2, 3], [4, 5, 6]]
let c = [a[1], a[0], [7, 8, a[1][0]]]
```

tohrnii:
@Overcastan I think so. We already support vectors like `[5, 6, a[1][0]]`. In my mind, we should just change the parsing rules for matrices to allow rows from other matrices or inlined vectors.

So, we could change this parsing rule: 
```
"[" <matrix_value: CommaElems<Vector<BoundaryExpr>>> "]"
```
and instead define another enum for matrix rows where they could be identifiers (referring to declared vectors), vector access (referring to row in another matrix) or inlined vectors.

It's possible I'm missing other cases which don't get handled with this or maybe there's a better way to handle this.

tohrnii:
Actually, I think this might complicate things because we will have matrices defined as vectors in the AST. For eg,
```
let a = [1, 2, 3]
let b = [4, 5, 6]
let c = [[1, 2, 3], [4, 5, 6]]
let d = [a, b, c[1]]
```
This would be defined as a vector at the parser level.
```
```
let a = [1, 2, 3]
let b = [4, 5, 6]
let c = [a, b, [4, 5, 6]]
```
This would be defined as a matrix at the parser level.

Then the IR could get complicated. So, if it makes thing easier, we should probably restrict the syntax here. Maybe we could make it so that matrix could either contain all inlined vectors or all expressions. So, something like `[a, b, [4, 5, 6]]` would be invalid. If users wanted to write a statement like this, they could just define another intermediate variable for `[4, 5, 6]` and use it that way.

@grjte Wdyt?

**Comment by tohrnii** [comment]
If we go with restricting syntax here, then we probably don't need to update the parser and only need to add some rules to your current PR.

I think we should support statements like these for now:

```
let a = [1, 2, 3]
let b = [4, 5, 6]
let c = [[1, 2, 3], [4, 5, 6]]
let d = c[0]
let e = [a + b for (a, b) in (a, b)]
let f = [a, b, c[0]]
let g = [a, b, d]
let h = [a, b, e]
```

Supporting non-nested list comprehension doesn't require any extra work for this PR since currently we just unroll it and store it as a vector in the symbol table.

**Comment by Overcastan** [comment]
Current approach a little bit differently fills `variable_roots`, and I'm not sure is this a problem or not.

**Comment by grjte** [comment]
@tohrnii @Overcastan sorry I didn't see the question above about restricting syntax earlier. Yes, I think this is fine, especially for this PR. If we want to allow the more flexible syntax that can be addressed separately as its own issue and PR

File: ir/src/constraints/graph.rs [hunk 1]
```diff
@@ -399,23 +399,57 @@ impl AlgebraicGraph {
...
+            IdentifierType::Variable(scope, variable) => match variable.value() {
+                VariableType::Vector(vector) => self.insert_variable(
+                    symbol_table,
+                    variable_roots,
+                    domain,
+                    *scope,
+                    VariableValue::Vector(vector_access.clone()),
+                    &vector[vector_access.idx()],
+                ),
+                VariableType::Scalar(matrix_vector_access_expr) => {
```
grjte, line 411: nit: let's put `Scalar` before `Vector`

ir/src/constraints/graph.rs
```diff
@@ -399,23 +399,57 @@ impl AlgebraicGraph {
...
+                            self.insert_variable(
+                                symbol_table,
+                                variable_roots,
+                                domain,
+                                *scope,
+                                VariableValue::Vector(vector_access.clone()),
+                                &matrix_access,
+                            )
+                        }
+                        Expression::Elem(elem) => {
```
grjte, line 428: nit: let's put `Elem` before `VectorAccess`

ir/src/constraints/graph.rs
```diff
@@ -463,23 +497,81 @@ impl AlgebraicGraph {
...
+            IdentifierType::Variable(scope, variable) => match variable.value() {
+                VariableType::Matrix(matrix) => self.insert_variable(
+                    symbol_table,
+                    variable_roots,
+                    domain,
+                    *scope,
+                    VariableValue::Matrix(matrix_access.clone()),
+                    &matrix[matrix_access.row_idx()][matrix_access.col_idx()],
+                ),
+                VariableType::Vector(vector) => {
```
grjte, line 509: nit: let's put `Scalar` before `Vector` before `Matrix`

ir/src/constraints/graph.rs
```diff
@@ -463,23 +497,81 @@ impl AlgebraicGraph {
...
+                            self.insert_variable(
+                                symbol_table,
+                                variable_roots,
+                                domain,
+                                *scope,
+                                VariableValue::Matrix(matrix_access.clone()),
+                                &internal_matrix_access,
+                            )
+                        }
+                        Expression::Elem(elem) => {
```
grjte, line 528: nit: let's put `Elem` before `VectorAccess`

grjte (changes requested):
Thanks @Overcastan this looks good! I left a few nits inline. Additionally, I think we should update the docs about variables to describe the syntax restriction that you and @tohrnii agreed on

docs/src/description/variables.md
```diff
@@ -34,6 +34,38 @@ integrity_constraints:
...
+
+    Example:
+
+    ```
+    ...
+    boundary_constraints:
+        let a = [1, 2]
+        let b = [a, [3, 4]]  <-- b consists of array `[3, 4]` and reference to array `a`
+        enf ...
+    ...
```
tohrnii, line 50: I think this format looks great!

docs/src/description/variables.md
```diff
@@ -34,6 +34,38 @@ integrity_constraints:
...
+    ```
+2. Create variables with list comprehension for which the source array is a plain array, a matrix row, or a range in matrix row.
```
tohrnii, line 52: minor nit: Should we say "inlined vector" instead of "plain array"?
cc: @grjte

ir/src/tests/variables.rs
```diff
@@ -121,6 +121,52 @@ fn ic_with_variables() {
...
+        let e = d
+        enf clk' = c[0] + e[2][0] + e[0][1]";
+
+    let parsed = parse(source).expect("Parsing failed");
+
+    let result = AirIR::new(&parsed);
+    assert!(result.is_ok());
+}
+
+// We can not parse matrix variable that consists of array and non-array elements
```
tohrnii, line 148: minor nit: I think the comment should probably go inside the function body.

tohrnii (changes requested):
Thank you @Overcastan, looks good to me. I left a few very minor nits. It would be great to have parser tests for expressions we don't support.

tohrnii (approved):
Thank you @Overcastan, looks good to me.

docs/src/description/variables.md
```diff
@@ -34,6 +34,38 @@ integrity_constraints:
...
+
+1. Create matrices containing both arrays and references to arrays.
+
+    Example:
+
+    ```
+    ...
+    boundary_constraints:
+        let a = [1, 2]
+        let b = [a, [3, 4]]  <-- `b` consists of array `[3, 4]` and reference to array `a`
```
grjte, line 48: For clarity, it might be good to have examples of the things that are ok, e.g.

```
let a = [[1,2], [3,4]] # <-- this is allowed
let b = [1, 2]
let c = [3, 4]
let d = [b, c]  # <-- this is allowed
let e = [b, [3, 4]]  # <-- this is not allowed, because `b` consists of array `[3, 4]` and reference to array `a`
```

nit: I would put a `#` before the comment:

```
      let b = [a, [3, 4]]  # <-- `b` consists of array `[3, 4]` and reference to array `a`
```

docs/src/description/variables.md
```diff
@@ -34,6 +34,38 @@ integrity_constraints:
...
+
+    ```
+    ...
+    integrity_constraints:
+        let a = [[1, 2], [3, 4]]
+        let b = [5, 6]
+        let c = 7
+        let d = [e for e in [8, c, b[0], a[0][1]]]  <-- source array is a plain array
+        let f = [g for g in a[1]]  <-- source is a matirx row
+        let h = [i for i in a[0][0..2]]  <-- source is a range in matrix row
```
grjte, line 64: in the declaration of `d` there are a lot of things going on and I think it's not clear what it is that isn't allowed. Can you simplify this example?

nits: 
1. I would put a `#` before the comments. 
2. I would change "plain array" to "inline vector"
3. There's a typo in "matrix"

```
        let d = [e for e in [8, c, b[0], a[0][1]]]  # <-- source array is an inline vector
        let f = [g for g in a[1]]  # <-- source is a matrix row
        let h = [i for i in a[0][0..2]]  # <-- source is a range in matrix row
```

ir/src/symbol_table.rs
```diff
@@ -366,17 +366,18 @@ impl SymbolTable {
                 validate_matrix_access(matrix_access, matrix.len(), matrix[0].len())?;
                 Ok(symbol_type)
             }
-            IdentifierType::Variable(_, variable) => {
-                if let VariableType::Matrix(matrix) = variable.value() {
+            IdentifierType::Variable(_, variable) => match variable.value() {
+                VariableType::Scalar(_) => Ok(symbol_type),
+                VariableType::Vector(_) => Ok(symbol_type),
```
grjte, line 371: nit: we should be able to combine these

```
    VariableType::Scalar(_) | VariableType::Vector(_) => Ok(symbol_type),
```

ir/src/tests/variables.rs
```diff
@@ -121,6 +121,52 @@ fn ic_with_variables() {
...
+        stack_inputs: [16]
+    boundary_constraints:
+        enf clk.first = 7
+        enf clk.last = 8
+    integrity_constraints:
+        let a = [[1, 2, 3, 4], [5, 6, 7, 8]]
+        let b = [9, 10]
+        let c = 12
+        let c = [a[0], [11, c, b[1], a[1][0]]]
+        d = [[11, c, b[1], a[1][0]], a[0]]
```
grjte, line 164: was the duplicate declaration of c here intentional?

parser/src/parser/tests/variables.rs
```diff
@@ -40,3 +40,13 @@ fn err_invalid_matrix_element() {
...
     build_parse_test!(source).expect_unrecognized_token();
 }
+
+#[test]
+fn err_matrix_variable_from_vector_and_reference() {
+    let source = "integrity_constraints:
+    let a = [[1, 2], [3, 4]]
+    let b = [5, 6]
+    let c = [b, [7, 8]]
+    let d = [[7, 8], a[0]]";
```
grjte, line 50: I believe this line with `let d` never gets tested, since it fails on the line above, so I think we should probably split this into 2 tests

grjte (changes requested):
Thanks @Overcastan, looks good! Thanks for the docs as well - I left a couple of clarity points to make sure that people are still clear on what's accepted syntax and a couple of other small nits/questions inline

docs/src/description/variables.md
grjte: It's up to you - I think that would be nice to have, but since list comprehension has so many more options I might end up being a lot. It would probably be helpful though

grjte (approved):
Looks good, thank you!

PR #148, opened 2023-02-16
fix: fix codegen for aux transition constraints that use main cols

tohrnii:
This PR refactors the codegen to correctly generate code for constraints that use both aux trace columns and main trace columns.

### DIFF
--- a/codegen/winterfell/src/air/graph.rs
+++ b/codegen/winterfell/src/air/graph.rs
@@ -1,3 +1,5 @@
+use ir::IndexedTraceAccess;
+
 use super::{AirIR, ConstantValue, ElemType, IntegrityConstraintDegree, NodeIndex, Operation};
 
 // RUST STRING GENERATION FOR THE CONSTRAINT GRAPH
@@ -26,6 +28,26 @@ impl Codegen for IntegrityConstraintDegree {
                 cycles
             )
         }
+    }
+}
+
+impl Codegen for IndexedTraceAccess {
+    fn to_string(&self, _ir: &AirIR, _elem_type: ElemType) -> String {
+        let frame = if let 0 = self.trace_segment() {
+            "main"
+        } else {
+            "aux"
+        };
+        let row_offset = match self.row_offset() {
+            0 => {
+                format!("current[{}]", self.col_idx())
+            }
+            1 => {
+                format!("next[{}]", self.col_idx())
+            }
+            _ => panic!("Winterfell doesn't support row offsets greater than 1."),
+        };
+        format!("{}_{}", frame, row_offset)
     }
 }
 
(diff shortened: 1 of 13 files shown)


codegen/winterfell/src/air/graph.rs
```diff
@@ -30,15 +32,15 @@ impl Codegen for IntegrityConstraintDegree {
...
         let op = ir.constraint_graph().node(self).op();
-        op.to_string(ir, elem_type)
+        op.to_string(ir, elem_type, trace_segment)
     }
 }
 
 impl Codegen for Operation {
     // TODO: Only add parentheses in Add and Mul if the expression is an arithmetic operation.
-    fn to_string(&self, ir: &AirIR, elem_type: ElemType) -> String {
+    fn to_string(&self, ir: &AirIR, elem_type: ElemType, trace_segment: TraceSegment) -> String {
```
bobbinth, line 43: Why do we need to pass this as a parameter here? It seems like it is only used in `Operation::TraceElement(trace_access) => {` branch and there `trace_access` already contains `trace_segment` info. Or am I missing something?

tohrnii: This is the trace segment of the overall constraint instead of just that particular `trace_access` which would be different for aux transition constraints which access main trace columns. I think we can remove this if we are ok with changing the name of the main evaluation frame in "evaluate_transition" function from "frame" to "main_frame" and handle it the same as "evaluate_aux_transition" (i.e. `main_frame.current()[]`) but I didn't know if we wanted to change the name.

bobbinth: I think changing the name is better than passing an extra variable around.

Eventually, we should probably declare local variables like `main_current`, `main_next`, `aux_current`, `aux_next` - but this is probably for another PR.

bobbinth: That would be great! Would it be possible to output `aux_current` and `aux_next` only if we have auxiliary constraints? (i.e., when the number of segments is greater than 1)

tohrnii: Done.

codegen/winterfell/src/air/graph.rs
```diff
@@ -68,15 +68,23 @@ impl Codegen for Operation {
...
+                let row_offset = match trace_access.row_offset() {
+                    0 => {
+                        format!("current[{}]", trace_access.col_idx())
+                    }
+                    1 => {
+                        format!("next[{}]", trace_access.col_idx())
+                    }
+                    _ => panic!("Winterfell doesn't support row offsets greater than 1."),
+                };
+                format!("{}_{}", frame, row_offset)
```
bobbinth, line 86: This works - but I think implementing `Codegen` on `IndexedTraceAccess` would be cleaner.

tohrnii: Makes sense. Thanks. Done.

bobbinth (approved):
All looks good! Thank you!

Issue #138, opened 2023-02-08
Break up IR tests into multiple files


tohrnii:
IR tests file is getting quite large and should probably be broken up.

tohrnii:
Closed by #151

PR #151, opened 2023-02-17
refactor IR tests

grjte:
This PR includes a trivial clippy fix and a quick refactor to split the IR tests into modules to make them more manageable and easier to update later. This addresses #138


tohrnii (approved):
Thank you @grjte, looks good to me. One question is if we should replace `boundary_constraints` and `integrity_constraints` in function names to their abbreviated forms (`bc` and `ic`) to be consistent except for `boundary_constraints()` and `integrity_constraints()`? If you agree with this change maybe it can be done with this PR?

grjte:
> Thank you @grjte, looks good to me. One question is if we should replace `boundary_constraints` and `integrity_constraints` in function names to their abbreviated forms (`bc` and `ic`) to be consistent except for `boundary_constraints()` and `integrity_constraints()`? If you agree with this change maybe it can be done with this PR?

Do you mean the accessor functions? I would probably leave them as it is and keep this PR as a tests refactor

tohrnii:
No I actually meant just for test names. For eg, renaming `boundary_constraints_with_constants` to `bc_with_constants` etc.

grjte:
> No I actually meant just for test names. For eg, renaming `boundary_constraints_with_constants` to `bc_with_constants` etc.

Ah, I see. The way I had done it was to shorten it in the error case and leave it unshortened for the main test, but I don't feel strongly. I've changed it in all cases except when there's nothing else in the name (e.g. `boundary_constraints`)

Issue #153, opened 2023-02-18
List folding is incomplete


tohrnii:
Currently list folding doesn't work on vectors or intermediate variables representing list comprehension.

For example, the following doesn't work:

```
let a = [1, 2, 3]
let b = sum(a)
```

```
let a = sum([1, 2, 3])
```

```
let a = [b for b in b]
let c = sum(a)
```

tohrnii:
Closed by #154

PR #154, opened 2023-02-18
Add support for list folding on vectors and identifiers

tohrnii:
Addressing #153. This PR adds support for list folding on vectors and identifiers referring to constant vectors, variable vectors, trace columns or list comprehension.


ir/src/constraints/list_comprehension.rs
```diff
@@ -185,9 +187,18 @@ fn parse_list_folding(
...
+            let iterable_context =
+                if let ListFoldingValueType::ListComprehension(lc) = lf_value_type {
+                    build_iterable_context(lc)?
+                } else {
+                    BTreeMap::new()
+                };
+            assert!(
+                !list.is_empty(),
+                "List on which list folding is applied is empty."
+            );
```
grjte, line 201: I think maybe this should either return a `SemanticError` or be a `debug_assert`. What was your thinking here?

ir/src/error.rs
```diff
@@ -75,4 +76,14 @@ impl SemanticError {
             matrix_col_len
         ))
     }
+
+    pub(super) fn invalid_list_folding(
+        lf_value_type: &air_script_core::ListFoldingValueType,
+        symbol_type: &IdentifierType,
+    ) -> SemanticError {
+        SemanticError::InvalidListFolding(format!(
+            "Symbol type {} is not supported for List folding value type {:?}",
```
grjte, line 85: nit: I would change "...List folding..." to "...list folding..."

ir/src/constraints/helpers.rs
```diff
@@ -0,0 +1,44 @@
...
+///
+/// # Errors
+/// Returns an error if:
+/// - the list folding value is an identifier that does not exist in the symbol table
+/// - the list folding value is an identifier that does not refer to a vector
+pub fn build_list_from_list_folding_value(
+    lf_value_type: &ListFoldingValueType,
+    symbol_table: &SymbolTable,
+) -> Result<Vec<Expression>, SemanticError> {
+    match lf_value_type {
```
grjte, line 18: Should we put this in `list_folding.rs` instead? It would follow the structure of the `tests` more closely

ir/src/tests/list_folding.rs
```diff
@@ -0,0 +1,107 @@
...
+        enf c[2].first = 0
+    integrity_constraints:
+        let x = [a + c, 1, c[2] * d[2]]
+        let y = sum(x)
+        let z = prod(x)
+        enf clk = z - y";
+
+    let parsed = parse(source).expect("Parsing failed");
+    let result = AirIR::new(&parsed);
+    assert!(result.is_ok());
```
grjte, line 42: Do we allow the syntax `a+c`? I think we probably shouldn't - it should only be `a+c[i]` that's supported. Since this is passing, maybe there's a bug to fix (and a corresponding new test to add).

grjte (changes requested):
Thanks @tohrnii, this looks good to me overall. A few nits and 2 points

1. I left a comment inline - maybe we have a bug?
2. maybe it would be good to update the integration test or to split the list folding apart from the current list comprehension integration test and add a new integration test for list folding. Wdyt?

ir/src/tests/list_folding.rs
tohrnii: Thanks for catching this. I've included a fix in this PR.

ir/src/error.rs
```diff
@@ -75,4 +76,30 @@ impl SemanticError {
...
+            lf_value_type
+        ))
+    }
+
+    pub(super) fn single_trace_column_allowed(ident: &str) -> SemanticError {
+        SemanticError::InvalidUsage(format!(
+            "{} should be a single trace column and not a group.",
+            ident
+        ))
+    }
```
grjte, line 104: nit: I think there might be a better name for this. It's a type mismatch issue, so "single_trace_column_allowed" seems like a misnomer to me. Maybe "invalid_trace_binding" and the error message could be something like "Expected {} to be a binding to a single trace column."? In general, I want to move to the "binding" terminology in the future anyway

It might not be worth changing this now though. Maybe we should just revisit the errors again after this is merged and the refactors are done

ir/src/constraints/list_folding.rs
```diff
@@ -0,0 +1,67 @@
+use super::{
+    graph::CURRENT_ROW, list_comprehension::unfold_lc, IdentifierType, SemanticError, SymbolTable,
+};
+use air_script_core::{
+    ConstantType, Expression, IndexedTraceAccess, ListFoldingValueType, VariableType,
+};
```
grjte, line 6: nit: I'd prefer to import these from super so that we can see everything that is used from `air_script_core` by looking at the `lib.rs` file at the top of the crate

ir/src/error.rs
grjte, line 104: nit: can you update all of these for the new formatting?

e.g.
```
"Symbol type {symbol_type} is not supported for list folding value type {lf_value_type:?}"
```

Also, maybe it would be better to implement `Display` for the listfolding value type than to use debug printing here - but that could be a separate PR so that we can get this merged

grjte (approved):
Approved, thank you! I left a couple of minor nits inline, but they could be addressed later. I'll leave it up to you which to do now, if any

grjte:
@tohrnii let's wait until #172 is merged before merging this

ir/src/error.rs
tohrnii: Thanks. I've changed this for now but agreed we should revisit all the errors again.

tohrnii: I haven't implemented the Display trait yet. Will open a separate PR.

PR #155, opened 2023-02-19
fix(codegen): remove redundant parens from generated code

tohrnii:
Addressing #27


codegen/winterfell/src/air/graph.rs
```diff
@@ -100,30 +100,69 @@ impl Codegen for Operation {
...
+}
+
+/// Returns a string representation of a binary operation.
+fn binary_op_to_string(
+    l_idx: &NodeIndex,
+    r_idx: &NodeIndex,
+    ir: &AirIR,
+    binary_op: char,
+    elem_type: ElemType,
+) -> String {
```
grjte, line 137: nit: I'd probably pass `ir` first, since we do that in the other codegen functions

codegen/winterfell/src/air/graph.rs
```diff
@@ -100,30 +100,69 @@ impl Codegen for Operation {
...
+            let rhs = r_idx.to_string(ir, elem_type);
+            format!("{lhs} + {rhs}")
+        }
+        '-' => {
+            let lhs = l_idx.to_string(ir, elem_type);
+            let rhs = if is_leaf(r_idx, ir) {
+                r_idx.to_string(ir, elem_type)
+            } else {
+                format!("({})", r_idx.to_string(ir, elem_type))
+            };
```
grjte, line 150: nit: we could actually be even stronger here. The only time this matters is with `Operation::Add`, right?

codegen/winterfell/src/air/graph.rs
```diff
@@ -100,30 +100,69 @@ impl Codegen for Operation {
...
+            let lhs = if is_leaf(l_idx, ir) {
+                l_idx.to_string(ir, elem_type)
+            } else {
+                format!("({})", l_idx.to_string(ir, elem_type))
+            };
+            let rhs = if is_leaf(r_idx, ir) {
+                r_idx.to_string(ir, elem_type)
+            } else {
+                format!("({})", r_idx.to_string(ir, elem_type))
+            };
```
grjte, line 163: nit: similar to above - on the rhs we only care about `Operation::Add` and `Operation::Sub`. I think we just don't have any examples of e.g. 2 multiplications in a row, so there aren't examples of the redundancy.

grjte (changes requested):
Looks good, thanks @tohrnii! I think we can reduce the redundancy a bit more. See my comments inline

codegen/winterfell/src/air/graph.rs
tohrnii: We had an example of two simultaneous multiplications in bitwise constraints. I missed it before. Thanks for catching this. I've added a fix but that doesn't reduce the redundancy here. I can see that there should be a better solution so I'll try to update the PR soon.

To reduce the redundancy, I'm assuming you had something like this in mind?:

```
/// Returns a string representation of a binary operation.
fn binary_op_to_string(
    ir: &AirIR,
    l_idx: &NodeIndex,
    r_idx: &NodeIndex,
    binary_op: char,
    elem_type: ElemType,
) -> String {
    let lhs = match binary_op {
        '+' | '-' => l_idx.to_string(ir, elem_type),
        '*' => {
            if is_leaf(l_idx, ir) {
                l_idx.to_string(ir, elem_type)
            } else {
                format!("({})", l_idx.to_string(ir, elem_type))
            }
        }
        _ => panic!("Unsupported binary operation."),
    };
    let rhs = match binary_op {
        '+' => r_idx.to_string(ir, elem_type),
        '-' | '*' => {
            if is_leaf(r_idx, ir) {
                r_idx.to_string(ir, elem_type)
            } else {
                format!("({})", r_idx.to_string(ir, elem_type))
            }
        }
        _ => panic!("Unsupported binary operation."),
    };
    format!("{lhs} {binary_op} {rhs}")
}
```

But incorporating the change for multiple multiplications in a row probably makes the previous approach make more sense.

tohrnii: Nvm, I realize you had something else in mind.

tohrnii:
@grjte I've updated the PR. I think this should work correctly.

grjte (approved):
Thanks @tohrnii, this precedence solution looks good. Thanks for updating this!

PR #156, opened 2023-02-20
refactor(parser): refactor grammar to enforce consts are declared at top

tohrnii:


tohrnii:
I think it might be better to keep this check outside of the grammar (in the IR) as we should probably also check other things like only one airdef section is present or only one random_values section is present etc and we should keep the grammar light.

Issue #150, opened 2023-02-17
Potential refactorings post v0.2


bobbinth:
IR building process got pretty complicated and I think we can refactor it to make it more modular and easier to follow. I think the main thing is that there are quite a few functions which take a large number of arguments and keeping track of what's going on is difficult (at least for me). Below are a few suggestions. I'm not sure all of them will make sense as I haven't thought through everything in detail. So, maybe some (or all) of them will end up being discarded.

First, maybe we should separate building of constraint expression graph from the graph itself. Many method on `AlgebraicGraph` take a large number of arguments because they need access to the current context (e.g., `SymbolTable`, `VariableRoots`). Instead, maybe it would make sense to introduce a "builder" struct which would be responsible for maintaining the context and building the graph. Maybe something like this:
```Rust
struct ConstraintBuilder {
    symbol_table: SymbolTable,
    variable_roots: VariableRoots,
    constrained_boundaries: BTreeSet<ConstrainedBoundaries>,
    graph: AlgebraicGraph,
}
```

This struct would have methods like `insert_boundary_stmt()` and `insert_integrity_stmt()`. After all statements are inserted, it would be converted into a `Constraints` struct like so:
```Rust
impl ConstraintBuilder {
    fn into_constraints(self) -> Constraints
}
```
`ConstraintBuilder` would also contain most expression processing functions (e.g., `insertExpr` which is currently in `AlgebraicGraph`). Only the methods which modify the graph directly (e.g., `insert_op()`) would remain in `AlgebraicGraph`.

Second, we seem to be keeping track of some additional constraint info while inserting expressions into the graph. Specifically, we have this `ExprDetail` which is returned from many functions in the `AlgebraicGraph`. I wonder if we can take the same approach here as we did with computing constraint degree. Specifically, instead of keeping track of this additional info while we build a constraint, we could determine by recursively traversing the graph after the constraint is built. This would probably be less efficient, but I don't think we care about performance for this specific part.

Third, it may make sense to make some cosmetic changes to the symbol table. Specifically, we could adopt the nomenclature that symbol table stores symbols, not identifier types. I think this would make things a bit easier to understand as we'd use "symbol" everywhere rather than a mix of symbol, identifier type, element type etc.

grjte:
I agree that a refactor would be good - I've been wanting to revisit this as well, so I'd like to work on this issue.

> First, maybe we should separate building of constraint expression graph from the graph itself. 

I think this makes sense.

[...]

PR #160, opened 2023-02-21
Add support for boundary variables

grjte:
This PR addresses #106. As mentioned in #150, I'd like to do some refactoring, including the symbol table, so I consider this implementation to be a preliminary version

### DIFF
--- a/ir/src/constraints/graph.rs
+++ b/ir/src/constraints/graph.rs
@@ -1,6 +1,6 @@
 use super::{
     BTreeMap, ConstantType, ConstraintDomain, ExprDetails, Expression, Identifier, IdentifierType,
-    IndexedTraceAccess, IntegrityConstraintDegree, MatrixAccess, SemanticError, SymbolTable,
+    IndexedTraceAccess, IntegrityConstraintDegree, MatrixAccess, Scope, SemanticError, SymbolTable,
     TraceSegment, VariableRoots, VariableType, VectorAccess,
 };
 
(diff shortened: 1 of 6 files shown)


ir/src/constraints/graph.rs
```diff
@@ -287,17 +287,44 @@ impl AlgebraicGraph {
...
+                        Scope::BoundaryConstraints => {
+                            // TODO: deal with boundary conflict properly
+                            Ok(ExprDetails::new(
+                                expr.root_idx(),
+                                expr.trace_segment(),
+                                domain,
+                            ))
+                        }
+                        Scope::IntegrityConstraints => Ok(*expr),
+                    }
```
grjte, line 316: This can be addressed during #150, but it actually requires significant refactoring to do it cleanly

ir/src/tests/variables.rs
```diff
@@ -66,29 +162,121 @@ fn err_ic_variable_access_before_declaration() {
...
+    trace_columns:
+        main: [clk]
+    public_inputs:
+        stack_inputs: [16]
+    integrity_constraints:
+        let a = 1
+        enf clk' = clk + 1
+    boundary_constraints:
+        enf clk.first = 0
+        enf clk.last = a";
```
tohrnii, line 195: Is it intentional to reverse the order of integrity and boundary constraints sections here?

tohrnii (approved):
Thank you @grjte, looks great!

ir/src/tests/variables.rs
grjte: Oops, no! Good catch, thanks

grjte: Oh wait, yes - it is. It needs to be reversed there in order to test defining the variable in the `integrity_constraints` and then using it in the `boundary_constraints` (out of scope)

tohrnii: My mistake, I thought that it would work the other way around as well but you probably did it to make the example clearer but it wouldn't work the other way around.

PR #161, opened 2023-02-22
chore: add changelog for v0.1 and v0.2

grjte:
This PR adds a changelog in preparation for the release of v0.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,21 @@
+# Changelog
+
+## 0.2.0 (2023-02-22)
+
+- Added support for named constants (scalars, vectors, and matrices).
+- Added support for intermediate variables in `boundary_constraints` and `integrity_constraints` sections (scalars, vectors, and matrices).
+- Added support for binding identifier names to groups of trace columns (in addition to single columns, which were already supported).
+- Added the `$main` and `$aux` built-ins for accessing columns in the execution trace by index.
+- [BREAKING] Replaced the `$rand` built-in for accessing random values with an explicit `random_values` declaration section, enabling use of a custom identifier for the random values array.
+- Added support for binding identifiers to specific random values or groups of random values.
+- Made significant changes to the IR, including:
+  - Moved the boundary constraints into the algebraic constraint graph.
+  - Made the trace reference more general to support additional trace segments in the future.
+  - Added analysis to differentiate between validity and transition constraints.
+  - Added the `Sub` operation and removed the `Neg` operation.
+- [FIX] Fixed a bug in the Winterfell codegen output for auxiliary transition constraints.
+- Improved the Winterfell codegen by consolidating code generation for boundary and transition constraints and removing redundant parentheses.
+
+## 0.1.0 (2022-11-10)
+
+- Initial release of AirScript, including a minimal but complete implementation of the AirScript compiler for an initial basic set of language features.


grjte:
We can wait to merge this until the docs PR #152 is done and everything else is ready to go.

CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
+# Changelog
+
+## 0.2.0 (2023-02-22)
+
+- Added support for named constants (scalars, vectors, and matrices).
+- Added support for intermediate variables in `boundary_constraints` and `transition_constraints` sections (scalars, vectors, and matrices).
```
tohrnii, line 6: should this be `integrity_constraints` section?

CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
...
+- Added support for binding identifier names to groups of trace columns (in addition to single columns, which were already supported).
+- Added the `$main` and `$aux` built-ins for accessing columns in the execution trace by index.
+- [BREAKING] Replaced the `$rand` built-in for accessing random values with an explicit `random_values` declaration section, enabling use of a custom identifier for the random values array.
+- Added support for binding identifiers to specific random values.
```
tohrnii, line 10: Should we mention groups as well? Maybe something like:
"Added support for binding identifiers to specific random values or groups of random values."
or do you think that is already covered in the above statement?

CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
...
+- Made significant changes to the IR, including:
```
tohrnii, line 11: Should we say "IR structure" or maybe "IR design" here instead?

CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
...
+  - Moved the boundary constraints into the algebraic constraint graph.
+  - Made the trace reference more general to support additional trace segments in the future.
+  - Added analysis to differentiate between validity and transition constraints.
```
tohrnii, line 14: Do you think replacing "analysis" with "ability" would be better?

CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
...
+  - Added the `Sub` operation and removed the `Neg` operation.
+- [FIX] Fixed a bug in the Winterfell codegen output for auxiliary transition constraints.
```
tohrnii, line 16: should we be more specific here?

"Fixed a bug in the Winterfell codegen output for auxiliary transition constraints using main trace columns"

tohrnii (commented):
Thank you @grjte, looks good to me. I just left a few questions inline.

CHANGELOG.md
grjte: Hm, no - what I meant is that we now analyze the constraints to determine the type. I could change it to the following if you think it's better: "Added analysis of integrity constraints to identify the constraint type as validity or transition"

CHANGELOG.md
grjte: I think I'd leave it as it is - we've made changes to both the design and the structure, as well as the interface used for codegen, and "IR" encompasses all of that.

grjte:
Thanks for the review @tohrnii, I've made some of the updates and addressed your other points

CHANGELOG.md
tohrnii: No in that case I think what you had originally is probably better.

CHANGELOG.md
tohrnii: Makes sense.

tohrnii (approved):
Looks good to me. Thanks @grjte.

PR #169, opened 2023-02-23
fix(codegen): convert base field to extension field

tohrnii:


codegen/winterfell/src/air/graph.rs
```diff
@@ -47,19 +47,23 @@ impl Codegen for IndexedTraceAccess {
             }
             _ => panic!("Winterfell doesn't support row offsets greater than 1."),
         };
-        format!("{frame}_{row_offset}")
+        if self.trace_segment() == 0 && self.trace_segment() != trace_segment {
+            format!("E::from({frame}_{row_offset})")
+        } else {
+            format!("{frame}_{row_offset}")
+        }
```
grjte, line 54: for the future: ideally it would be better to use the `elem_type` instead of also passing the `trace_segment` into this function. Maybe we should make an issue to revisit the codegen in the future?

grjte (approved):
Thanks @tohrnii, it looks good!

codegen/winterfell/src/air/graph.rs
tohrnii: I opened #170

PR #172, opened 2023-02-23
Refactoring: Use specialized SemanticError constructors

### LINKED ISSUE #109: Replace errors in IR to call constructor methods
It would be better to refactor the errors in IR to call constructor methods for SemanticError instead of building the error string where the error is thrown. This will improve code readbility.

_Originally proposed by @bobbinth in https://github.com/0xPolygonMiden/air-script/pull/94#discussion_r1040239118_

jjcnn:
Fixes #109.

This PR introduces specialized constructors for SemanticError for the various errors that can occur in the IR. This makes for better code readability, and for more standardized error messages.

I'm unsure whether I've picked appropriate names for the constructors, and I'm also not sure whether the `SemanticError` enum should be extended.

### DIFF
--- a/ir/src/constraints/graph.rs
+++ b/ir/src/constraints/graph.rs
@@ -360,9 +360,7 @@ impl AlgebraicGraph {
                         variable_expr,
                     )
                 } else {
-                    Err(SemanticError::InvalidUsage(format!(
-                        "Identifier {ident} was declared as a {elem_type} which is not a supported type."
-                    )))
+                    Err(SemanticError::unsupported_identifer_type(ident, elem_type))
                 }
             }
             IdentifierType::PeriodicColumn(index, cycle_len) => {
@@ -376,9 +374,7 @@ impl AlgebraicGraph {
                 let node_index = self.insert_op(Operation::TraceElement(trace_access));
                 Ok(ExprDetails::new(node_index, trace_segment, domain))
             }
-            _ => Err(SemanticError::InvalidUsage(format!(
-                "Identifier {ident} was declared as a {elem_type} which is not a supported type."
-            ))),
+            _ => Err(SemanticError::unsupported_identifer_type(ident, elem_type)),
         }
     }
 
@@ -510,9 +506,7 @@ impl AlgebraicGraph {
             IdentifierType::RandomValuesBinding(_, _) => {
                 self.insert_random_value(symbol_table, index, trace_segment, domain)
             }
-            _ => Err(SemanticError::InvalidUsage(format!(
-                "Identifier {name} was declared as a {elem_type} not as a random values"
-            ))),
+            _ => Err(SemanticError::unsupported_identifer_type(name, elem_type)),
         }
     }
 }
--- a/ir/src/constraints/mod.rs
+++ b/ir/src/constraints/mod.rs
@@ -6,7 +6,7 @@ use super::{
 use std::collections::{BTreeMap, BTreeSet};
 
 mod constraint;
-use constraint::ConstrainedBoundary;
+pub use constraint::ConstrainedBoundary;
 pub use constraint::{ConstraintDomain, ConstraintRoot};
 
 mod degree;
@@ -264,9 +264,9 @@ impl Constraints {
                 // add the boundary to the set of constrained boundaries.
                 if !self.constrained_boundaries.insert(constrained_boundary) {
                     // raise an error if the same boundary was previously constrained
-                    return Err(SemanticError::TooManyConstraints(format!(
-                        "A constraint was already defined at {constrained_boundary}",
-                    )));
+                    return Err(SemanticError::boundary_already_constrained(
+                        &constrained_boundary,
+                    ));
                 }
 
                 // add the trace access at the specified boundary to the graph.
@@ -288,7 +288,7 @@ impl Constraints {
                 // trace segment inference defaults to the lowest segment (the main trace) and is
                 // adjusted according to the use of random values and trace columns.
                 if lhs.trace_segment() < rhs.trace_segment() {
-                    return Err(SemanticError::InvalidUsage("Random values cannot be used in boundary constraints defined against prior trace segments".to_string()));
+                    return Err(SemanticError::trace_segment_mismatch(&lhs.trace_segment()));
                 }
 
                 // merge the two sides of the expression into a constraint.
@@ -316,8 +316,8 @@ impl Constraints {
 
         // the constraint should not be against an undeclared trace segment.
         if symbol_table.num_trace_segments() <= trace_segment {
-            return Err(SemanticError::InvalidConstraint(
-                "Constraint against undeclared trace segment".to_string(),
+            return Err(SemanticError::trace_segment_mismatch(
+                &constraint.trace_segment,
             ));
         }
 
--- a/ir/src/error.rs
+++ b/ir/src/error.rs
@@ -1,5 +1,8 @@
-use super::{MatrixAccess, VectorAccess};
-use crate::symbol_table::IdentifierType;
+use super::{
+    constraints::{ConstrainedBoundary, ConstraintDomain},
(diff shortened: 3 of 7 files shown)


ir/src/error.rs
```diff
@@ -74,4 +80,150 @@ impl SemanticError {
...
+
+    pub(super) fn random_value_access_out_of_bounds(index: usize, size: u16) -> Self {
+        SemanticError::IndexOutOfRange(format!(
+            "Random value index {index} is greater than or equal to the total number of random values ({size})."
+        ))
+    }
+
+    // --- DECLARATION ERRORS ---------------------------------------------------------------------
+
+    fn missing_declaration(missing_section: &str) -> Self {
```
grjte, line 127: nit: I would probably call this `missing_section_declaration` for clarity

ir/src/error.rs
```diff
@@ -74,4 +80,150 @@ impl SemanticError {
...
+            "random_values section requires aux_trace_columns section, which is missing"
+                .to_string(),
+        )
+    }
+
+    pub(super) fn illegal_identifer_type(ident_name: &str, ident_type: &IdentifierType) -> Self {
+        SemanticError::InvalidUsage(format!(
+            "Identifier {ident_name} was declared as a {ident_type} which is not a supported type."
+        ))
+    }
```
grjte, line 158: should this be moved down under the "ILLEGAL IDENTIFIER ERRORS" header?

ir/src/error.rs
```diff
@@ -74,4 +80,150 @@ impl SemanticError {
...
+
+    pub(super) fn illegal_trace_column_identifier(
+        ident_name: &str,
+        ident_type: &IdentifierType,
+    ) -> Self {
+        SemanticError::InvalidUsage(format!(
+            "Identifier {ident_name} was declared as a {ident_type} not as a trace column"
+        ))
+    }
```
grjte, line 182: I think this error gets used when we expect a trace column but instead we get some other type, so I'm not sure that it's an illegal identifier error - it's more of a type mismatch error. I think `illegal_trace_column_identifier` is a bit of a misnomer, and maybe the error message should be rephrased as well. Wdyt?

ir/src/error.rs
```diff
@@ -74,4 +80,150 @@ impl SemanticError {
...
+        other: &ConstraintDomain,
+    ) -> Self {
+        SemanticError::InvalidConstraintDomain(format!(
+            "The specified constraint domain {other:?} is not compatible with the base domain {base:?}"
+        ))
+    }
+
+    pub(super) fn constraint_already_defined(boundary: &ConstrainedBoundary) -> Self {
+        SemanticError::TooManyConstraints(format!("A constraint was already defined at {boundary}"))
+    }
```
grjte, line 218: nit: I think `boundary_already_constrained` might be better. The constraint causing the error could actually be a completely new constraint. The problem is that we can only enforce one constraint at any given boundary, and there's already one specified at this boundary.

ir/src/error.rs
```diff
@@ -74,4 +80,150 @@ impl SemanticError {
...
+    // --- INVALID CONSTRAINT ERRORS --------------------------------------------------------------
+
+    pub(super) fn incompatible_constraint(
+        base: &ConstraintDomain,
+        other: &ConstraintDomain,
+    ) -> Self {
+        SemanticError::InvalidConstraintDomain(format!(
+            "The specified constraint domain {other:?} is not compatible with the base domain {base:?}"
+        ))
+    }
```
grjte, line 214: Let's adjust the error message to "The specified constraint domains {base:?} and {other:?} are incompatible" and change the name to `incompatible_constraint_domains`

ir/src/error.rs
```diff
@@ -74,4 +80,150 @@ impl SemanticError {
...
+        SemanticError::InvalidPeriodicColumn(format!(
+            "cycle length must be a power of two, but was {length} for cycle {cycle_name}"
+        ))
+    }
+
+    pub(super) fn cycle_length_too_small(length: usize, cycle_name: &str) -> Self {
+        SemanticError::InvalidPeriodicColumn(format!(
+            "cycle length must be at least {MIN_CYCLE_LENGTH}, but was {length} for cycle {cycle_name}"
+        ))
+    }
```
grjte, line 196: let's prefix these with `periodic_` for (e.g. `periodic_cycle_length_not_power_of_two`)

ir/src/constraints/mod.rs
```diff
@@ -288,7 +288,9 @@ impl Constraints {
                 // trace segment inference defaults to the lowest segment (the main trace) and is
                 // adjusted according to the use of random values and trace columns.
                 if lhs.trace_segment() < rhs.trace_segment() {
-                    return Err(SemanticError::InvalidUsage("Random values cannot be used in boundary constraints defined against prior trace segments".to_string()));
+                    return Err(
+                        SemanticError::random_value_in_constraint_against_prior_trace_segment(),
+                    );
```
grjte, line 293: I think we should simplify/generalize this error to something like "The constraint expression cannot be enforced against trace segment {}." and maybe call the error `trace_segment_mismatch` and pass in the `lhs.trace_segment()`

ir/src/error.rs
```diff
@@ -74,4 +80,150 @@ impl SemanticError {
...
+
+    pub(super) fn constraint_already_defined(boundary: &ConstrainedBoundary) -> Self {
+        SemanticError::TooManyConstraints(format!("A constraint was already defined at {boundary}"))
+    }
+
+    pub(super) fn random_value_in_constraint_against_prior_trace_segment() -> Self {
+        SemanticError::InvalidUsage(
+            "Random values cannot be used in boundary constraints defined against prior trace segments".to_string()
+        )
+    }
```
grjte, line 224: I think we should do something more general, as noted elsewhere

grjte (changes requested):
Thanks @jjcnn! I left a few comments inline, mostly in cases where the original error could have been improved or where you may have been missing context related to AirScript that might have suggested a slightly different name

ir/src/error.rs
jjcnn: Probably - I don't know the relationship between identifiers and their types, so I figured the type was part of the declaration of the identifier.

I'll move it.

ir/src/error.rs
jjcnn: I like `TYPE ERRORS` as a section, and the one where the type of an identifier is illegal can go into that section as well. Does that make sense?

jjcnn:
@grjte : Review comments addressed.

ir/src/error.rs
grjte: Yep, that sounds good to me!

ir/src/constraints/graph.rs
```diff
@@ -360,9 +360,7 @@ impl AlgebraicGraph {
                         variable_expr,
                     )
                 } else {
-                    Err(SemanticError::InvalidUsage(format!(
-                        "Identifier {ident} was declared as a {elem_type} which is not a supported type."
-                    )))
+                    Err(SemanticError::unsupported_identifer_type(ident, elem_type))
```
grjte, line 363: comment for the future: we should revisit this. The problem isn't that the types are unsupported generally. It's that they're not allowed in the place where they're being used. However, this section will be changing significantly during the refactor #150, so I think it's best to merge it as it is.

grjte (changes requested):
Thanks @jjcnn everything looks great to me now! Can you just squash the commits and update the commit message to the semantic style specified in our [contributor doc](https://github.com/0xPolygonMiden/air-script/blob/main/CONTRIBUTING.md)? Then we'll be ready to merge

jjcnn:
> Thanks @jjcnn everything looks great to me now! Can you just squash the commits and update the commit message to the semantic style specified in our [contributor doc](https://github.com/0xPolygonMiden/air-script/blob/main/CONTRIBUTING.md)? Then we'll be ready to merge

Done (I think).

grjte (approved):
Approved! Thank you!

PR #177, opened 2023-02-27
Replace Felt::new with Felt::ZERO or Felt::ONE where appropriate

### LINKED ISSUE #48: Update codegen to use Felt::ZERO, Felt::ONE instead of `new` & `from`
not for this PR, but I think it would be better to use `Felt::ZERO` and `Felt::ONE` when converting from `0` or `1`. Can you open an issue? (Same thing with `E::ZERO` and `E::ONE`)

_Originally posted by @grjte in https://github.com/0xPolygonMiden/air-dsl/pull/42#discussion_r1007581722_

jjcnn:
Fixes #48.

When generating code for `Felt::new(0)` or `Felt::new(1)`, instead use `Felt::ZERO` and `Felt::ONE`, respectively.

Similarly, use `E::0_u64` and `E::`1_u64` instead of `E::PositiveInteger::from{0}_u64` and `E::PositiveInteger::from{1}_u64`, respectively.

The tests pass, but I am unsure whether the generated Rust code actually runs, or whether the test is simply a `diff` against the expected output.


tohrnii (approved):
Thank you @jjcnn, looks good to me.

tohrnii:
@jjcnn Just to let you know for future, we try to use semantic commit messages in this and other miden repos.

jjcnn:
> @jjcnn Just to let you know for future, we try to use semantic commit messages in this and other miden repos.

Could you help me out with a suggestion for alternative commit message? I've tried to follow the guidelines, but I'm quite sure I missed a few things, so an example would be helpful.

tohrnii:
The format we use is `type of commit` (`crate name`): `commit message`
So, for e.g., the commits in this PR could be something like:
```
refactor(codegen): Replace Felt::new with Felt::ZERO or Felt::ONE where appropriate
refactor(codegen): Simplify exp expression when exponent is 0 or 1
```
Honestly, I personally don't have a strong preference but it would be good to be consistent throughout our repo if possible.

jjcnn:
Got it, thanks. I had missed the `type of commit` and `crate name` parts.

PR #179, opened 2023-02-27
IR refactor: create ConstraintBuilder

grjte:
This PR is the first in a series addressing #150. This one simply reorganizes the IR crate but does not yet move the functions that build the graph into the `ConstraintBuilder` as mentioned in that issue. That will be addressed in a subsequent PR. There are a number of `TODOs` throughout the code which will be removed by the end of this refactor

The next PRs will:

- [ ] simplify symbol validation
- [ ] refactor the symbol table to contain scoped symbols
- [ ] move the functions that process expressions from the `AlgebraicGraph` to the `ConstraintBuilder`

PR #181, opened 2023-02-27
IR refactor: simplify symbol table validation of identifier accesses

grjte:
This is the second in a series of PRs to address #150 . This one begins to simplify the symbol table, and especially the handling of references to identifiers. It is built on top of #179 . I haven't updated documentation yet because I wanted to see what people think of the approach first

### DIFF
--- a/ir/src/symbol_table/access_validation.rs
+++ b/ir/src/symbol_table/access_validation.rs
@@ -1,0 +1,94 @@
+use super::{
+    ConstantType, IdentifierType, MatrixAccess, NamedTraceAccess, SemanticError, VariableType,
+    VectorAccess,
+};
+
+/// Checks that the specified access into an identifier is valid and returns an error otherwise.
+/// # Errors:
+/// - Returns an error if the type of the identifier does not allow the access type. For example,
+///   VariableType::Vector does not allow a MatrixAccess.
+/// - Returns an error if any indices specified for the access are out of bounds fo the specified
+///   identifier.
+pub(super) trait ValidateIdentifierAccess {
+    fn validate(&self, symbol_type: &IdentifierType) -> Result<(), SemanticError>;
+}
+
+impl ValidateIdentifierAccess for NamedTraceAccess {
+    fn validate(&self, symbol_type: &IdentifierType) -> Result<(), SemanticError> {
+        match symbol_type {
+            IdentifierType::TraceColumns(columns) => {
+                if self.idx() >= columns.size() {
+                    return Err(SemanticError::named_trace_column_access_out_of_bounds(
+                        self,
+                        columns.size(),
+                    ));
+                }
+            }
+            _ => {
+                return Err(SemanticError::not_a_trace_column_identifier(
+                    self.name(),
+                    symbol_type,
+                ))
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// Checks that the specified vector access is valid and returns an error otherwise.
+impl ValidateIdentifierAccess for VectorAccess {
+    /// TODO: docs (errors)
+    fn validate(&self, symbol_type: &IdentifierType) -> Result<(), SemanticError> {
+        let vector_len = match symbol_type {
+            IdentifierType::Constant(ConstantType::Vector(vector)) => vector.len(),
+            IdentifierType::PublicInput(size) => *size,
+            IdentifierType::RandomValuesBinding(_, size) => *size,
+            IdentifierType::TraceColumns(trace_columns) => trace_columns.size(),
+            IdentifierType::Variable(_, variable) => {
+                match variable.value() {
+                    // TODO: scalar can be ok; check this symbol in the future
+                    VariableType::Scalar(_) => return Ok(()),
+                    VariableType::Vector(vector) => vector.len(),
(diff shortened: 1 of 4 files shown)


bobbinth (approved):
Thank you! I like this structure - it looks much cleaner.

One potential suggestion I have is maybe to rename `IdentifierType` into `Symbol`. It seems to me that data contained in this enum is more than just type information. This also will clear up some confusing naming patters. For example, in:
```Rust
let symbol_type = symbol_table.get_type(symbol_name)?;
```
`get_type()` returns `IdentifierType` but we name it `symbol_type`. I'd probably have it as:
```Rust
let symbol = symbol_table.get_symbol(symbol_name)?;
```
And then `get_symbol()` would return `Symbol` type.

This could also make naming easier throughout. For example, `ValidateIdentifierAccess` trait could be named `SymbolAccessValidator`.

It is entirely possible that this naming makes sense just to me - so, I don't have a strong preference here.

grjte:
> Thank you! I like this structure - it looks much cleaner.
> 
> One potential suggestion I have is maybe to rename `IdentifierType` into `Symbol`. It seems to me that data contained in this enum is more than just type information. This also will clear up some confusing naming patters. For example, in:
> 
> ```rust
> let symbol_type = symbol_table.get_type(symbol_name)?;
> ```
> 
> `get_type()` returns `IdentifierType` but we name it `symbol_type`. I'd probably have it as:
> 
> ```rust
> let symbol = symbol_table.get_symbol(symbol_name)?;
> ```
> 
> And then `get_symbol()` would return `Symbol` type.
> 
> This could also make naming easier throughout. For example, `ValidateIdentifierAccess` trait could be named `SymbolAccessValidator`.
> 
> It is entirely possible that this naming makes sense just to me - so, I don't have a strong preference here.

Yes, this is in a future PR. Breaking this stuff up into review-sized chunks has been a challenge in itself

ir/src/symbol_table/mod.rs
```diff
@@ -250,34 +253,20 @@ impl SymbolTable {
...
-            _ => Err(SemanticError::InvalidUsage(format!(
-                "Identifier {} was declared as a {} not as a trace column",
-                trace_access.name(),
-                elem_type
-            ))),
-        }
+        let symbol_type = self.get_type(trace_access.name())?;
+        trace_access.validate(symbol_type)?;
+
+        let IdentifierType::TraceColumns(columns) = symbol_type else { panic!("symbol table validation of named trace access failed.") };
```
tohrnii, line 264: nit: Any reason to panic here instead of returning an error?

tohrnii (approved):
Thank you @grjte, this is pretty neat. I just left a couple of non-blocking nits inline.

ir/src/symbol_table/access_validation.rs
```diff
@@ -0,0 +1,96 @@
...
+    VectorAccess,
+};
+
+/// Checks that the specified access into an identifier is valid and returns an error otherwise.
+/// # Errors:
+/// - Returns an error if the type of the identifier does not allow the access type. For example,
+///   VariableType::Vector does not allow a MatrixAccess.
+/// - Returns an error if any indices specified for the access are out of bounds fo the specified
+///   identifier.
+pub(super) trait ValidateIdentifierAccess {
```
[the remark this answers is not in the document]
grjte: I did that first, but then I was thinking about updating the validation of the trace and random values accesses in as well, and those would be handled slightly differently. In essence, they would be `ValidateAccess` and these would be `ValidateIdentifierAccess`. I decided to hold off on that change, but kept the naming here for clarity

ir/src/symbol_table/mod.rs
grjte: Yes - the reason I did this was that I want to move all the errors to the validation. This error is already handled in the `validate` call above, so the else case here should never happen. If it does happen, then something has gone badly wrong with the validation. I could put this into an `unreachable` instead for clarity if you think that's better

ir/src/symbol_table/mod.rs
```diff
@@ -81,6 +84,14 @@ pub struct SymbolTable {
 }
 
 impl SymbolTable {
+    /// Consumes this symbol table and returns the information required for declaring constants,
+    /// public inputs, periodic columns and columns amount for the AIR.
+    pub(super) fn into_declarations(self) -> Declarations {
+        self.declarations
+    }
+
+    // --- SYMBOL INSERTIONS ----------------------------------------------------------------------
```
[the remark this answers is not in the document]
grjte: I think I was looking at different types of mutators and therefore trying to be more specific for better organization and clarity. However, it looks like that's no longer applicable. I can change this to `MUTATORS` and if I update in the future do `MUTATORS: SYMBOL INSERTIONS` instead

grjte:
I'm planning to wait to merge this until we've merged #179

ir/src/symbol_table/mod.rs
tohrnii: Thanks, I thought that could have been the reasoning. I don't have a strong opinion but to me `unreachable` does seem clearer for these cases but I'll defer to you.

Overcastan (approved):

Issue #178, opened 2023-02-27
write MidenVM constraints for the range checker (excludes multiset check constraints)


tohrnii:
The constraints for the range checker are described [here](https://0xpolygonmiden.github.io/miden-vm/design/range.html#execution-trace-constraints) and implemented in Rust [here](https://github.com/0xPolygonMiden/miden-vm/blob/main/air/src/range.rs).

The following constraints related to multiset checks should be excluded from this issue:

- 8-bit range checks table
- range checker communication bus

grjte:
closed by #184

PR #184, opened 2023-02-28
Add range checker constraints

tohrnii:
Partly addressing #178. This PR adds range checker constraints against the main trace. Auxiliary constraints will be added in a later PR.

### DIFF
--- a/constraints/miden-vm/range-checker.air
+++ b/constraints/miden-vm/range-checker.air
@@ -1,0 +1,41 @@
+mod RangeCheckerAir
+
+### Helper functions ##############################################################################
+
+# Returns binary negation of the value.
+fn binary_not(e: scalar) -> scalar:
+    return 1 - e
+
+### Helper evaluators #############################################################################
+
+# Returns the change in value between current and next rows.
+fn delta(e: scalar) -> scalar:
+    return e' - e
+
+# Enforces that column must be binary.
+ev is_binary(main: [v]):
+    enf v^2 = v
+
+# Enforces correct transition from 8-bit to 16-bit section of the table.
+ev transition_8_to_16_bit_validity(main: [t, v]):
+    # Ensure that values in column t can flip from 0 to 1 only once
+    enf t * binary_not(t') = 0
+
+    # Ensure that when column t flips, column v must equal 255
+    enf v = 255 when t' & !t
+
+    # Ensure that when column t flips, v' must be reset to 0
+    enf v' = 0 when t' & !t
+
+### Range checker Air Constraints #################################################################
+
+ev range_checker(main: [t, s0, s1, v]):
+    # Check selector flags are binary
+    let selectors = [t, s0, s1]
+    enf check_binary([s]) for s in selectors
+
+    # Constrain the row transitions in the 8-bit section of the table
+    enf change(v) * (change(v) - 1) = 0 when !t0'
+
+    # Constrain the transition from 8-bit to 16-bit section of the table
+    enf transition_8_to_16_bit_validity([t, v])


constraints/miden-vm/range-checker.air
```diff
@@ -0,0 +1,32 @@
...
+    # t^2 - t = 0
+    # s0^2 - s0 = 0
+    # s1^2 - s1 = 0
+    let selectors = [t, s0, s1]
+    enf check_binary([s]) for s in selectors
+
+    # Constrain the row transitions in the 8-bit section of the table
+    enf change(v) * (change(v) - 1) = 0 when !t0'
+
+    # Constrain the transition from 8-bit to 16-bit section of the table
```
Overcastan, line 23: If I got it right, this is just a "section separator" comment that is telling us what constraints are located below. For me it's a little bit confusing -- at first I thought that constraint is just missing. Maybe we should somehow highlight such  comments? For example, using two `#`s instead of one, and, maybe, using in before and after comment? Something like `## Comment ##`. Wdyt?

Overcastan (changes requested):
Looks good @tohrnii! Just two small nits.

constraints/miden-vm/range-checker.air
tohrnii: I moved the constraints to another evaluator function instead. I'm wondering if as a general rule maybe we can try to move a group of constraints into another evaluator (which is also what you've done in the other chiplet constraints PRs)?

Overcastan: Yes, I think it is a good idea. It looks better in my opinion.

Overcastan (approved):
Looks good to me, thank you @tohrnii! Just one thought: this is fairly small file, but maybe for the consistency we should use delimiters for the sections, like grjte did in bitwise chiplet. Wdyt?

constraints/miden-vm/range-checker.air
grjte: > I'm wondering if as a general rule maybe we can try to move a group of constraints into another evaluator (which is also what you've done in the other chiplet constraints PRs)?

Yes, I prefer this as a general rule too

constraints/miden-vm/range-checker.air
```diff
@@ -0,0 +1,31 @@
+mod RangeCheckerAir
+
+fn binary_not(e: scalar) -> scalar:
+    return 1 - e
+
+fn change(e: scalar) -> scalar:
+    return e' - e
```
grjte, line 7: nit: In the Memory chiplet constraints, we used `delta`. I'm not sure that it's necessarily the best function name, but I'd prefer to try to be consistent in our naming across constraint modules

constraints/miden-vm/range-checker.air
```diff
@@ -0,0 +1,31 @@
...
+    # Ensure that when column t flips, column v must equal 255
+    enf v = 255 when t' & !t
+
+    # Ensure that when column t flips, v' must be reset to 0
+    enf v' = 0 when t' & !t
+
+ev range_checker(main: [t, s0, s1, v]):
+    # Check selector flags are binary
+    let selectors = [t, s0, s1]
+    enf check_binary([s]) for s in selectors
```
grjte, line 25: can we not just do the following?

```
enf check_binary([s]) for s in [t, s0, s1]
```

grjte (approved):
This looks correct, and it's readable. I left a few nits inline and a question. Also, I would prefer to add section headers for functions, evaluators, and constraints the way we've started doing in the other modules (e.g. bitwise/memory/hash chiplets)

constraints/miden-vm/range-checker.air
tohrnii: Currently we don't support inlined vectors as iterables in list or constraint comprehension but I've opened #230.

Issue #124, opened 2023-01-26
Add support for evaluator functions


tohrnii:
We can introduce "evaluator" functions which would take the description of main and auxiliary traces as parameters which would allow us to do something like:

```
ev foo(main: [a, b, ..2], aux: [c, d]):
  enf a' = a + b
```

One advantage of this approach is that we can name evaluator functions and then invoke them from other places. For example:

```
ev foo(main: [a, b, ..2], aux: [c, d]):
  enf bar(main[0..2])
  // some other constraints

ev bar(main: [a, b]):
  enf a' = a + b
```

Rescue Prime constraints in this format could look something like this:

```
rescue.air
--------------------
cycl ark1 = [
  /* 12 periodic column definitions */
]
cycl ark2 = [
  /* 12 periodic column definitions */
]

const mds = [
  /* 12x12 matrix */
]
const inv_mds = [
  /* 12x12 matrix */
]

#[prover("mds.rs")]
fn apply_mds(state: vector[12]) -> vector[12]:
  [
    sum([m * s for (m, s) in (mds_row, state)])
    for mds_row in mds
  ]

fn apply_inv_mds(state: vector[12]) -> vector[12]:
  [
    sum([m * s for (m, s) in (inv_mds_row, state)])
    for inv_mds_row in inv_mds
  ]

ev rescue_prime(main: [state[12]]):
  let s1 = [x^7 for x in state]
  let s1 = apply_mds(s1)
  let s1 = [x + k for (x, k) in (s1, ark1)]

  let s2 = [x' + k for (x, k) in (state, ark2)]
  let s2 = apply_inv_mds(s2)
  let s2 = [x^7 for x in s2]

  enf x = y for (x, y) in (s1, s2)
```

And then we could invoke them from another constraint file like so:

```
hasher.air
--------------------
use rescue::rescue_prime

ev hasher(main: [s[3], r, h[12], idx]):
  enf rescue_prime(h)
```

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-vm/discussions/254#discussioncomment-3302565_

bobbinth:
It may make sense to separate this into two issues:
1. Evaluator functions in a single file.
2. Invoking evaluator functions from imported files.

grjte:
closed by #227  and #260

Issue #122, opened 2023-01-26
Allow constraints to be written in a modular way


tohrnii:
We should be able to write constraints in a modular way. For example, one module for memory constraints, another module for decoder constraints etc.

We use the `use` keywords to specify external modules and then can invoke this modules using the `enf` keywords.

For example, let's say we have a module bar defined as follows:

```
trace_columns:
  main: [a, b, c]
  aux: [d]

enf a' = b + c
enf d' = d * ($rand[0] + a)
```

Let's also say we have a module `foo` from which we'd like to invoke `bar`. We could do this as follows:

```
use: bar

trace_columns:
  main: [a, b, c, d]
  aux: [e, f]

enf bar(main[0..2], aux[0..1])
```

Since the expected number of main and auxiliary columns for `foo` and `bar` are different, we need to specify which columns from `foo` are to be passed to `bar`. We do this using range notation (i.e., `main[0..2]`).

If the number of columns were the same, we could have done something like:

```
enf bar(main, aux)
```

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/miden-vm/discussions/254#discussioncomment-3284930_

grjte:
Broadly, there are 2 approaches that could be taken towards modularity:

1. Each module can function as its own AIR, which means it contains all required declarations and constraints and can stand on its own and be compiled by AirScript as a valid AIR.
2. Sub-modules contain definitions of constants, functions, evaluators, but do not contain any other declarations. In this case, a top-level file is needed to hold all of the declarations for the AIR describing the shape of the trace, etc. That top level file would import all sub-modules and use the defined functions and evaluators in the constraint definitions.

## Now

Option 1 is what is implied by the description in this issue. It's also what we've done so far when writing constraints, for example in the `bitwise` chiplet constraints example.

However, option 2 will make it _much_ easier to support modularity and complete this issue. If we go with option 2 and restrict sub-modules so that they do not contain declarations, then to complete this issue we can just concatenate all sub-modules together at the end of the parent module (with the trace declarations) and then compile that file. Nothing else would need to be done.

Given the short timeline right now, I think we should move forward with this second approach, and @bobbinth has said that this is what he had in mind for initial support of modularity.

## Future

[...]

PR #185, opened 2023-02-28
Add support for evaluator functions to parser

tohrnii:
Partly addressing #124

### DIFF
--- a/parser/src/ast/evaluator_function.rs
+++ b/parser/src/ast/evaluator_function.rs
@@ -1,7 +1,7 @@
 use super::{Identifier, IntegrityStmt, TraceCols};
 
-/// Evaluator functions take description of the main and auxiliary traces as input, and returns
-/// integrity constraints that are enforced on those trace columns.
+/// Evaluator functions take descriptions of the main and auxiliary traces as input, and enforce
+/// integrity constraints on those trace columns.
 #[derive(Debug, Eq, PartialEq)]
 pub struct EvaluatorFunction {
     name: Identifier,
@@ -45,6 +45,17 @@ impl EvaluatorFunction {
     pub fn integrity_stmts(&self) -> &[IntegrityStmt] {
         &self.integrity_stmts
     }
+
+    /// Returns the name, main trace columns, auxiliary trace columns, and integrity statements
+    /// of the evaluator function.
+    pub fn into_parts(self) -> (String, Vec<TraceCols>, Vec<TraceCols>, Vec<IntegrityStmt>) {
+        (
+            self.name.into_name(),
+            self.main_trace,
+            self.aux_trace,
+            self.integrity_stmts,
+        )
+    }
 }
 
 /// Evaluator function call is used to invoke an evaluator function. It takes a list of trace
@@ -69,5 +80,10 @@ impl EvaluatorFunctionCall {
     /// Returns the arguments of the evaluator function.
     pub fn args(&self) -> &Vec<Vec<TraceCols>> {
         &self.args
+    }
+
+    /// Returns the name and arguments of the evaluator function.
+    pub fn into_parts(self) -> (String, Vec<Vec<TraceCols>>) {
+        (self.name.into_name(), self.args)
     }
 }
--- a/parser/src/ast/evaluator_function.rs
+++ b/parser/src/ast/evaluator_function.rs
@@ -1,0 +1,73 @@
+use super::{Identifier, IntegrityStmt, TraceCols};
+
+/// Evaluator functions take description of the main and auxiliary traces as input, and returns
+/// integrity constraints that are enforced on those trace columns.
+#[derive(Debug, Eq, PartialEq)]
+pub struct EvaluatorFunction {
+    name: Identifier,
+    main_trace: Vec<TraceCols>,
+    aux_trace: Vec<TraceCols>,
+    integrity_stmts: Vec<IntegrityStmt>,
+}
+
+impl EvaluatorFunction {
+    /// Creates a new function.
+    pub fn new(
+        name: Identifier,
+        main_trace: Vec<TraceCols>,
+        aux_trace: Vec<TraceCols>,
+        integrity_stmts: Vec<IntegrityStmt>,
+    ) -> Self {
+        Self {
+            name,
+            main_trace,
+            aux_trace,
+            integrity_stmts,
+        }
+    }
+
+    /// Returns the name of the evaluator function.
+    pub fn name(&self) -> &str {
+        self.name.name()
+    }
+
+    /// Returns the main trace of the evaluator function.
+    pub fn main_trace(&self) -> &Vec<TraceCols> {
+        &self.main_trace
+    }
+
+    /// Returns the auxiliary trace of the evaluator function.
+    pub fn aux_trace(&self) -> &Vec<TraceCols> {
+        &self.aux_trace
+    }
+
+    /// Returns the integrity statements of the evaluator function.
+    pub fn integrity_stmts(&self) -> &[IntegrityStmt] {
+        &self.integrity_stmts
+    }
+}
+
+/// Evaluator function call is used to invoke an evaluator function. It takes a list of trace
+/// columns.
+#[derive(Debug, Eq, PartialEq)]
+pub struct EvaluatorFunctionCall {
+    name: Identifier,
+    args: Vec<Vec<TraceCols>>,
+}
+
+impl EvaluatorFunctionCall {
+    /// Creates a new evaluator function call.
+    pub fn new(name: Identifier, args: Vec<Vec<TraceCols>>) -> Self {
+        Self { name, args }
+    }
+
+    /// Returns the name of the evaluator function.
+    pub fn name(&self) -> &str {
+        self.name.name()
+    }
+
+    /// Returns the arguments of the evaluator function.
+    pub fn args(&self) -> &Vec<Vec<TraceCols>> {
+        &self.args
+    }
+}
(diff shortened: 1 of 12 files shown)


parser/src/ast/evaluator_function.rs
```diff
@@ -0,0 +1,73 @@
+use super::{Identifier, IntegrityStmt, TraceCols};
+
+/// Evaluator functions take description of the main and auxiliary traces as input, and returns
+/// integrity constraints that are enforced on those trace columns.
```
grjte, line 4: As noted in #122 and #213, we may need to allow passing periodic columns and random values.

Also, I would rephrase the end of this from "and returns..." to "and enforces integrity constraints on those trace columns."

parser/src/ast/evaluator_function.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// Returns the auxiliary trace of the evaluator function.
+    pub fn aux_trace(&self) -> &Vec<TraceCols> {
+        &self.aux_trace
+    }
+
+    /// Returns the integrity statements of the evaluator function.
+    pub fn integrity_stmts(&self) -> &[IntegrityStmt] {
+        &self.integrity_stmts
+    }
```
grjte, line 47: I think we'll want to add a new `into_parts` method after this so that we can consume each `EvaluatorFunction` and add it to the symbol table when consuming the `AST` during the IR step. However, if you prefer to add this during the PR for the IR changes that's ok too

parser/src/ast/evaluator_function.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/// Evaluator function call is used to invoke an evaluator function. It takes a list of trace
+/// columns.
+#[derive(Debug, Eq, PartialEq)]
+pub struct EvaluatorFunctionCall {
+    name: Identifier,
+    args: Vec<Vec<TraceCols>>,
+}
```
grjte, line 56: similar comment here - I think we may need to revisit this, if we're allowing args to be identifiers as well (e.g. for periodic columns or random values)

parser/src/ast/evaluator_function.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// Returns the name of the evaluator function.
+    pub fn name(&self) -> &str {
+        self.name.name()
+    }
+
+    /// Returns the arguments of the evaluator function.
+    pub fn args(&self) -> &Vec<Vec<TraceCols>> {
+        &self.args
+    }
```
grjte, line 72: similar comment here about adding `into_parts`

parser/src/ast/mod.rs
```diff
@@ -36,6 +39,8 @@ pub struct Source(pub Vec<SourceSection>);
 ///   elements specified. The array length is expected to be a power of 2, but this is not checked
 ///   during parsing.
 /// - RandomValues: Random Values represent the randomness sent by the Verifier.
+/// - EvaluatorFunction: Evaluator Functions take description of the main and auxiliary traces as
+///   input, and returns integrity constraints that are enforced on those trace columns.
```
grjte, line 43: 1. nit: there's a typos "take description" -> "take descriptions"
2. I would adjust the end of it to "and enforce integrity constraints on those trace columns."
3. This may need to change if we allow periodic columns or random values as inputs

parser/src/parser/tests/integrity_constraints.rs
```diff
@@ -232,7 +232,8 @@ fn err_missing_integrity_constraint() {
         let b = [a, 2 * a]
         let c = [[a - 1, a^2], [b[0], b[1]]]";
     let error = Error::ParseError(ParseError::MissingIntegrityConstraint(
-        "Declaration of at least one integrity constraint is required".to_string(),
+        "Declaration of at least one integrity constraint or evaluator function call is required"
+            .to_string(),
```
grjte, line 236: I probably wouldn't change this. In my mind, an evaluator function call is just a special type of constraint. It's still prefixed by the `enf` keyword that we use for constraints, so I don't think the user should need to think about it as a separate thing.

parser/src/parser/grammar.lalrpop
```diff
@@ -155,6 +155,67 @@ RandElem: RandBinding = {
     <name: Identifier> <size: Size> => RandBinding::new(name, size)
 }
 
+// EVALUATOR FUNCTIONS
+// ================================================================================================
+
+EvaluatorFunction: EvaluatorFunction = {
+    "ev" <evaluator_fn_name: Identifier>
+        "(" <main_cols: MainCols?> <cols_separator: (",")?> <aux_cols: AuxCols?> ")" ":"
```
grjte, line 163: Could we simplify this to the following? Then we could remove the error checking for the `cols_separator` to simplify the error handling below as well.

```
"(" <main_cols: MainCols?> "," <aux_cols: AuxCols?> ")" ":"
```

parser/src/parser/grammar.lalrpop
```diff
@@ -240,14 +301,14 @@ IntegrityConstraints: Vec<IntegrityStmt> = {
...
-            IntegrityStmt::Constraint(_) => true,
+            IntegrityStmt::Constraint(_) | IntegrityStmt::EvaluatorFunctionCall(_) => true,
             _ => false,
         });
         if !integrity_constraints_exist {
             return Err(ParseError::User {
                 error: Error::ParseError(
                     MissingIntegrityConstraint(
-                        "Declaration of at least one integrity constraint is required".to_string()
+                        "Declaration of at least one integrity constraint or evaluator function call is required".to_string()
```
grjte, line 311: This looks like the same error message as above. It would be good to extract this so it's not duplicated. This could be improved as part of #188 though

grjte (changes requested):
@tohrnii I think this looks great. I left a few minor comments inline and also flagged a few places that might be affected by the proposed modularity changes mentioned in #122 and #213.

However, I think those changes (if we decide to make them) could be addressed in a subsequent PR, so we could still merge this one and implement the IR changes for this initial evaluator function support first (once the other minor comments are addressed)

parser/src/parser/grammar.lalrpop
tohrnii: Hanging comma seems a little weird to me personally. (e.g. `(main: [a, b, c], )` or `(, aux: [a, b, c])`. But if you think it's worth it to keep the grammar cleaner, I can change it.

parser/src/parser/tests/integrity_constraints.rs
tohrnii: Makes sense. Thanks.

parser/src/ast/evaluator_function.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Returns the integrity statements of the evaluator function.
+    pub fn integrity_stmts(&self) -> &[IntegrityStmt] {
+        &self.integrity_stmts
+    }
+
+    /// Returns the name, main trace columns, auxiliary trace columns, and integrity statements
+    /// of the evaluator function.
+    pub fn into_parts(self) -> (String, Vec<TraceCols>, Vec<TraceCols>, Vec<IntegrityStmt>) {
+        (
+            self.name.into_name(),
```
grjte, line 53: I would return the Identifier here for consistency, since that's what's been done elsewhere

parser/src/ast/evaluator_function.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    /// Returns the arguments of the evaluator function.
+    pub fn args(&self) -> &Vec<Vec<TraceCols>> {
+        &self.args
+    }
+
+    /// Returns the name and arguments of the evaluator function.
+    pub fn into_parts(self) -> (String, Vec<Vec<TraceCols>>) {
+        (self.name.into_name(), self.args)
+    }
```
grjte, line 88: I would return the `Identifier` here instead of a `String` for consistency with other places

grjte (changes requested):
@tohrnii thanks! I left 2 minor comments and a suggestion for simplifying the grammar

parser/src/parser/grammar.lalrpop
grjte: Another benefit of adjusting in this way is that you can use `IntegrityStmts` for the `integrity_constraints` parsing so that the error handling isn't duplicated

parser/src/ast/evaluator_function.rs
tohrnii: I think `String` is returned elsewhere as well?

parser/src/parser/grammar.lalrpop
tohrnii: Thank you! This is much cleaner. Fixed.

parser/src/ast/evaluator_function.rs
grjte: Ah yes, it looks like I misremembered this and you're completely right, sorry. I went through so many iterations of things over the course of the refactor that it seems to just be a weird jumble in my head at this point

parser/src/parser/tests/evaluator_functions.rs
```diff
@@ -0,0 +1,162 @@
...
+            Identifier("advance_clock".to_string()),
+            vec![vec![TraceCols::new(Identifier("clk".to_string()), 1)]],
+        )),
+    ])]);
+
+    build_parse_test!(source).expect_ast(expected);
+}
+
+#[test]
+fn ev_fun_call_inside_ev_fn() {
```
grjte, line 96: nit: I would probably do `ev_fn_call_inside_ev_fn` for consistency with our other names

grjte (approved):
Looks good, thanks @tohrnii! I left one optional nit inline

PR #190, opened 2023-03-02
Refactor the IR

grjte:
This is a large WIP refactor of the IR, addressing #150.

The primary goals are:
- simplify & clarify the code
- reduce duplication
- simplify error handling

There is still a lot of cleanup required, so this PR is not intended to be merged directly - **I'm just hoping to get feedback on the new overall structure,** so I recommend just checking out the branch and looking at the current status locally. After that I'll make adjustments, then start opening PRs sequentially for detailed review. 

Notes: 
1. I ended up temporarily removing the `VariableRoots` for the moment. Ultimately, this map is an optimization, and the handling of variable expressions was one of the biggest sources of complexity. I hope to think of a good way to add this back before the end of this refactor, but there were a few issues related to it that kept coming up.
2. I'd like to make the access validation of symbols nicer in the symbol table (`Symbol::validate_access`), but it's a more localized change so it can be done after the big changes are made
3. Similar to (2), I think list comprehension and list folding could be condensed a bit now, but this is more localized and can be done separately.


bobbinth (commented):
Overall look great! Thank you! I do have a few assorted comments/questions. Not all of them are related to this PR, and some of them might be simply because I didn't look at things deeply enough.

1. I found method naming for `BoundaryConstraint::access()` a bit confusing. Took me a bit to understand that it actually refers to the column against which the constraint is placed. Would `BoundaryConstraint::column()` be a better name?
2. I probably would move `ConstrainedBoundary` struct from `constraints` module to `constraint_builder` module as I think it is used only during constraint building process.
3. This is probably for a different PR, but I still think we can combine `validity_constraints` and `transition_constraints` into a single vector in the `Constraints` struct. It feels like splitting them out makes some assumptions about how backends might want to view the IR - but I think it might make sense to mimic the source code more closely.
4. Putting list folding and list comprehension logic into the `SymbolTable` felt a bit strange to me. I would have expected it to be in the `ConstraintBuilder`. And it doesn't seem like we'd complicate things too much by putting it there since we can access the symbol table via `self.symbol_table` from the `ConstraintBuilder`.
5. It also felt like handling of scopes within `SymbolTable` is potentially more complicated than it needs to be. For example, since we process integrity and boundary constraints separately, we could first process integrity constraints, then remove all symbols related to them from the symbol table, and then process boundary constraints. This way, symbol table never needs to keep track of symbols which are not relevant for the section being processed. If we do this, we should be able to change type of `SymbolTable::symbols` to `BTreeMap<String, Symbol>`.
6. With the above, we should be able to convert `Symbol` struct into an enum (since we no longer need to track scope for it). Basically, it could be the same as `SymbolType` enum but with a string added to each variant to specify the name.
7. The pattern described in point 5 above could work for the `ConstraintBuilder` struct as well. For example, it seems like one of the reason we pass `domain` around through various methods of the `ConstraintBuilder` is to determine whether we are processing a boundary or an integrity constraint. Why not track this info as member variable of `ConstraintBuilder`? This way, after we are done processing boundary constraints, we would set this variable to indicate that we are now processing integrity constraints (e.g., we could have `start_boundary_constrants()` and `start_integrity_constraints()` methods on the `ConstraintBuilder`.
8. It seems like another reason we pass `domain` around in the `ConstraintBuilder` is to determine the domain of the entire constraint. If so, I still think it might be better to separate building of constraints from gathering information about them. For example, I wonder if could insert all constraints first, and then traverse the graph for each constraint to build up all relevant details (degree, domain, maybe something else).


With the last two comments, I wonder if we can simplify the `insert_expr` method of `ConstraintBuilder` to look something like this:
```Rust
pub(super) fn insert_expr(
    &mut self,
    expr: &Expression,
```
This would also simplify all downstream methods accordingly.

grjte:
> Overall look great! Thank you! I do have a few assorted comments/questions. Not all of them are related to this PR, and some of them might be simply because I didn't look at things deeply enough.

Thanks for your feedback! I'll leave my thoughts below.

>     1. I found method naming for `BoundaryConstraint::access()` a bit confusing. Took me a bit to understand that it actually refers to the column against which the constraint is placed. Would `BoundaryConstraint::column()` be a better name?

This is fine with me. It was a bit of a placeholder.

>     2. I probably would move `ConstrainedBoundary` struct from `constraints` module to `constraint_builder` module as I think it is used only during constraint building process.

Yes, agreed.


>     3. This is probably for a different PR, but I still think we can combine `validity_constraints` and `transition_constraints` into a single vector in the `Constraints` struct. It feels like splitting them out makes some assumptions about how backends might want to view the IR - but I think it might make sense to mimic the source code more closely.

I like it as it is, but it seems like you feel more strongly about this than I do, and it's a reasonable argument, so we can change it. I'll leave it for a different PR after the rest of the refactoring is done, since it's a small localised change

>     4. Putting list folding and list comprehension logic into the `SymbolTable` felt a bit strange to me. I would have expected it to be in the `ConstraintBuilder`. And it doesn't seem like we'd complicate things too much by putting it there since we can access the symbol table via `self.symbol_table` from the `ConstraintBuilder`.

The reason I originally move the list comprehension into the symbol table was that it is only used when adding variables of type ListComprehension to the symbol table as symbols and not when processing constraints. However, with the new pattern of getting values from the symbol table and unrolling expressions in the ConstraintBuilder, the list folding definitely belongs in the ConstraintBuilder, and the list comprehension feels like that too, so I'm happy to make this change.

>     5. It also felt like handling of scopes within `SymbolTable` is potentially more complicated than it needs to be. For example, since we process integrity and boundary constraints separately, we could first process integrity constraints, then remove all symbols related to them from the symbol table, and then process boundary constraints. This way, symbol table never needs to keep track of symbols which are not relevant for the section being processed. If we do this, we should be able to change type of `SymbolTable::symbols` to `BTreeMap<String, Symbol>`.
> 
>     6. With the above, we should be able to convert `Symbol` struct into an enum (since we no longer need to track scope for it). Basically, it could be the same as `SymbolType` enum but with a string added to each variant to specify the name.
> 
>     7. The pattern described in point 5 above could work for the `ConstraintBuilder` struct as well. For example, it seems like one of the reason we pass `domain` around through various methods of the `ConstraintBuilder` is to determine whether we are processing a boundary or an integrity constraint. Why not track this info as member variable of `ConstraintBuilder`? This way, after we are done processing boundary constraints, we would set this variable to indicate that we are now processing integrity constraints (e.g., we could have `start_boundary_constrants()` and `start_integrity_constraints()` methods on the `ConstraintBuilder`.
> 
>     8. It seems like another reason we pass `domain` around in the `ConstraintBuilder` is to determine the domain of the entire constraint. If so, I still think it might be better to separate building of constraints from gathering information about them. For example, I wonder if could insert all constraints first, and then traverse the graph for each constraint to build up all relevant details (degree, domain, maybe something else).

I agree with some of the core premises here - I would prefer to simplify the `SymbolTable::symbols` map to `BTreeMap<String, Symbol>`, and had thought about that a bit. I also would prefer to separate the domain inference from the `ConstraintBuilder::insert_expr` method. I think the approaches you've suggested are interesting, though I'm not sure that we won't hit some other subtle complications. I'll try all of them and see

grjte:
I've made updates addressing the major changes you suggested, focusing on the things that affect the structure significantly.

>     1. I found method naming for `BoundaryConstraint::access()` a bit confusing. Took me a bit to understand that it actually refers to the column against which the constraint is placed. Would `BoundaryConstraint::column()` be a better name?

This change would be in the parser, and I'd rather not make it now as part of this PR. Additionally, it's not exactly a column that's being returned. It's a `NamedTraceAccess`, which consists of an identifier for a trace column binding, a index within that binding, and a row offset. We check in the IR that the row offset is zero. We can revisit this and see if we want to change this type to something else, but in the meantime I think using the name `column` would be misleading (though I'm sure we could still do better than `access`).

>     6. With the above, we should be able to convert `Symbol` struct into an enum (since we no longer need to track scope for it). Basically, it could be the same as `SymbolType` enum but with a string added to each variant to specify the name.

I'm not 100% clear on your proposal here. Are you suggesting something like the following?

```
pub(crate) enum SymbolType {
    /// an identifier for a constant, containing its type and value
    Constant(String, ConstantType),
    /// an identifier for a trace column, containing trace column information with its trace
    /// segment, its size and its offset.
    TraceColumns(String, TraceColumns),
    /// an identifier for a public input, containing the size of the public input array
    PublicInput(String, usize),
    /// an identifier for a periodic column, containing its index out of all periodic columns and
    /// its cycle length in that order.
    PeriodicColumn(String, usize, usize),
    /// an expression or set of expressions associated with a variable
    Variable(String, VariableType),
    /// an identifier for random value, containing its index in the random values array and its
    /// length if this value is an array. For non-array random values second parameter is always 1.
    RandomValuesBinding(String, usize, usize),
}
```

The enums in this codebase are quite heavy and there's a lot of nesting going on, which I think reduces the clarity of the code. I would prefer to keep the `Symbol` as a struct with `name` and `symbol_type`. It feels much cleaner to me, especially given how much the `name` and `symbol_type` are accessed in other places, but I might be misunderstanding your suggestion.

bobbinth:
> I'm not 100% clear on your proposal here. Are you suggesting something like the following?

Yes, that's what I was thinking, but my understanding was that the conversion to `Value` would be pretty localized. If that's not the case, let's do what you are suggesting. If anything, we can revisit later.

grjte:
Final comment - I'd prefer to defer any localized changes until later, since this I want to wrap this refactor up this week so people can touch the IR again next week. Can you let me know what you think of the overall structure now @bobbinth? If you're generally happy with it I'll try to figure out how to split it into a few smaller PRs for detailed review

grjte:
> Yes, that's what I was thinking, but my understanding was that the conversion to `Value` would be pretty localized. If that's not the case, let's do what you are suggesting. If anything, we can revisit later.

It might be localized after list comprehension/folding get updated for the new value handling, but for the moment it's not, because I haven't really worked those. Since they're more localized, I figured they can be revisited. Let's defer the enum until later, as you said

bobbinth (commented):
The structure looks great! Thank you!

There are a few things we can change in the future - but these things are much more localized. One example is to create the `Constraints` struct only when `ConstraintBuilder` is consumed (via the `into_air()` method). This way, it wouldn't need to contain the pass-through methods to the `graph` instance.

Issue #182, opened 2023-02-27
Write Memory Chiplet constraints in AirScript


Overcastan:

bobbinth:
I believe this is currently blocked by https://github.com/0xMiden/miden-vm/issues/1614.

Al-Kindi-0:
Indeed, should we prioritize it then ?

bobbinth:
> Indeed, should we prioritize it then ?

I think so. I believe you already had a proposal for how to solve it - it just needs to be reviewed, right?

Al-Kindi-0:
Yes, and I think the solution should be implemented as well to make sure we haven't missed anything which basically means doing the range checks and sending them through the ACE wiring bus.

Issue #187, opened 2023-03-01
Add support for ConstraintComprehension


tohrnii:
We should add support for statements like these:

```
enf a = a^2 for a in a_limb
```
OR
```
enf x = y for (x, y) in (s1, s2)
```
We could handle this similar to how we handle list comprehension where we unfold this into multiple constraints in the IR for each `a` in `a_limb` etc.

bobbinth:
Closing as this seems to be already supported.

PR #191, opened 2023-03-03
Write Memory Chiplet constraints in AirScript

Overcastan:
Addressing: #182 
This PR adds Memory Chiplet constraints.

### DIFF
--- a/constraints/miden-vm/memory.air
+++ b/constraints/miden-vm/memory.air
@@ -1,27 +1,20 @@
-def MemoryAir
+mod MemoryAir
 
+### Helper functions/evaluators ###################################################################
 
+# Enforces that columns must be zero.
 ev equal_zero(&main: [4]):
     enf v = 0 for v in &main
 
+# Enforces that values in the columns in the current row must be equal to the values in the next 
+# row.
 ev equal_seq_lines(&main: [4]):
     enf v' = v for v in &main
 
+### Memory Chiplet Air Constraints ################################################################
 
-trace_columns:
-    main: [s[2], ctx, addr, clk, v[4], d[2], t, dummy]
-
-public_inputs:
-    stack_inputs: [16]
-
-random_values:
-    alphas: [9]
-
-boundary_constraints:
-    # This is a dummy trace column to satisfy requirement of at least one boundary constraint.
-    enf dummy.first = 0
-
-integrity_constraints:
+# Enforces the constraints on the memory chiplet, given the columns of the memory execution trace.
+ev memory_chiplet(main: [s[2], ctx, addr, clk, v[4], d[2], t]):
     let delta_c = ctx' - ctx
     let delta_a = addr' - addr
     let delta_i = clk' - clk - 1
--- a/constraints/miden-vm/memory.air
+++ b/constraints/miden-vm/memory.air
@@ -1,66 +1,124 @@
-mod MemoryAir
+mod MemoryChipletAir
 
-### Helper functions/evaluators ###################################################################
+### Helper functions ##############################################################################
 
-# Enforces that columns must be zero.
-ev equal_zero(&main: [4]):
-    enf v = 0 for v in &main
+fn delta(v: scalar) -> scalar:
+    return v' - v
+
+
+# Returns the n0 flag which is set to 1 when context changes and 0 otherwise.
+fn get_n0(ctx: scalar, t: scalar) -> scalar:
+    return delta(ctx) * t'
+
+
+# Returns the n1 flag. If context remains the same, n1 = 1 when address changes and 0 otherwise.
+fn get_n1(addr: scalar, t: scalar) -> scalar:
+    return delta(addr) * t'
+
+
+### Helper evaluators #############################################################################
+
+# Enforces that the provided columns must be zero.
+ev is_zero(main: [values[4]]):
+    enf v = 0 for v in values
+
 
 # Enforces that values in the columns in the current row must be equal to the values in the next 
 # row.
-ev equal_seq_lines(&main: [4]):
-    enf v' = v for v in &main
+ev is_unchanged(main: [values[4]]):
+    enf v' = v for v in values
+
+
+# Enforces that the provided columns must be binary.
+ev is_binary(main: [a]):
+    enf a^2 = a
+
+
+# Enforces that created flags have valid values during the program execution.
+ev flags_validity(main: [ctx, addr, t]):
+    # n0 = 1 when context changes and 0 otherwise.
+    let n0 = get_n0(ctx, t)
+
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    let n1 = get_n1(addr, t)
+
+    # Enforce that n0 must be binary.
+    enf n0^2 = n0
+
+    # Enforce that when context changes, n0 = 1 (or when n0 = 0, context remains the same).
+    enf ctx' = ctx when !n0
+
+    # Enforce that n1 must be binary. An additional condition ensures that the check of n1 
+    # occurs only if the context does not change (n0 = 0).
+    enf n1^2 = n1 when !n0
+
+    # Enforce that if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    enf addr' = addr when !n0 & !n1
+
+
+# Enforces that selectors take the correct values under certain conditions.
+ev enforce_selectors(main: [s[2], ctx, addr, t]):
+    # Enforce that values in the selectior columns must be binary.
+    # s[0] is set to 0 for write operations and to 1 for read operations.
+    enf is_binary([selector]) for selector in s
+
+    # n0 = 1 when context changes and 0 otherwise.
+    let n0 = get_n0(ctx, t)
+
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    let n1 = get_n1(addr, t)
+
+    # Enforce that s[1]' = 1 when the operation is a read and `ctx` and `addr` columns are both 
+    # unchanged.
+    enf s[1]' = 1 when !n0 & !n1 & s[0]'
+
+    # Enforce that s[1]' = 0 when either the context changed, the address changed, or the operation
+    # is a write.
+    enf s[1]' = 0 when n0 | n1 | !s[0]'
+
+
+# Enforces that the delta between two consecutive contexts, addresses, or clock cycles is updated 
+# and decomposed into the `d1` and `d0` columns correctly.
+ev enforce_delta(main: [ctx, addr, clk, d[2], t]):
+    # n0 = 1 when context changes and 0 otherwise.
+    let n0 = get_n0(ctx, t)
+
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    let n1 = get_n1(addr, t)
+
+    let d_next_agg = 2^16 * d[1]' + d[0]'
+
+    # Enforce that values of context (`ctx`), address (`addr`), and clock cycle (`clk`) grow 
+    # monotonically
+    match enf:
+        d_next_agg = delta(ctx) when n0
+        d_next_agg = delta(addr) when !n0 & n1
+        d_next_agg = delta(clk) - 1 when !n0 & !n1
+
+
+# Enforces that memory is initialized to zero when it is read before being written and that when 
+# existing memory values are read they remain unchanged.
+ev enforce_values(main: [s[2], v[4]]):
+    # Enforce that values at a given memory address are always initialized to 0.
+    enf is_zero([v]) when s[0] & !s[1]
+
+    # Enforce that for the same context/address combination, the v columns of the current row are
+    # equal to the corresponding v columns of the next row
+    enf is_unchanged([v]) when s[1]
+
 
 ### Memory Chiplet Air Constraints ################################################################
 
 # Enforces the constraints on the memory chiplet, given the columns of the memory execution trace.
 ev memory_chiplet(main: [s[2], ctx, addr, clk, v[4], d[2], t]):
-    let delta_c = ctx' - ctx
-    let delta_a = addr' - addr
-    let delta_i = clk' - clk - 1
+    enf flags_validity([ctx, addr, t])
 
-    # n0 = 1 when context changes and 0 otherwise.
-    let n0 = delta_c * t'
-    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
-    let n1 = delta_a * t'
+    enf enforce_selectors([s, ctx, addr, t])
 
-    # Enforce that n0 must be binary.
-    enf n0^2 = n0
-    # Enforce that when context changes, n0 = 1 (or when n0 = 0, context remains the same).
-    enf delta_c = 0 when !n0
-    # Enforce that n1 must be binary. An additional condition ensures that the check of n1 
-    # occurs only if the context does not change (n0 = 0).
-    enf n1^2 = n1 when !n0
-    # Enforce that if context remains the same, n1 = 1 when address changes and 0 otherwise.
-    enf delta_a = 0 when !n0 & !n1
-
-    # Enforce that values in the selectior columns must be binary.
-    # s[0] is set to 0 for write operations and to 1 for read operations.
-    enf col = col^2 for col in s
-
-    # Enforce that s[1]' = 1 when the operation is a read and `ctx` and `addr` columns are both 
-    # unchanged.
-    enf s[1]' = 1 when !n0 & !n1 & s[0]'
-    # Enforce that when either the context changed, the address changed, or the operation is a 
-    # write, then s[1]' = 0
-    enf s[1]' = 0 when n0 | n1 | !s[0]'
-
-    let d_next_aggr = 2^16 * d[1]' + d[0]'
-
-    # Enforce that values of context (`ctx`), address (`addr`), and clock cycle (`clk`) grow 
-    # monotonically
-    match enf:
-        d_next_aggr = delta_c when n0
-        d_next_aggr = delta_a when !n0 & n1
-        d_next_aggr = delta_i when !n0 & !n1
+    enf enforce_delta([ctx, addr, clk, d, t])
 
     # TODO: perform range checks for values in columns d[0] and d[1]
 
-    # Enforce that values at a given memory address are always initialized to 0.
-    enf equal_zero([v]) when s[0] & !s[1]
-
-    # Enforce that for the same context/address combination, the v columns of the current row are
-    # equal to the corresponding v columns of the next row
-    enf equal_seq_lines([v]) when s[1]
+    enf enforce_values([s, v])
 
     # TODO: add constraint for chiplet bus
--- a/constraints/miden-vm/memory.air
+++ b/constraints/miden-vm/memory.air
@@ -1,0 +1,73 @@
+def MemoryAir
+
+
+ev equal_zero(&main: [4]):
+    enf v = 0 for v in &main
+
+ev equal_seq_lines(&main: [4]):
+    enf v' = v for v in &main
+
+
+trace_columns:
+    main: [s[2], ctx, addr, clk, v[4], d[2], t, dummy]
+
+public_inputs:
+    stack_inputs: [16]
+
+random_values:
+    alphas: [9]
+
+boundary_constraints:
+    # This is a dummy trace column to satisfy requirement of at least one boundary constraint.
+    enf dummy.first = 0
+
+integrity_constraints:
+    let delta_c = ctx' - ctx
+    let delta_a = addr' - addr
+    let delta_i = clk' - clk - 1
+
+    # n0 = 1 when context changes and 0 otherwise.
+    let n0 = delta_c * t'
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    let n1 = delta_a * t'
+
+    # Enforce that n0 must be binary.
+    enf n0^2 = n0
+    # Enforce that when context changes, n0 = 1 (or when n0 = 0, context remains the same).
+    enf delta_c = 0 when !n0
+    # Enforce that n1 must be binary. An additional condition ensures that the check of n1 
+    # occurs only if the context does not change (n0 = 0).
+    enf n1^2 = n1 when !n0
+    # Enforce that if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    enf delta_a = 0 when !n0 & !n1
+
+    # Enforce that values in the selectior columns must be binary.
+    # s[0] is set to 0 for write operations and to 1 for read operations.
+    enf col = col^2 for col in s
+
+    # Enforce that s[1]' = 1 when the operation is a read and `ctx` and `addr` columns are both 
+    # unchanged.
+    enf s[1]' = 1 when !n0 & !n1 & s[0]'
+    # Enforce that when either the context changed, the address changed, or the operation is a 
+    # write, then s[1]' = 0
+    enf s[1]' = 0 when n0 | n1 | !s[0]'
+
+    let d_next_aggr = 2^16 * d[1]' + d[0]'
+
+    # Enforce that values of context (`ctx`), address (`addr`), and clock cycle (`clk`) grow 
+    # monotonically
+    match enf:
+        d_next_aggr = delta_c when n0
+        d_next_aggr = delta_a when !n0 & n1
+        d_next_aggr = delta_i when !n0 & !n1
+
+    # TODO: perform range checks for values in columns d[0] and d[1]
+
+    # Enforce that values at a given memory address are always initialized to 0.
+    enf equal_zero([v]) when s[0] & !s[1]
+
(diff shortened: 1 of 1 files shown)


air-script/tests/memory/memory.air
```diff
@@ -0,0 +1,71 @@
...
+
+    # Enforce that for the same context/address combination, the v[i] columns of the current row 
+    # are equal to the corresponding u[i] columns of the next row, for i from 0 to 3
+    enf s[1] * (v[0]' - v[0]) = 0
+    enf s[1] * (v[1]' - v[1]) = 0
+    enf s[1] * (v[2]' - v[2]) = 0
+    enf s[1] * (v[3]' - v[3]) = 0
+
+    # TODO: add constraint for chiplet bus
+    # How can I get chiplet bus value (`b_chip`) and operation label value?
```
Overcastan, line 71: How can these values be provided within this script? Are they a part of public inputs?

air-script/tests/memory/memory.air
```diff
@@ -0,0 +1,71 @@
...
+    # s[0] is set to 0 for write operations and to 1 for read operations
+    enf s[0]^2 - s[0] = 0
+    enf s[1]^2 - s[1] = 0
+
+    # Enforce that s[1] = 1 when the operation is a read and `ctx` and `addr` columns are both 
+    # unchanged.
+    (1 - n_0) * (1 - n_1) * s[0]' * (1 - s[1]') = 0
+    # Enforce that when either the context changed, the address changed, or the operation is a 
+    # write, then s[1] = 0
+    (n_0 + (1 - n_0) * n_1 + (1 - s[0]')) * s[1]' = 0
```
tohrnii, line 46: I think `enf` keyword is missing here.

air-script/tests/memory/memory.air
```diff
@@ -0,0 +1,71 @@
...
+    enf (n_0 * delta_c + (1 - n_0) * (n_1 * delta_a + (1 - n_1) * delta_i)) - d_next_aggr = 0
+
+    # TODO: perform range checks for values in columns d[0] and d[1]
+    # How can I call a range check for some columns?
+
+    # Enforce that values at a given memory address are always initialized to 0.
+    s[0] * (1 - s[1]) * v[0] = 0
+    s[0] * (1 - s[1]) * v[1] = 0
+    s[0] * (1 - s[1]) * v[2] = 0
+    s[0] * (1 - s[1]) * v[3] = 0
```
tohrnii, line 61: I think enf keyword is missing here.

air-script/tests/memory/memory.air
```diff
@@ -0,0 +1,71 @@
...
+
+    # Enforce that for the same context/address combination, the v[i] columns of the current row 
+    # are equal to the corresponding u[i] columns of the next row, for i from 0 to 3
+    enf s[1] * (v[0]' - v[0]) = 0
+    enf s[1] * (v[1]' - v[1]) = 0
+    enf s[1] * (v[2]' - v[2]) = 0
+    enf s[1] * (v[3]' - v[3]) = 0
```
tohrnii, line 68: Question: Should we wait to merge this and first merge constraint comprehension PR first? I think we should be able to merge it sometime next week and we'll be able to write these constraints in a much more concise way once that PR is merged.
cc: @grjte

tohrnii (commented):
Not a review, just wondering if we should wait to merge this till #187 is merged because the constraints would be much more concise after that.

air-script/tests/memory/memory.air
tohrnii: Or maybe we could mention it in comments that these could probably be written as:
```
fn change(v: scalar) -> scalar:
    return v' - v
enf s[1] * change(v[i]) = 0 for i in 0..4
```
once pure functions and constraint comprehension are implemented.

Wdyt?

Overcastan: I agree, it looks much more concise. I think it's worth waiting if it doesn't create additional problems or delays.

grjte: I don't think we should wait. I think we should just move where we put this file. Instead of putting these constraints into the `tests`, we should create a new directory and put all our constraint definitions for the miden vm there. That way the VM constraints can be defined before the syntax is supported. If we don't do that, I don't think we'll get everything done in time.

grjte: I don't think we should wait. I think we should just move where we put this file. Instead of putting these constraints into the `tests`, we should create a new directory and put all our constraint definitions for the miden vm there. That way the VM constraints can be defined before the syntax is supported. If we don't do that, I don't think we'll get everything done in time.

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,64 @@
...
+    enf (1 - n_0) * delta_c = 0
+    # Enforce that n_1 must be binary. An additional multiplier ensures that the check of n_1 
+    # occurs only if the context does not change.
+    enf (1 - n_0) * (n_1^2 - n_1) = 0
+    # Enforce that if context remains the same, n_1 = 1 when address changes and 0 otherwise.
+    enf (1 - n_0) * (1 - n_1) * delta_a = 0
+
+    # Enforce that values in the selectior columns must be binary.
+    # s[0] is set to 0 for write operations and to 1 for read operations
+    enf s[i]^2 - s[i] = 0 for i in (0..2)
```
bobbinth, line 38: Is there a reason not to write this as something like this:
```
enf col = col^2 for col in s
```

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,64 @@
...
+    enf (1 - n_0) * (1 - n_1) * s[0]' * (1 - s[1]') = 0
+    # Enforce that when either the context changed, the address changed, or the operation is a 
+    # write, then s[1] = 0
+    enf (n_0 + (1 - n_0) * n_1 + (1 - s[0]')) * s[1]' = 0
+
+    let d_next_aggr = 2^16 * d[1]' + d[0]'
+
+    # Enforce that values of context (`ctx`), address (`addr`), and clock cycle (`clk`) grow 
+    # monotonically
+    enf (n_0 * delta_c + (1 - n_0) * (n_1 * delta_a + (1 - n_1) * delta_i)) - d_next_aggr = 0
```
bobbinth, line 51: If we are planning to support selector syntax for v0.3, then this could be something like:
```
match enf:
  d_next_agg = delta_c when n_0
  d_next_agg = delta_a when (1 - n_0) * n_1
  d_next_agg = delta_i when (1 - n_0) * (1 - n_1)
```
A couple other thoughts regarding improving readability:
- Instead of `n_0` and `n_1` we could probably use `n0` and `n1`, or maybe even a vector `n` with two items - though, not sure we need to go as far as using a vector.
- I wonder if we should allow syntactic sugar for logical expressions on the right side of `when`.

For example, the above could look like:
```
match enf:
  d_next_agg = delta_c when n0
  d_next_agg = delta_a when !n0 & n1
  d_next_agg = delta_i when !n0 & !n1
```

Overcastan:
This PR has a problem: due to the widespread use of variables defined at the beginning, it is hard to rationally split the code without passing these variables to subfunctions.

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
+mod MemoryAir
+
+### Helper functions/evaluators ###################################################################
+
+# Enforces that columns must be zero.
+ev equal_zero(&main: [4]):
```
tohrnii, line 6: The syntax doesn't look correct here. I think we need named columns here and we shouldn't use `&`? Was there a discussion on using this syntax that maybe I missed?

nit: Should we rename this to `is_zero`?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+    enf v = 0 for v in &main
+
+# Enforces that values in the columns in the current row must be equal to the values in the next 
+# row.
+ev equal_seq_lines(&main: [4]):
```
tohrnii, line 11: The syntax doesn't look correct here. I think we need named columns here and we shouldn't use `&`?

nit: I'm not sure I understand what the name is supposed to mean?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+ev flags_validity(main: [s_0], n0, n1, delta_a, delta_c):
+    # Enforce that n0 must be binary.
+    enf n0^2 = n0
+    # Enforce that when context changes, n0 = 1 (or when n0 = 0, context remains the same).
+    enf delta_c = 0 when !n0
+    # Enforce that n1 must be binary. An additional condition ensures that the check of n1 
+    # occurs only if the context does not change (n0 = 0).
+    enf n1^2 = n1 when !n0
+    # Enforce that if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    enf delta_a = 0 when !n0 & !n1
```
tohrnii, line 24: nit: Should we define `is_binary` evaluator instead of `enf n0^2 = n0` and enf `n1^2 = n1`?

nit: I think we should add spaces to improve readability, i.e.

```
    # Enforce that n0 must be binary.
    enf n0^2 = n0

    # Enforce that when context changes, n0 = 1 (or when n0 = 0, context remains the same).
    enf delta_c = 0 when !n0

    # Enforce that n1 must be binary. An additional condition ensures that the check of n1 
    # occurs only if the context does not change (n0 = 0).
    enf n1^2 = n1 when !n0

    # Enforce that if context remains the same, n1 = 1 when address changes and 0 otherwise.
    enf delta_a = 0 when !n0 & !n1
```

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+
+    # Enforce that values in the selectior columns must be binary.
+    # s[0] is set to 0 for write operations and to 1 for read operations.
+    enf col = col^2 for col in s
```
tohrnii, line 28: nit: Should we define is_binary evaluator instead of `enf col = col^2`?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+
+# Enforces that selectors take the correct values under certain conditions.
+ev selectors_correctness(main: [s[2]], n0, n1):
```
tohrnii, line 31: nit: I think we settled on only allowing main and aux columns in evaluator functions, so the signature should probably be:

`ev selectors_correctness(main: [s[2], n0, n1])`

nit: I think we should rename the function. Maybe `enforce_selectors`?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+    # Enforce that s[1]' = 1 when the operation is a read and `ctx` and `addr` columns are both 
+    # unchanged.
+    enf s[1]' = 1 when !n0 & !n1 & s[0]'
+    # Enforce that when either the context changed, the address changed, or the operation is a 
+    # write, then s[1]' = 0
+    enf s[1]' = 0 when n0 | n1 | !s[0]'
```
tohrnii, line 37: nit: We should add blank line to improve readability.

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+
+# Enforces that certain columns grow monotonically.
+ev monotonically_grow(n0, n1, delta_a, delta_c, delta_i):
```
tohrnii, line 40: The signature looks incorrect. Should it be `ev monotonically_grow(main: [n0, n1, delta_a, delta_c, delta_i])`?

nit: Should we rename this to `grow_monotonically`?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+    let delta_c = ctx' - ctx
+    let delta_a = addr' - addr
+    let delta_i = clk' - clk - 1
+
+    # n0 = 1 when context changes and 0 otherwise.
+    let n0 = delta_c * t'
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    let n1 = delta_a * t'
+
+    enf flags_validity([s[0]], n0, n1, delta_a, delta_c)
```
tohrnii, line 63: I think the syntax isn't correct here. It should probably be:
`enf flags_validity([s[0], n0, n1, delta_a, delta_c])`

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+        d_next_aggr = delta_a when !n0 & n1
+        d_next_aggr = delta_i when !n0 & !n1
+
+
+### Memory Chiplet Air Constraints ################################################################
+
+# Enforces the constraints on the memory chiplet, given the columns of the memory execution trace.
+ev memory_chiplet(main: [s[2], ctx, addr, clk, v[4], d[2], t]):
+    let delta_c = ctx' - ctx
+    let delta_a = addr' - addr
```
tohrnii, line 55: maybe we could define a function:
```
fn delta(v: scalar) -> scalar:
    return v' - v
```

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+    let delta_i = clk' - clk - 1
+
+    # n0 = 1 when context changes and 0 otherwise.
+    let n0 = delta_c * t'
```
tohrnii, line 59: if we define a function for `v' - v`, then we could just use that here.
```
let n0 = delta(c) * t'
```

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    let n1 = delta_a * t'
+
+    enf flags_validity([s[0]], n0, n1, delta_a, delta_c)
+
+    enf selectors_correctness([s], n0, n1)
```
tohrnii, line 65: Should this be `enf selectors_correctness([s, n0, n1])`?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,78 @@
...
+
+    enf monotonically_grow(n0, n1, delta_a, delta_c, delta_i)
```
tohrnii, line 67: Should this be `enf monotonically_grow([n0, n1, delta_a, delta_c, delta_i])`?

tohrnii (changes requested):
Thank you @Overcastan, I left a few comments.

constraints/miden-vm/memory.air
Overcastan: I was trying to come up with some name that was supposed to express that two consecutive lines should be equal, so I came up with "equal sequential lines", which was shortened to `equal_seq_lines`. I'm happy to replace that name, but I haven't been able to come up with something better yet. Maybe `equal_cons_lines`?

constraints/miden-vm/memory.air
Overcastan: As I understood. for now we decided to be strict and only pass columns to the evaluators, so I can't pass variables `n0` and `n1` to the evaluator.

tohrnii: Ah, sorry, you're right. My bad.

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
+mod MemoryChipletAir
+
+### Helper functions/evaluators ###################################################################
+
+fn delta(v: scalar) -> scalar:
+    return v' - v
+
+
+# Enforces that columns must be zero.
+ev is_zero(main: [value[4]]):
```
tohrnii, line 10: nit: Should this be `values` instead of `value` or is the idea to consider a `word` a single `value`? If so, should we rename this to `word` instead? Wdyt?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
...
+    enf v = 0 for v in value
+
+
+# Enforces that values in the columns in the current row must be equal to the values in the next 
+# row.
+ev equal_seq_lines(main: [value[4]]):
```
tohrnii, line 16: Question: Do you think a more appropriate name here would be something like `is_unchanged`?

nit: Should this be values instead of value or is the idea to consider a word a single value? If so, should we rename this to word instead? Wdyt?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
+mod MemoryChipletAir
+
+### Helper functions/evaluators ###################################################################
+
+fn delta(v: scalar) -> scalar:
+    return v' - v
+
+
+# Enforces that columns must be zero.
```
tohrnii, line 9: nit: Should we change this to `Enforces that the provided column must be zero.`?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
...
+    enf v = 0 for v in value
+
+
+# Enforces that values in the columns in the current row must be equal to the values in the next 
+# row.
+ev equal_seq_lines(main: [value[4]]):
+    enf v' = v for v in value
+
+
+# Enforces that column must be binary.
```
tohrnii, line 20: nit: Should we change this to `Enforces that the provided column must be binary.`?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
...
+ev is_binary(main: [a]):
+    enf a^2 - a = 0
+
+
+# Enforces that created flags and s[0] column have valid values during the program execution.
+ev flags_validity(main: [s[2], ctx, addr, t]):
+    # n0 = 1 when context changes and 0 otherwise.
+    let n0 = delta(ctx) * t'
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    let n1 = delta(addr) * t'
```
tohrnii, line 30: nit: I'd put a blank line between these two constraints as well.

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
...
+
+    # Enforce that n0 must be binary.
+    enf n0^2 = n0
+
+    # Enforce that when context changes, n0 = 1 (or when n0 = 0, context remains the same).
+    enf delta_c = 0 when !n0
```
tohrnii, line 36: `delta_c` is undefined. Should this be `delta(ctx)` instead?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
...
+
+    # Enforce that n1 must be binary. An additional condition ensures that the check of n1 
+    # occurs only if the context does not change (n0 = 0).
+    enf n1^2 = n1 when !n0
+
+    # Enforce that if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    enf delta_a = 0 when !n0 & !n1
```
tohrnii, line 43: `delta_a` is undefined. Should this be `delta(addr)` instead?

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
...
+
+    # Enforce that values in the selectior columns must be binary.
+    # s[0] is set to 0 for write operations and to 1 for read operations.
+    enf is_binary([s_i]) for s_i in s
```
tohrnii, line 47: nit: I'd rename `s_i` to `selector`.

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
...
+
+
+# Enforces that selectors take the correct values under certain conditions.
+ev enforce_selectors(main: [s[2], ctx, addr, t]):
+    # n0 = 1 when context changes and 0 otherwise.
+    let n0 = delta(ctx) * t'
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
```
tohrnii, line 54: nit: I'd add a blank line in between the constraints.

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
...
+    let n1 = delta(addr) * t'
+
+    # Enforce that s[1]' = 1 when the operation is a read and `ctx` and `addr` columns are both 
+    # unchanged.
+    enf s[1]' = 1 when !n0 & !n1 & s[0]'
+
+    # Enforce that when either the context changed, the address changed, or the operation is a 
+    # write, then s[1]' = 0
```
tohrnii, line 62: nit: Should we keep the order of the comment same as above, i.e.,
```
# Enforce that s[1]' = 0 when either the context changed, the address changed, or the operation is a 
# write.
```

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,101 @@
...
+    enf s[1]' = 0 when n0 | n1 | !s[0]'
+
+
+# Enforces that certain columns grow monotonically.
+ev grow_monotonically(main: [ctx, addr, clk, d[2], t]):
+    # n0 = 1 when context changes and 0 otherwise.
+    let n0 = delta(ctx) * t'
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
```
tohrnii, line 70: nit: I'd add a blank line.

tohrnii (changes requested):
Thank you @Overcastan, it looks good overall now. I left a few more comments.

constraints/miden-vm/memory.air
Overcastan: I agree, that sounds better. But should we say `columns` instead of `column`, since we are checking 4 columns at once?

tohrnii: Yeah, it should be `columns`.

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,104 @@
...
+
+
+# Enforces that the provided columns must be zero.
+ev is_zero(main: [values[4]]):
+    enf v = 0 for v in values
+
+
+# Enforces that values in the columns in the current row must be equal to the values in the next 
+# row.
+ev equal_seq_lines(main: [values[4]]):
```
tohrnii, line 16: I still think maybe we should probably pick a better name here. To me something like `is_unchanged` seems slightly better. It would probably be a good idea to think of a good name for evaluator like this since we will probably have constraints like these very commonly.

tohrnii (approved):
Thank you @Overcastan, looks good to me. I left one naming related nit inline.

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,104 @@
...
+    let n0 = delta(ctx) * t'
+
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    let n1 = delta(addr) * t'
+
+    # Enforce that n0 must be binary.
+    enf n0^2 = n0
+
+    # Enforce that when context changes, n0 = 1 (or when n0 = 0, context remains the same).
+    enf delta(ctx) = 0 when !n0
```
tohrnii, line 37: optional nit: Should we define an evaluator for this without the selector instead?

```
enf remains_same(ctx) = 0 when !n0
```

tohrnii (approved):

constraints/miden-vm/memory.air
Overcastan: Oops, forgot to change it. Yes, `is_unchanged` seems much better, I'll fix it right away.

constraints/miden-vm/memory.air
Overcastan: Not sure that it worth it. Maybe write it like `enf ctx' = ctx when !n0`, even without `delta` evaluator?

tohrnii: Yeah, you're right, `enf ctx' = ctx when !n0` seems cleaner than the alternatives.

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,104 @@
...
+# Enforces that created flags and s[0] column have valid values during the program execution.
+ev flags_validity(main: [s[2], ctx, addr, t]):
+    # n0 = 1 when context changes and 0 otherwise.
+    let n0 = delta(ctx) * t'
+
+    # if context remains the same, n1 = 1 when address changes and 0 otherwise.
+    let n1 = delta(addr) * t'
+
+    # Enforce that n0 must be binary.
+    enf n0^2 = n0
```
grjte, line 34: Similar to my earlier comment about consistency - in the hash chiplet PR we use `b^2 - b = 0`. I prefer this way of doing it, and would prefer to do it in the same way everywhere. This would suggest changing to `b^2 = b` in the hash chiplet PR

grjte (changes requested):
Looks good, thank you! I haven't checked the constraints for correctness yet, but I've left a few comments inline about the code style

constraints/miden-vm/memory.air
```diff
@@ -0,0 +1,122 @@
...
+fn delta(v: scalar) -> scalar:
+    return v' - v
+
+
+fn get_n0(ctx: scalar, t: scalar) -> scalar:
+    return delta(ctx) * t'
+
+
+fn get_n1(addr: scalar, t: scalar) -> scalar:
+    return delta(addr) * t'
```
grjte, line 14: nit: let's add comments above these functions to explain n0 and n1. You could use the same ones that are inline below

grjte (approved):
Looks good! I left one final nit inline. Can you update that and then merge?

PR #193, opened 2023-03-06
Unroll exponentiation expressions during code generation

### LINKED ISSUE #130: unroll exponentiation into multiplications in winter codegen
We should remember that using `exp` could be quite expensive as it uses constant-time exponentiation. In the future, we might want o "unroll" this into a sequence of multiplications.

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/air-script/pull/126#discussion_r1090953576_

jjcnn:
Fixes #130.

Exponentiation expressions are now unrolled during code generation for better performance.

### DIFF
--- a/air-script/tests/binary/binary.rs
+++ b/air-script/tests/binary/binary.rs
@@ -75,8 +75,8 @@ impl Air for BinaryAir {
     fn evaluate_transition<E: FieldElement<BaseField = Felt>>(&self, frame: &EvaluationFrame<E>, periodic_values: &[E], result: &mut [E]) {
         let main_current = frame.current();
         let main_next = frame.next();
-        result[0] = main_current[0].exp(E::PositiveInteger::from(2_u64)) - main_current[0] - E::ZERO;
-        result[1] = main_current[1].exp(E::PositiveInteger::from(2_u64)) - main_current[1] - E::ZERO;
+        result[0] = (main_current[0] * main_current[0]) - main_current[0] - E::ZERO;
+        result[1] = (main_current[1] * main_current[1]) - main_current[1] - E::ZERO;
     }
 
     fn evaluate_aux_transition<F, E>(&self, main_frame: &EvaluationFrame<F>, aux_frame: &EvaluationFrame<E>, _periodic_values: &[F], aux_rand_elements: &AuxTraceRandElements<E>, result: &mut [E])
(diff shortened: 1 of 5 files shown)


bobbinth:
I only looked at this briefly - so, might have missed something.

I think using a simple chain of multiplications would not give us the best result. For example, $x^7$ would require 7 multiplications while we can compute it using only 4 multiplications as:
```
fn exp7(a):
  let a2 = a * a;
  let a4 = a2 * a2;
  let a3 = a2 * a;
  return a3 * a4;
```

jjcnn:
Ah, yes - that's what was meant by linear exponentiation in #130.

Your suggestion can only be implemented by introducing fresh variables in the generated code, and I don't know how to do that. I can try to figure it out if you think it's worth my time?

Another thing I don't know is how to test the implementation. None of our current tests have exponents greater than 3, and that would lead to the same number of multiplications as without the rewrite.

bobbinth:
> Your suggestion can only be implemented by introducing fresh variables in the generated code, and I don't know how to do that. I can try to figure it out if you think it's worth my time?

I'd probably create a set of functions for exponentiation where the exponent is greater than 3 - e.g., `exp4`, `exp5` etc. Not sure how easy it would be to create a set of these functions dynamically - so, we can even consider pre-creating some set of them. We can probably safely assume that the exponent is not going be greater than 32 or something like that (for Miden, the exponent will never be greater than 9).

> Another thing I don't know is how to test the implementation. None of our current tests have exponents greater than 3, and that would lead to the same number of multiplications as without the rewrite.

If we don't have them already, we should soon write constraints for the ROP hash function. These would have exponentiation to power 7.

tohrnii:
Question: Can't we generalize this with something like:

```
fn pow<E: FieldElement<BaseField = Felt>>(base: E, exp: u64) -> E {
    if exp == 1 {
        return base;
    } else if exp % 2 == 0 {
        let result = pow(base, exp / 2);
        return result * result;
    } else {
        let result = pow(base, (exp - 1)/2);
        return base * result * result;
    }
}
```
or am I missing something?

bobbinth:
@tohrnii - we can't use conditional statements or `%` operation when computing AIR constraints. We can use only field operations (or statements which reduce to field operations).

tohrnii:
@bobbinth - Ah, got it. Thanks.

bobbinth:
Superseded by #333.

PR #195, opened 2023-03-07
Refactor the IR (2 of 4)

grjte:
This is the second of 4 PRs refactoring the IR, addressing https://github.com/0xPolygonMiden/air-script/issues/150.

The primary goals are:

- simplify & clarify the code
- reduce duplication
- simplify error handling

In this PR, the AST is consumed during the construction of the IR, which will reduce some cloning of data.

### DIFF
--- a/parser/src/ast/pub_inputs.rs
+++ b/parser/src/ast/pub_inputs.rs
@@ -27,4 +27,8 @@ impl PublicInput {
     pub fn size(&self) -> usize {
         self.size
     }
+
+    pub fn into_parts(self) -> (String, usize) {
+        (self.name.into_name(), self.size)
+    }
 }
--- a/air-script-core/src/constant.rs
+++ b/air-script-core/src/constant.rs
@@ -28,6 +28,10 @@ impl Constant {
     pub fn value(&self) -> &ConstantType {
         &self.value
     }
+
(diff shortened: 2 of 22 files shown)


bobbinth (approved):
All looks good! Thank you!

parser/src/ast/pub_inputs.rs
```diff
@@ -27,4 +27,8 @@ impl PublicInput {
     pub fn size(&self) -> usize {
         self.size
     }
+
+    pub fn into_parts(self) -> (String, usize) {
```
tohrnii, line 31: optional minor nit: Might be good to add docs or a todo here and other places.

tohrnii (approved):
Thank you @grjte, looks good to me.

parser/src/ast/pub_inputs.rs
grjte: Yeah, we should have docs, but the whole codebase will need a walk-through for docs updates, which I'm planning to do this afterwards when it won't be blocking anyone else. I didn't want to spend too much time checking these things step-by-step when I was reorganizing the commits and splitting this for the multiple refactors since it'll need a solid review anyway.

Overcastan (approved):

PR #215, opened 2023-03-13
Write Hash Chiplet constraints in AirScript

Overcastan:
Addressing: #202 
This PR adds Hash Chiplet constraints.

### DIFF
--- a/constraints/miden-vm/hash.air
+++ b/constraints/miden-vm/hash.air
@@ -1,90 +1,205 @@
-mod HashAir
+mod HashChipletAir
 
 ### Constants and periodic columns ################################################################
 
 periodic_columns:
-    k0: [0, 0, 0, 0, 0, 0, 0, 1]
-    k1: [0, 0, 0, 0, 0, 0, 1, 0]
-    k2: [1, 0, 0, 0, 0, 0, 0, 0]
+    cycle_row_0: [1, 0, 0, 0, 0, 0, 0, 0]
+    cycle_row_6: [0, 0, 0, 0, 0, 0, 1, 0]
+    cycle_row_7: [0, 0, 0, 0, 0, 0, 0, 1]
 
-### Helper functions/evaluators ###################################################################
+
+### Helper functions ##############################################################################
+
+# Returns binary negation of the value.
+fn binary_not(value: scalar) -> scalar:
+    return 1 - value
+
+
+# Set to 1 when selector flags are (1,0,1) on rows which are multiples of 8. This is flag of 
+# the instruction that initiates Merkle path verification computation.
+fn get_f_mp(s: vector[3]) -> scalar:
+    return cycle_row_0 & s[0] & binary_not(s[1]) & s[2]
+
+
+# Set to 1 when selector flags are (1,1,0) on rows which are multiples of 8. This is flag of 
+# the instruction that initiates Merkle path verification for the "old" node value during 
+# Merkle root update computation.
+fn get_f_mv(s: vector[3]) -> scalar:
+    return cycle_row_0 & s[0] & s[1] & binary_not(s[2])
+
+
+# Set to 1 when selector flags are (1,1,1) on rows which are multiples of 8. This is flag of 
+# the instruction that initiates Merkle path verification for the "new" node value during
+# Merkle root update computation.
+fn get_f_mu(s: vector[3]) -> scalar:
+    return cycle_row_0 & s[0] & s[1] & s[2]
+
+
+# Set to 1 when selector flags are (1,0,0) on rows which are 1 less than a multiple of 8. This 
+# is flag of the instruction that absorbs a new set of elements into the hasher state when 
+# computing a linear hash of many elements.
+fn get_f_abp(s: vector[3]) -> scalar:
+    return cycle_row_7 & s[0] & binary_not(s[1]) & binary_not(s[2])
+
+
+# Set to 1 when selector flags are (1,0,1) on rows which are 1 less than a multiple of 8. This 
+# is flag of the instruction that absorbs the next Merkle path node into the hasher state 
+# during Merkle path verification computation.
+fn get_f_mpa(s: vector[3]) -> scalar:
+    return cycle_row_7 & s[0] & binary_not(s[1]) & s[2]
+
+
+# Set to 1 when selector flags are (1,1,0) on rows which are 1 less than a multiple of 8. This 
+# is flag of the instruction that absorbs the next Merkle path node into the hasher state 
+# during Merkle path verification for the "old" node value during Merkle root update 
+# computation.
+fn get_f_mva(s: vector[3]) -> scalar:
+    return cycle_row_7 & s[0] & s[1] & binary_not(s[2])
+
+
+# Set to 1 when selector flags are (1,1,1) on rows which are 1 less than a multiple of 8. This 
+# is flag of the instruction that absorbs the next Merkle path node into the hasher state 
+# during Merkle path verification for the "new" node value during Merkle root update 
+# computation.
+fn get_f_mua(s: vector[3]) -> scalar:
+    return cycle_row_7 & s[0] & s[1] & s[2]
+
+
+# We can define two flags:
+# 1. Flag f_hout = cycle_row_7 & binary_not(s[0]) & binary_not(s[1]) & binary_not(s[2]), 
+# which is set to 1 when selector flags are (0,0,0) on rows which are 1 less than a multiple 
+# of 8. This is flag of the instruction that returns the result of the currently running 
+# computation.
+# 2. Flag f_sout = cycle_row_7 & binary_not(s[0]) & binary_not(s[1]) & s[2], which is set to 1 
+# when selector flags are (0,0,1) on rows which are 1 less than a multiple of 8. This is flag 
+# of the instruction that returns the whole hasher state.
+#
+# Flag f_out is set to 1 when either f_hout = 1 or f_sout = 1 in the current row.
+fn get_f_out(s: vector[3]) -> scalar:
+    return cycle_row_7 & binary_not(s[0]) & binary_not(s[1])
+
+
+# Flag f_out_next is set to 1 when either f_hout = 1 or f_sout = 1 in the next row.
+fn get_f_out_next(s: vector[3]) -> scalar:
+    return cycle_row_6 & binary_not(s[0]') & binary_not(s[1]')
+
+
+### Helper evaluators #############################################################################
 
 # Enforces that column must be binary.
 ev is_binary(main: [a]):
-    enf a^2 - a = 0
+    enf a^2 = a
+
+
+# Enforces that value in column is copied over to the next row.
+ev is_unchanged(main: [column]):
+    ev column' = column
+
 
 # Enforce selector columns constraints
-ev selector_columns(main: [s[3]], f_out, f_out_next, f_comp):
+ev selector_columns(main: [s[3]]):
+    let f_out = get_f_out(s)
+    let f_out_next = get_f_out_next(s)
+    let f_abp = get_f_abp(s)
+    let f_mpa = get_f_mpa(s)
+    let f_mva = get_f_mva(s)
+    let f_mua = get_f_mua(s)
+
+    # Flag that is true when the performed operation is one of the represented by flags f_abp, 
+    # f_mpa, f_mva or f_mua
+    let f_comp = f_abp + f_mpa + f_mva + f_mua
+
     # Enforce that selector columns are binary.
-    enf is_binary([s_i]) for s_i in s
+    enf is_binary([selector]) for selector in s
+
     # Enforce that unless f_out = 1 or f_out' = 1, the values in columns s[1] and s[2] are copied 
     # over to the nex row.
-    enf (s[1]' - s[1]) * (1 - f_out_next) * (1 - f_out) = 0
-    enf (s[2]' - s[2]) * (1 - f_out_next) * (1 - f_out) = 0
+    enf is_unchanged([s[1]]) when !f_out & !f_out_next
+    enf is_unchanged([s[2]]) when !f_out & !f_out_next
+
     # Enforce that if any of f_abp, f_mpa, f_mva, f_mua flags is set to 1, the next value of s[0] 
     # is 0.
     enf s[0]' * f_comp = 0
+
     # Enforce that no invalid combinations of flags are allowed.
-    enf k0 * (1 - s[0]) * s[1] = 0
+    enf cycle_row_7 * binary_not(s[0]) * s[1] = 0
+
 
 # Enforce node index constraints
-ev node_index(main: [i], b, f_an, f_out):
+ev node_index(main: [s[3], i]):
+    let f_out = get_f_out(s)
+    let f_mp = get_f_mp(s)
+    let f_mv = get_f_mv(s)
+    let f_mu = get_f_mu(s)
+    let f_mpa = get_f_mpa(s)
+    let f_mva = get_f_mva(s)
+    let f_mua = get_f_mua(s)
+
+    # b is the value of the bit which is discarded during shift by one bit to the right.
+    let b = i - 2 * i'
+
+    # Flag that allows to enforce constraint that b is binary only when a new node is absorbed into
+    # the hasher state (when the hash operation is one of Merkle path verification operations or 
+    # next Merkle path node absorption operations)
+    let f_an = f_mp + f_mv + f_mu + f_mpa + f_mva + f_mua
+
     # Enforce that b is binary only when a new node is absorbed into the hasher state.
     enf f_an * (b^2 - b) = 0
+
     # Enforce that when a computation is finished i = 0.
     enf f_out * i = 0
+
     # Enforce that the value in i is copied over to the next row unless we are absorbing a new row 
     # or the computation is finished.
-    enf (1 - f_an - f_out) * (i' - i) = 0
+    let absorbing_or_comp_finished = 1 - f_an - f_out
+    enf is_unchanged([i]) when absorbing_or_comp_finished
+
 
 # Enforce hasher state constraints
-ev hasher_state(main: [h[12]], b, f_abp, f_abs_node):
+ev hasher_state(main: [s[3], h[12], i]):
+    let f_mp = get_f_mp(s)
+    let f_mv = get_f_mv(s)
+    let f_mu = get_f_mu(s)
+    let f_abp = get_f_abp(s)
+
+    # Flag that is true when the performed operation includes absorbing the next node during Merkle
+    # path computation.
+    let f_absorb_node = f_mp + f_mv + f_mu
+
+    # b is the value of the bit which is discarded during shift by one bit to the right.
+    let b = i - 2 * i'
+
     # Enforce that when absorbing the next set of elements into the state during linear hash 
     # computation (i.e. f_abp = 1) the first 4 elements (the capacity portion) are carried over to 
     # the next row.
     enf f_abp * (h[j]' - h[j]) = 0 for j in 0..4
+
     # Enforce that when absorbing the next node during Merkle path computation 
     # (i.e. f_mp + f_mv + f_mu = 1), the result of the previous hash (h[4], ..., h[7]) are copied 
     # over either to (h[4]', ..., h[7]') or to (h[8]', ..., h[11]') depending on the value of b.
-    enf (f_mp + f_mv + f_mu) * ((1 - b) * (h[j + 4]' - h[j + 4]) + b * (h[j + 8]' - h[j + 4])) = 0
+    match enf:
+        is_unchanged(h[j + 4]) for j in 0..4 when !b & f_absorb_node
+        h[j + 8]' = h[j + 4] for j in 0..4 when b & f_absorb_node
 
-### Bitwise Chiplet Air Constraints ###############################################################
+
+### Hash Chiplet Air Constraints ##################################################################
 
 # Enforces the constraints on the hash chiplet, given the columns of the hash execution trace.
 ev hash_chiplet(main: [s[3], r, h[12], i]):
-
-    ## Instruction flags ##
-    let f_rpr = 1 - k0
-    let f_bp = k2 * s[0] * (1 - s[1]) * (1 - s[2])
-    let f_mp = k2 * s[0] * (1 - s[1]) * s[2]
-    let f_mv = k2 * s[0] * s[1] * (1 - s[2])
-    let f_mu = k2 * s[0] * s[1] * s[2]
-    let f_hout = k0 * (1 - s[0]) * (1 - s[1]) * (1 - s[2])
-    let f_sout = k0 * (1 - s[0]) * (1 - s[1]) * s[2]
-    let f_out = k0 * (1 - s[0]) * (1 - s[1])
-    let f_out_next = k1 * (1 - s[0]') * (1 - s[1]')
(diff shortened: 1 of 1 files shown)


Overcastan:
This PR has the same problem as #191, that is described in [this comment](https://github.com/0xPolygonMiden/air-script/pull/191#issuecomment-1466039654)

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,76 @@
...
+ev is_binary(main: [a]):
+    enf a^2 - a = 0
+
+### Bitwise Chiplet Air Constraints ###############################################################
+
+# Enforces the constraints on the hash chiplet, given the columns of the hash execution trace.
+ev hash_chiplet(main: [s[3], r, h[12], i]):
+
+    ## Instruction flags ##
+    let f_rpr = 1 - k0
```
tohrnii, line 22: Should we refactor this as:
`let f_rpr = !k0`

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,76 @@
...
+    let f_bp = k2 * s[0] * (1 - s[1]) * (1 - s[2])
```
tohrnii, line 23: 1. Should we rewrite this as:
`let f_bp = k2 & s[0] & !s[1] & !s[2]`

2. Should we declare an intermediate variable for `k2 & s[0]`?

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,76 @@
...
+    let f_mp = k2 * s[0] * (1 - s[1]) * s[2]
+    let f_mv = k2 * s[0] * s[1] * (1 - s[2])
+    let f_mu = k2 * s[0] * s[1] * s[2]
+    let f_hout = k0 * (1 - s[0]) * (1 - s[1]) * (1 - s[2])
+    let f_sout = k0 * (1 - s[0]) * (1 - s[1]) * s[2]
+    let f_out = k0 * (1 - s[0]) * (1 - s[1])
+    let f_out_next = k1 * (1 - s[0]') * (1 - s[1]')
+    let f_abp = k0 * s[0] * (1 - s[1]) * (1 - s[2])
```
tohrnii, line 31: Should we declare an intermediate variable for `k0 & s[0]`?

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,76 @@
...
+    ## Instruction flags ##
+    let f_rpr = 1 - k0
+    let f_bp = k2 * s[0] * (1 - s[1]) * (1 - s[2])
+    let f_mp = k2 * s[0] * (1 - s[1]) * s[2]
+    let f_mv = k2 * s[0] * s[1] * (1 - s[2])
+    let f_mu = k2 * s[0] * s[1] * s[2]
+    let f_hout = k0 * (1 - s[0]) * (1 - s[1]) * (1 - s[2])
+    let f_sout = k0 * (1 - s[0]) * (1 - s[1]) * s[2]
+    let f_out = k0 * (1 - s[0]) * (1 - s[1])
+    let f_out_next = k1 * (1 - s[0]') * (1 - s[1]')
```
tohrnii, line 30: Should we rewrite this as:
`let f_out_next = k1 & !s[0]' & !s[1]'`

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,76 @@
...
+    # Enforce that the value in i is copied over to the next row unless we are absorbing a new row 
+    # or the computation is finished.
+    enf (1 - f_an - f_out) * (i' - i) = 0
+
+    ## Hasher state constraints ##
+    # TODO: apply RPO constraints to the hasher state
+    # Enforce that when absorbing the next set of elements into the state during linear hash 
+    # computation (i.e. f_abp = 1) the first 4 elements (the capacity portion) are carried over to 
+    # the next row.
+    enf f_abp * (h[j]' - h[j]) for j in 0..4
```
tohrnii, line 69: This doesn't seem correct as there is no rhs?

tohrnii (changes requested):
Thank you @Overcastan. I left a few comments.

I am wondering if we should allow logical operators for in intermediate variables and for intermediate variables? For e.g., should the following be valid?:

```
let a = s0 & !s1
let b = !a
```

constraints/miden-vm/hash.air
Overcastan: Good catch! It should be equal 0. Thank you!

Overcastan:
Allowing logical operators in intermediate variables looks really useful, but won't this complicate the process of parsing and processing expressions?

grjte:
> Allowing logical operators in intermediate variables looks really useful, but won't this complicate the process of parsing and processing expressions?

I think this is a good idea and would be useful, but let's not do it now. Can we create an issue and put it in the "Nice to have" section of [this tracking issue](https://github.com/0xPolygonMiden/air-script/issues/162)?

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,76 @@
...
+    ## Row address constraint ##
+    # TODO: Apply row address constraint
+
+    ## Selector columns constraints ##
+    # Enforce that selector columns are binary.
+    enf is_binary([s_i]) for s_i in s
+    # Enforce that unless f_out = 1 or f_out' = 1, the values in columns s[1] and s[2] are copied 
+    # over to the nex row.
+    enf (s[1]' - s[1]) * (1 - f_out_next) * (1 - f_out) = 0
+    enf (s[2]' - s[2]) * (1 - f_out_next) * (1 - f_out) = 0
```
grjte: Since these are constraints rather than variables, can we make use of selectors to do the following?

```
enf s[1]' = s[1] when !f_out & !f_out_next
enf s[2]' = s[2] when !f_out & !f_out_next
```

In fact, it might be best to make an `is_copied` function, because we check for that a lot in this module and you could use it elsewhere as well


```
enf is_copied([s[1]]) when !f_out & !f_out_next
enf is_copied([s[2]]) when !f_out & !f_out_next
```

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,76 @@
...
+    enf s[0]' * (f_abp + f_mpa + f_mva + f_mua) = 0
+    # Enforce that no invalid combinations of flags are allowed.
+    enf k0 * (1 - s[0]) * s[1] = 0
+
+    ## Node index constraints ##
+    # b is the value of the bit which is discarded during shift by one bit to the right.
+    let b = i - 2 * i'
+    let f_an = f_mp + f_mv + f_mu + f_mpa + f_mva + f_mua
+    # Enforce that b is binary only when a new node is absorbed into the hasher state.
+    enf f_an * (b^2 - b) = 0
```
grjte: We can use `is_binary` here, right?

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,76 @@
...
+    # Enforce that when a computation is finished i = 0.
+    enf f_out * i = 0
+    # Enforce that the value in i is copied over to the next row unless we are absorbing a new row 
+    # or the computation is finished.
+    enf (1 - f_an - f_out) * (i' - i) = 0
```
grjte: We could use `is_copied` and a selector here, which I think improves readability:

```
enf is_copied([i]) when 1 - f_an - f_out
```

constraints/miden-vm/hash.air
grjte: Note for the future - it might be nice to be able to combine constraint comprehension with selectors, e.g.

```
enf is_copied([h[j]]) for j in 0..4 when f_abp
```

What do you both think? (Or will this already be possible without extra work?)

tohrnii:
> I think this is a good idea and would be useful, but let's not do it now. Can we create an issue and put it in the "Nice to have" section of this tracking issue?

Added #217

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,76 @@
...
+    ## Hasher state constraints ##
+    # TODO: apply RPO constraints to the hasher state
+    # Enforce that when absorbing the next set of elements into the state during linear hash 
+    # computation (i.e. f_abp = 1) the first 4 elements (the capacity portion) are carried over to 
+    # the next row.
+    enf f_abp * (h[j]' - h[j]) for j in 0..4
+    # Enforce that when absorbing the next node during Merkle path computation 
+    # (i.e. f_mp + f_mv + f_mu = 1), the result of the previous hash (h[4], ..., h[7]) are copied 
+    # over either to (h[4]', ..., h[7]') or to (h[8]', ..., h[11]') depending on the value of b.
+    enf (f_mp + f_mv + f_mu) * ((1 - b) * (h[j + 4]' - h[j + 4]) + b * (h[j + 8]' - h[j + 4])) = 0
```
grjte: A few points here:
- I think using a selector for the absorption flag would improve readability (`when f_mp + f_mv + f_mu`)
- I think using `is_copied` would improve readability (e.g. `is_copied([h[j+4]])`)
- I think using selectors for `b` and `!b` would improve readability. This would mean splitting it into 2 constraints using a `match enf`, one of which is enforced `when !b & (f_mp + f_mv + f_mu)` and the other of which is enforced `when b & (f_mp + f_mv + f_mu)`.
- It needs to be applied to the entire digest, but it looks like it's only being applied to the first element of the digest right now

constraints/miden-vm/hash.air
tohrnii: Currently we only allow identifiers as selectors and not expressions like `1 - f_an - f_out`. Do you think we should allow expressions as selectors? (To me it doesn't look very clean but not a strong opinion). Maybe we could define another intermediate variable for `1 - f_an - f_out` and use that.

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,76 @@
+mod HashAir
```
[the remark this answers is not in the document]
Overcastan: My thought was that this file contains constraints for general hash operations, that's why I thought that it's possible to name it just `HashAir`, but `HashChipletAir` better displays the essence of the entire script, so I agree with you, we should rename it to `HashChipletAir`.

constraints/miden-vm/hash.air
tohrnii: @grjte I think Bobbin suggested to write this as

```
ev foo(main: h[4]):
    enf is_copied(h) for h in h
enf foo(h[4]) when f_abp
```
instead.


Also, should we write this as:
```
enf is_copied([h]) for h in h[0..4]
```

constraints/miden-vm/hash.air
Overcastan: As I understood, unfortunately, no. We should pass only columns in `is_binary` evaluator, and b is a variable. But it would be great if I got it wrong.

constraints/miden-vm/hash.air
Overcastan: I'm not sure I understand what do you mean by digest in this context. Can you explain in more detail?

constraints/miden-vm/hash.air
Overcastan: I'm not sure why should we create intermediate variables for the `k0 * (1 - s[0])` and similar expressions. I think in this case readability will be decreased since every row is described with set of selector values (e.g. `# this flag is set to 1 when flags are (i, j, k) and ... `) and in this case it will be more difficult to keep track of the value of the first flag

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,156 @@
...
+
+
+# Enforces that value in column is copied over to the nex row.
+ev is_copied(main: [column]):
+    ev column' = column
+
+
+# Enforce selector columns constraints
+ev selector_columns(main: [s[3]]):
+    # Set to 1 when either f_hout = 1 or f_sout = 1 in the current row.
```
tohrnii, line 30: `f_hout` and `f_sout` are undefined.

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,156 @@
...
+    let f_out = k0 * binary_not(s[0]) * binary_not(s[1])
+    # Set to 1 when either f_hout = 1 or f_sout = 1 in the next row.
```
tohrnii, line 32: `f_hout` and `f_sout` are undefined.

nit: I'd add a blank line. I think by convention we should have a blank line between a constraint and if the next line is a comment. Wdyt?

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,156 @@
...
+# Enforce node index constraints
+ev node_index(main: [s[3], i]):
+    # Set to 1 when selector flags are (1,0,1) on rows which are multiples of 8.
+    let f_mp = k2 * s[0] * binary_not(s[1]) * s[2]
+    # Set to 1 when selector flags are (1,1,0) on rows which are multiples of 8.
+    let f_mv = k2 * s[0] * s[1] * binary_not(s[2])
+    # Set to 1 when selector flags are (1,1,1) on rows which are multiples of 8.
+    let f_mu = k2 * s[0] * s[1] * s[2]
+
+    # Set to 1 when either f_hout = 1 or f_sout = 1 in the current row.
```
tohrnii, line 71: `f_hout` and `f_sout` are undefined.

tohrnii (changes requested):
Thank you @Overcastan, I just left a few comments. There is a recent change to logical operators and now they can be used in intermediate variables. Can you please refactor the code to incorporate that?

So, for example, something like:
```
let f_mpa = k0 * s[0] * binary_not(s[1]) * s[2]
```
can be re-written as
```
let f_mpa = k0 & s[0] & !s[1] & s[2]
```
Can you please also check in the memory chiplets PR if these changes need to be made?

constraints/miden-vm/hash.air
grjte: > enf is_copied([h]) for h in h[0..4]
Yes, thank you.

> ev foo(main: h[4]):
>    enf is_copied(h) for h in h
> enf foo(h[4]) when f_abp

Yep, this is fine

constraints/miden-vm/hash.air
grjte: The digest is the 4 field elements that make up the hash

constraints/miden-vm/hash.air
grjte: Yes, I think we should always do this. I left a similar comment previously

constraints/miden-vm/hash.air
grjte: > Maybe we could define another intermediate variable for 1 - f_an - f_out and use that.

yep, I think this would be better

constraints/miden-vm/hash.air
grjte: Hmm you're right. I was thinking we'd use an is_binary function (rather than an evaluator), but then we still have to check something like `is_binary(b) = 0` which is probably less clear.

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,156 @@
...
+    # (i.e. f_mp + f_mv + f_mu = 1), the result of the previous hash (h[4], ..., h[7]) are copied 
+    # over either to (h[4]', ..., h[7]') or to (h[8]', ..., h[11]') depending on the value of b.
+    # TODO: once the support for combination of constraint comprehension and selectors is 
+    # implemented, update this constraint as:
+    # ```
+    # match enf:
+    #     is_copied(h[j + 4]) for j in 0..4 when !b & f_abs_node
+    #     h[j + 8]' = h[j + 4] for j in 0..4 when b & f_abs_node
+    # ```
+    enf f_abs_node * ((1 - b) * is_copied(h[j + 4]) + b * (h[j + 8]' - h[j + 4])) = 0 for j in 0..4
```
grjte, line 131: I don't think we need to wait, since this isn't going into the tests. I would just change this now:

 ```
 match enf:
     is_copied(h[j + 4]) for j in 0..4 when !b & f_abs_node
     h[j + 8]' = h[j + 4] for j in 0..4 when b & f_abs_node
 ```

constraints/miden-vm/hash.air
grjte: I think you're imagining this in a different way from what we intended. If you were to follow this suggestion, it would be given a meaningful name, so we would have something like the following:

```
let last_row_and_s0 = k0 * s[0]
let first_row_and_s0 = k2 * s[0]
```

But looking at this, I wonder if the most readable thing to do would actually be to rename the periodic columns like:

```
periodic_columns:
    cycle_row_7: [0, 0, 0, 0, 0, 0, 0, 1]
    cycle_row_6: [0, 0, 0, 0, 0, 0, 1, 0]
    cycle_row_0: [1, 0, 0, 0, 0, 0, 0, 0]
```

grjte (changes requested):
Looks good, thanks @Overcastan. In addition to the things @tohrnii mentioned, I left a few things inline.

Also more generally - the comments on the flags tell when they're set, but it would be really helpful to include the semantic meaning as well.

For example:

```
    # Set to 1 when selector flags are (1,0,0) on rows which are 1 less than a multiple of 8.
    # Indicates the hash operation is absorption
    let f_abp = k0 * s[0] * binary_not(s[1]) * binary_not(s[2])
```

If you're not sure of the semantic meaning of the flags, feel free to message me

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,156 @@
...
+### Helper functions/evaluators ###################################################################
+
+# Enforces that column must be binary.
+ev is_binary(main: [a]):
+    enf a^2 - a = 0
+
+
+# Returns binary negation of the value.
+fn binary_not(value: scalar) -> scalar:
+    return 1 - value
```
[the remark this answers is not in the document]
Overcastan: I agree, I put function on the top in the Memory Chiplet for the same reason. Not sure about additional headers though. On the one hand functions and evaluators are very similar, so this separation is not needed. But on the other hand if we have a lot of functions and evaluators, it will be very convenient to have this separation. I'll make this separation for now, let's see how it will look like.

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,243 @@
...
+    # of 8. This is flag of the instruction that returns the result of the currently running 
+    # computation.
+    # 2. Flag f_sout = cycle_row_7 * binary_not(s[0]) * binary_not(s[1]) * s[2], which is set to 1 
+    # when selector flags are (0,0,1) on rows which are 1 less than a multiple of 8. This is flag 
+    # of the instruction that returns the whole hasher state.
+    #
+    # Flag f_out is set to 1 when either f_hout = 1 or f_sout = 1 in the current row.
+    let f_out = cycle_row_7 * binary_not(s[0]) * binary_not(s[1])
+
+    # Falg f_out_nextet is to 1 when either f_hout = 1 or f_sout = 1 in the next row.
```
grjte, line 72: nit: There are typos here. It should be "Flag f_out_next is set to 1..."

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,243 @@
...
+    # 1. Flag f_hout = cycle_row_7 * binary_not(s[0]) * binary_not(s[1]) * binary_not(s[2]), 
+    # which is set to 1 when selector flags are (0,0,0) on rows which are 1 less than a multiple 
+    # of 8. This is flag of the instruction that returns the result of the currently running 
+    # computation.
+    # 2. Flag f_sout = cycle_row_7 * binary_not(s[0]) * binary_not(s[1]) * s[2], which is set to 1 
+    # when selector flags are (0,0,1) on rows which are 1 less than a multiple of 8. This is flag 
+    # of the instruction that returns the whole hasher state.
+    #
+    # Flag f_out is set to 1 when either f_hout = 1 or f_sout = 1 in the current row.
+    let f_out = cycle_row_7 * binary_not(s[0]) * binary_not(s[1])
```
grjte, line 70: Should we make a helper function for `f_out` as well? Then this comment can go above the function definition and this evaluator and the ones below can be much smaller

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,243 @@
...
+    # is flag of the instruction that absorbs the next Merkle path node into the hasher state 
+    # during Merkle path verification for the "old" node value during Merkle root update 
+    # computation.
+    let f_mva = get_f_mva(s)
+
+    # Set to 1 when selector flags are (1,1,1) on rows which are 1 less than a multiple of 8. This 
+    # is flag of the instruction that absorbs the next Merkle path node into the hasher state 
+    # during Merkle path verification for the "new" node value during Merkle root update 
+    # computation.
+    let f_mua = get_f_mua(s)
```
grjte, line 160: Similar to my previous comment - let's put all of these comments above the function definitions and remove them from here. That way this can be just 6 lines and this evaluator will be much easier to read. Let's also do that anywhere else any of these `get_f_*` functions are called

grjte (changes requested):
Looks good, thank you! Just a couple of final comments about comments for the functions, then it will be ready to merge

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,215 @@
...
+ev hasher_state(main: [s[3], h[12], i]):
+    let f_mp = get_f_mp(s)
+    let f_mv = get_f_mv(s)
+    let f_mu = get_f_mu(s)
+
+    # Flag that is true when the performed operation includes absorbing the next node during Merkle
+    # path computation.
+    let f_absorb_node = f_mp + f_mv + f_mu
+
+    let f_abp = get_f_abp(s)
```
grjte, line 177: nit: I would prefer to move this up a few rows so it's declared with the other flags at the top of the evaluator

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,215 @@
...
+
+# We can define two flags:
+# 1. Flag f_hout = cycle_row_7 * binary_not(s[0]) * binary_not(s[1]) * binary_not(s[2]), 
+# which is set to 1 when selector flags are (0,0,0) on rows which are 1 less than a multiple 
+# of 8. This is flag of the instruction that returns the result of the currently running 
+# computation.
+# 2. Flag f_sout = cycle_row_7 * binary_not(s[0]) * binary_not(s[1]) * s[2], which is set to 1 
+# when selector flags are (0,0,1) on rows which are 1 less than a multiple of 8. This is flag 
+# of the instruction that returns the whole hasher state.
+#
```
grjte, line 90: Since we already have this above the `get_f_out` function, I think we can remove it and just keep the line below

grjte (approved):
Looks great, thank you! I've left 2 final nits inline that I think it would be great to fix before merging, but they're not blocking. I'll leave it up to you to merge and you can make those updates first if you agree with them

constraints/miden-vm/hash.air
```diff
@@ -0,0 +1,205 @@
...
+
+# Returns binary negation of the value.
+fn binary_not(value: scalar) -> scalar:
+    return 1 - value
+
+
+# Set to 1 when selector flags are (1,0,1) on rows which are multiples of 8. This is flag of 
+# the instruction that initiates Merkle path verification computation.
+fn get_f_mp(s: vector[3]) -> scalar:
+    return cycle_row_0 * s[0] * binary_not(s[1]) * s[2]
```
tohrnii, line 21: nit: should we use logical operators here and other flag calculations as well?

tohrnii (approved):
Thank you @Overcastan, looks good to me. I just left one nit to write flag computation in terms of logical operators if you agree with the change.

constraints/miden-vm/hash.air
Overcastan: Yes, I think we should do that. Thank you!

PR #216, opened 2023-03-13
Refactor IntegrityStmt

tohrnii:
This PR refactors the `IntegrityStmt` to make it easy to incorporate selectors.


ir/src/constraint_builder/integrity_constraints/mod.rs
```diff
@@ -4,9 +4,10 @@ use super::{
     ListFoldingValueType, NamedTraceAccess, SemanticError, Symbol, SymbolType, Variable,
     VariableType, VectorAccess, CURRENT_ROW,
 };
+// TODO: Move this to air-script-core
+use parser::ast::ConstraintType;
```
grjte, line 8: Does it make sense to just put this in `air-script-core` now?

parser/src/parser/tests/trace_columns.rs
```diff
@@ -1,3 +1,5 @@
+use crate::ast::ConstraintType;
+
```
grjte, line 2: nit: Let's remove the extra line here

ir/src/constraint_builder/integrity_constraints/mod.rs
tohrnii: Sorry, I didn't realize then that we will have to move all it's dependencies to `air-script-core` (which would essentially be almost everything in parser). So, I have just imported from `super::ast` instead. Please let me know if you disagree with the change.

tohrnii:
Sorry, I merged the PRs in wrong order. The selector parser PR was merged into this branch instead.

grjte (approved):
Looks good, thanks!

ir/src/constraint_builder/integrity_constraints/mod.rs
grjte: Makes sense!

Issue #125, opened 2023-01-26
Add support for selectors


tohrnii:
Frequently, we want to apply some set of constraints only if some condition is true. For example, for Rescue Prime constraints, we usually want to apply them only on the first 7 steps of an 8 step cycle. This, of course, could be baked into the constraints themselves, but that would make them much less readable and modular.

So, some syntax which will help with applying selectors could be very helpful. Here is how this could work:

### Single selector

First, if we have just a single selector, we could do something like this:

```
// define a cyclic column consisting of 7 ones and 1 zero.
cycl k0 = [1, 1, 1, 1, 1, 1, 1, 0]

ev foo(main: [a, b, c]):
  enf bar([a, b]) when k0

ev bar(main: [a, b]):
  enf a' = a + b
```

The above specifies that to all constraints returned from bar evaluator would be multiplied by k0. Specifically, the final constraint would be:

$$
k_0 \cdot (a' - a - b) = 0
$$

We could also have the compiler check whether $k_0$ is binary and output either an error or a warning if it isn't.

### Multi selector

While we should be able to describe everything using single selectors, frequently it could be much more convenient to specify that one out of several constraints should apply in a given situation. For this, we could use something like that:

```
ev foo(main: [a, b, c]):
  // make sure values in a are either ones or zeros
  enf a^2 = a

  match enf:
    bar_add([b, c]) when a
    bar_mul([b, c]) when 1 - a

ev bar_add(main: [a, b]):
  enf a' = a + b

ev bar_mul(main: [a, b]):
  enf a' = a * b
```

The above would translate into the following constraint:

$$
a \cdot (b' - b - c) + (1 - a) \cdot (b' - b \cdot c) = 0
$$

We could also have the compiler check whether all `when` conditions are mutually exclusive and output an error or a warning if they are not.

The above methodology would also work if evaluators return different number of constraints. For example:

```
ev foo(main: [a, b, c]):
  // make sure values in a are either ones or zeros
  enf a^2 = a

  match enf:
    bar_add([b, c]) when a
    bar_mul([b, c]) when 1 - a

ev bar_add(main: [a, b]):
  enf a' = a + b

ev bar_mul(main: [a, b]):
  enf a' = a * b
  enf b' = a
```
The above would translate to two constraints:

$$
a \cdot (b' - b - c) + (1 - a) \cdot (b' - b \cdot c) = 0
$$

$$
(1 - a) \cdot (c' - b) = 0
$$

[...]

PR #218, opened 2023-03-13
Add support for selectors

tohrnii:
Partly addressing #125

### DIFF
--- a/parser/src/lexer/mod.rs
+++ b/parser/src/lexer/mod.rs
@@ -121,6 +121,9 @@ pub enum Token {
     #[token("enf")]
     Enf,
 
+    #[token("match")]
+    Match,
+
     // OPERATORS
     // --------------------------------------------------------------------------------------------
     /// Asserts LHS of the expression is equal to RHS of the expression.
@@ -138,6 +141,20 @@ pub enum Token {
 
     #[token("^")]
     Exp,
+
+    // SELECTOR KEYWORDS
+    // --------------------------------------------------------------------------------------------
+    #[token("&")]
+    And,
+
+    #[token("|")]
+    Or,
+
+    #[token("!")]
+    Not,
+
+    #[token("when")]
+    When,
 
     // DELIMITERS
     // --------------------------------------------------------------------------------------------
--- a/parser/src/parser/grammar.lalrpop
+++ b/parser/src/parser/grammar.lalrpop
@@ -382,8 +382,8 @@ IntegrityAtom: Expression = {
     "$" <ident: Identifier> <idx: Index> => Expression::Rand(ident, idx),
     <n: Num_u64> => Expression::Const(n),
     <ident: Identifier> => Expression::Elem(ident),
-    "!" <ident: Identifier> =>
-        Expression::Sub(Box::new(Expression::Const(1)), Box::new(Expression::Elem(ident))),
+    "!" <expr: IntegrityAtom> =>
+        Expression::Sub(Box::new(Expression::Const(1)), Box::new(expr)),
     <vector_access: VectorAccess> => Expression::VectorAccess(vector_access),
     <matrix_access: MatrixAccess> => Expression::MatrixAccess(matrix_access),
     <trace_access: NamedTraceAccessWithOffset> => Expression::NamedTraceAccess(trace_access),
--- a/parser/src/parser/grammar.lalrpop
+++ b/parser/src/parser/grammar.lalrpop
@@ -145,7 +145,8 @@ RandValuesDecl: RandomValues = {
             )
         }
     ),
-    <name: Identifier> ":" "[" <rand_vec: CommaElems<RandElem>> "]" => RandomValues::new(name, rand_vec.iter().map(|v| v.size()).sum(), rand_vec)
+    <name: Identifier> ":" "[" <rand_vec: CommaElems<RandElem>> "]" =>
+        RandomValues::new(name, rand_vec.iter().map(|v| v.size()).sum(), rand_vec)
 }
 
 RandElem: RandBinding = {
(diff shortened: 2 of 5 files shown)


parser/src/parser/grammar.lalrpop
```diff
@@ -283,16 +285,32 @@ IntegrityStmts: Vec<IntegrityStmt> = {
...
+    <integrity_stmt_with_selector: IntegrityStmtWithSelector> =>
+        integrity_stmt_with_selector,
+}
+
+IntegrityStmtWithSelector: IntegrityStmt = {
+    <lhs: IntegrityExpr> "=" <rhs: IntegrityExpr> "when" <selectors: SelectorExpr> =>
+        IntegrityStmt::Constraint(ConstraintType::Inline(IntegrityConstraint::new(lhs, rhs)),
+        Some(selectors)),
+    <evaluator_fn_call: EvaluatorFunctionCall> "when" <selectors: SelectorExpr> =>
+        IntegrityStmt::Constraint(ConstraintType::Evaluator(evaluator_fn_call), Some(selectors)),
```
grjte, line 313: Should these go into the "INTEGRITY CONSTRAINT EXPRESSIONS" section?

parser/src/parser/grammar.lalrpop
```diff
@@ -283,16 +285,32 @@ IntegrityStmts: Vec<IntegrityStmt> = {
...
+    <lhs: IntegrityExpr> "=" <rhs: IntegrityExpr> =>
         IntegrityStmt::Constraint(ConstraintType::Inline(IntegrityConstraint::new(lhs, rhs)), None),
-    "enf" <evaluator_fn_call: EvaluatorFunctionCall> =>
+    <evaluator_fn_call: EvaluatorFunctionCall> =>
         IntegrityStmt::Constraint(ConstraintType::Evaluator(evaluator_fn_call), None),
+    <integrity_stmt_with_selector: IntegrityStmtWithSelector> =>
+        integrity_stmt_with_selector,
+}
+
+IntegrityStmtWithSelector: IntegrityStmt = {
```
grjte, line 308: nit: This could probably be `IntegrityConstraintWithSelector`, since there are no variables allowed here

parser/src/parser/grammar.lalrpop
```diff
@@ -367,6 +385,29 @@ CommaElems<T>: Vec<T> = {
...
+        )
+    },
+    Selector
+}
+
+Selector: Expression = {
+    "(" <SelectorExpr> ")",
+    <selector: Identifier> => Expression::Elem(selector),
+    "!" <selector: Identifier> =>
+        Expression::Sub(Box::new(Expression::Const(1)), Box::new(Expression::Elem(selector))),
```
grjte, line 408: note: we don't really have a way of checking that this element is binary, but with this transformation we lose the knowledge that we're expecting a binary value. I wonder if we should do something else here. This is true for all of the selectors, really. We should probably enforce that we expect them to be binary, but we won't have this information in the IR. 

Maybe the best thing to do would be to create a new type of expression with `&`, `|`, `!` and let the IR handle rather than the parser.

@bobbinth do you have thoughts on this?

grjte (changes requested):
Thanks @tohrnii! I left a few small things inline and one bigger question

parser/src/lexer/mod.rs
```diff
@@ -139,6 +139,23 @@ pub enum Token {
...
+
+    #[token("!")]
+    Not,
+
+    #[token("when")]
+    When,
+
+    #[token("match")]
+    Match,
+
```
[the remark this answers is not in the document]
grjte: To be clear, what I mean by the above is the following:
- maybe the "match" token should be in the "general" section
- maybe we should have a separate discussion about making "match" more useful

parser/src/parser/grammar.lalrpop
grjte: Or I guess another option in the IR would be to enforce that every element (or trace reference, etc) used inside one of these expressions as a selector is a binary element, that none of these refer to random values, etc etc. Alternatively, we could just not enforce anything.

parser/src/lexer/mod.rs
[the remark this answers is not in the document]
tohrnii: Agreed, I think what you described seems more useful.

parser/src/parser/grammar.lalrpop
tohrnii: I was thinking we don't enforce this. I'm not sure we can enforce this at the IR level either? Maybe in the future we could throw a warning in the future if we see these constraints are defined without enforcing a constraint that the selectors used are binary?

grjte: Yeah, it's easier not to enforce this. Let's leave it for now.

grjte: But maybe we should open an issue for discussion?

tohrnii: Added #219.

parser/src/parser/grammar.lalrpop
tohrnii: I'm not sure how to categorize this. One option is to categorize these based on what they return. So since this returns IntegrityStmt, this should maybe stay here? Wdyt?

parser/src/lexer/mod.rs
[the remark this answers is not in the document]
grjte: Can you open an issue for us to discuss it?

parser/src/parser/grammar.lalrpop
grjte: That's fine. It's logical, and I don't feel strongly

grjte (approved):
Looks good, thank you!

PR #222, opened 2023-03-15
refactor(ir): move constraints and graph to ConstraintBuilder

tohrnii:
This PR moves the graph and constraints to `ConstraintBuilder`.

### DIFF
--- a/ir/src/constraints/graph.rs
+++ b/ir/src/constraints/graph.rs
@@ -98,7 +98,7 @@ impl AlgebraicGraph {
     // --- PUBLIC MUTATORS ------------------------------------------------------------------------
     /// Insert the operation and return its node index. If an identical node already exists, return
     /// that index instead.
-    pub(super) fn insert_op(&mut self, op: Operation) -> NodeIndex {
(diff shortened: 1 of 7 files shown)


ir/src/constraints/graph.rs
```diff
@@ -98,7 +98,7 @@ impl AlgebraicGraph {
     // --- PUBLIC MUTATORS ------------------------------------------------------------------------
     /// Insert the operation and return its node index. If an identical node already exists, return
     /// that index instead.
-    pub(super) fn insert_op(&mut self, op: Operation) -> NodeIndex {
+    pub(crate) fn insert_op(&mut self, op: Operation) -> NodeIndex {
```
grjte, line 101: nit: I've been wondering about changing this to `insert_node` or `insert_graph_node` as well. Wdyt?

grjte (approved):
Looks good, thank you! Just one optional thought inline

ir/src/constraints/graph.rs
tohrnii: Changed it to `insert_node` since it seems like `graph` is already implied here.

PR #223, opened 2023-03-15
feat(parser): add support for functions

tohrnii:


parser/src/ast/function.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum ArgumentType {
+    Scalar,
+    Vector(u64),
+    Matrix(u64, u64),
+}
+
+pub type ReturnType = ArgumentType;
```
grjte, line 83: I'm not sure it's useful to have this distinction, and I think it actually confuses things a bit, because it leads to wondering what the difference is between argument types and return types, when actually they're the same. I would prefer to just call it something like `ValueType` instead and use this same type everywhere.

parser/src/ast/mod.rs
```diff
@@ -52,10 +55,11 @@ pub enum SourceSection {
     AirDef(Identifier),
     Constant(Constant),
     Trace(Vec<Vec<TraceBinding>>),
+    EvaluatorFunction(EvaluatorFunction),
+    Function(Function),
```
grjte, line 59: The docs above this enum need to be updated, and the order in the docs should be kept consistent with the order of enum variants. Also, I wonder if we should alphabetize things like this in general, now that it's gotten more complicated, rather than trying to follow a mental ordering of how we like to define the AirScript files.

grjte, line 59: The docs above this enum need to be updated, and the order in the docs should be kept consistent with the order of enum variants. Also, I wonder if we should alphabetize things like this in general, now that it's gotten more complicated, rather than trying to follow a mental ordering of how we like to define the AirScript files.

parser/src/parser/grammar.lalrpop
```diff
@@ -403,6 +448,8 @@ IntegrityAtom: Expression = {
     <trace_access: TraceBindingAccessWithOffset> => Expression::TraceBindingAccess(trace_access),
     <list_folding_type: ListFoldingType<IntegrityExpr>> =>
         Expression::ListFolding(list_folding_type),
+    "$" <function_name: Identifier> "(" <args: CommaElems<IntegrityVariableType>> ")" =>
+        Expression::FunctionCall(function_name, args),
```
grjte, line 452: I don't think we want to use `IntegrityVariableType` for the args here. If we were to adjust the nomenclature of everything to some form of `Binding` and `BindingAccess` (like we've done for `TraceBinding` and `TraceBindingAccess`), then these args would be an array of `BindingAccess`. That means we could use `a[1]` as an argument. However, this parsing rule expects something like `[a]` instead, which I don't think matches what we want.

Hopefully fixing this will help resolve the other issue so that the `$` can be removed as well.

Also, I am finding the name `VariableType` increasingly confusing, especially in this PR. Maybe we should rename this to `VariableValue`. What do you think?

grjte (changes requested):
Thanks @tohrnii! I left some initial feedback. I'm finding some of the terms confusing, since we now have some things called `Type` that really correspond to values (e.g. `VariableType`) and other things called `Type` that correspond to types (like the new `ArgumentType`). I think this gets even more confusing in the grammar, since things are broken down into smaller items. For example, I think `IntegrityVariableType` is a misnomer. I think this has led to an error in the grammar definition. I recommended a different way of thinking about the grammar, but I also think we should address some of the naming issues and also consider trying to introduce the use of the `Binding`/`BindingAccess` terminology more consistently (e.g. maybe rename `Variable` to `VariableBinding` and have a separate `VariableBindingAccess` type for when we actually use variables

parser/src/ast/mod.rs
tohrnii: I think maybe keeping it in the same order as how it should ideally appear in the AirScript file seems better since I don't think the list of sections is going to increase much going forward.

PR #226, opened 2023-03-20
feat(parser): add support for slices

tohrnii:


tohrnii:
Support for slices has already been added.

PR #227, opened 2023-03-20
add support for evaluator functions to the IR

grjte:
Update: this PR is now ready for review and merge. Everything below is left for context.

This PR is not intended for merge. It's a sketch of an approach we could take for dealing with evaluators in the IR.

## Approach

Roughly, the approach is as follows:

- use the `ConstraintBuilder` more generally for anything that has a set of constraints
- add a new `SymbolType` `Parameter(ColumnGroup)`to the symbol table for referring to parameters
- add a new `Value` type `Parameter(TraceParameterAccess)` for referring to parameter accesses and use in leaf nodes in the graph
- process evaluators at the topmost level (when processing the AST), creating a `ConstraintBuilder` and then consuming it with `into_evaluator` to create an `Evaluator` which contains a graph, a list of nodes which are parameter references, a list of nodes which are constraint roots, and a copy of its own parameters for matching against arguments when the Evaluator is called
- collect all processed evaluators and provide them to each future `ConstraintBuilder`
- finally, process the integrity & boundary constraints, then consume the `ConstraintBuilder` with `into_air`

I think we could improve on the way the evaluators are done, but the main idea was to build a graph and then refer to these graphs when the evaluators are called

## Issues

The biggest issue that came up is that we need to refer to the trace in many different ways, and it's not managed very well. This is a separate issue from adding evaluators to the IR, but it should probably be dealt with before this IR change is made, and it should be sketched out based on the needs of this PR. I have made some attempts to clean this up, with the renaming of `NamedTraceAccess` to `TraceBindingAccess` and its generalization and the change of `TraceCols` to `TraceBinding`, but I'm not fully happy with it.

Notes: 
1. there was an error previously with the use of `NamedTraceAccess`. We cannot assume that these indicate trace references of width 1, so generalizing to a `TraceBindingAccess` with a size (width) is one way to force better error checking.
2. The type of the args in the `EvaluatorFunctionCall` should be different from the type of the parameters in the `EvaluatorFunction` definition, and I think it makes sense for them to be trace bindings which we can check against the symbol table and convert into absolutely indexed trace references (as I did here)


grjte:
Maybe we could simplify all of the different accesses that are being used by having a general `BindingAccess` type for referencing named bindings (trace, constants, variables). It would have the `binding` name, column offset, row offset, and size, which could be the full binding, a single index, or a slice. In addition to being useful where I've used `TraceBindingAccess`, it could then replace our `Elem`, `VectorAccess`, and `MatrixAccess`.

This might make error handling easier / more correct and maybe clean things up. On the other hand, maybe it would introduce more confusion. Curious what you think @bobbinth @tohrnii

grjte:
> Maybe we could simplify all of the different accesses that are being used by having a general BindingAccess type for referencing named bindings (trace, constants, variables). It would have the binding name, column offset, row offset, and size, which could be the full binding, a single index, or a slice. In addition to being useful where I've used TraceBindingAccess, it could then replace our Elem, VectorAccess, and MatrixAccess.

If we did this, we could also unify the trace handling so that instead of parsing it sometimes as `TraceBindingAccess` (when there's a row offset) and sometimes as `Elem`/`VectorAccess`/`MatrixAccess` (when there isn't), it would always be parsed as a `BindingAccess`, and the IR would only deal with this case in one place instead of in two places

bobbinth:
I've spent some time looking through this, and I think I understand the approach - but to make sure, here is my brief summary:

- We instantiate many `ConstraintBuilder`'s (one per evaluator). Their task is to "compile" evaluators into a constraint graph plus some additional data. The result of this compilation is the `Evaluator` struct.
- We then instantiate the final `ConstraintBuilder` with all compiled evaluators. This constraint builder processes integrity statements, and when it sees a call to an evaluator, to find the relevant evaluator and merges its constraint graph into its own constraint graph (I think this last part is still to be implemented).
  - During this merging process, it needs to map trace access from the evaluator graph to trace accesses against the "top level" trace definition, and I think this presents some challenges.

Assuming the above is directionally correct, here are some questions comments:

First, why do we need `param` and `param_nodes` members in the `Evaluator` struct? Once we have the graph, all trace references are `IndexedTraceAccess`'s, right? So, when we merge evaluator subgraph into the main graph, all we need to provide is a map of how trace indexes in the main graph map to trace indexes in the evaluator graph.

For example, say we have something like this:
```
trace_columns:
  main: [a, b, c, d]

integrity_constraints:
  enf foo([b, d])

ev foo(main: [a, b]):
  enf a^2 = a
  enf b^2 = b
```

Once the evaluator is compiled, `a` and `b` references are translated to indexes $0$ and $1$, right? So, when the evaluator is called, all we need to provide is a map that says $0 \rightarrow 1$ and $1 \rightarrow 3$. I believe this information can be figured out at the call site (i.e., at `enf foo([b, d])`). Once we have this mapping, we can just scan through the nodes of the evaluator's graph and update any `IndexedTraceAccess` as needed.

Second, when we instantiate a `ConstraintBuilder` for an evaluator ([here](https://github.com/0xPolygonMiden/air-script/blob/df7d2609fd596ef33541c50644a51c71c409f98f/ir/src/lib.rs#L121-L122)), why do we pass a clone of symbol table to it? Wouldn't it make sense to augment the passed-in symbol table to make it relevant for the specific evaluator? Specifically, I'm thinking we might want to do the following for the evaluator-specific symbol table:
- Clear variables (though, there probably shouldn't be any to begin with).
- Clear all `TraceColumns` symbols.
- Add evaluator-specific `TraceColumns` to the symbol table based on `trace_params` of the evaluator.

This should ensure that the evaluator cannot access variables and trace columns of the parent context, and would also define trace columns for this specific evaluator.

Third, it feels like `ConstraintBuilder` is now used for too many things (i.e., evaluators don't care about boundary constraints, and yet every `ConstraintBuilder` contains info about `constrainted_boundaries` etc. I wonder if there is a cleaner way to separate these. I don't have great suggestions here, but perhaps one thing I'd do is to get rid of `into_evaluator()` function and change `process_evaluator()` into something like:
```Rust
pub(crate) fn process_evaluator(
    mut self,
    integrity_stmts: Vec<IntegrityStmt>,
) -> Result<Evaluator, SemanticError>
```
And the `ConstraintBuilder` context for this function would be set to `IntegrityConstraint` as I don't believe we need to track anything special for evaluators.

grjte:
## First point

> First, why do we need param and param_nodes members in the Evaluator struct? Once we have the graph, all trace references are IndexedTraceAccess's, right? So, when we merge evaluator subgraph into the main graph, all we need to provide is a map of how trace indexes in the main graph map to trace indexes in the evaluator graph.

We need `params` so we can validate the arguments against the parameters when we process an `EvaluatorFnCall`.

For `param_nodes` - we don't need to do this. Instead, we could walk through all of the nodes in the graph and replace each one we come across that needs to be updated.

## Second point

> Second, when we instantiate a ConstraintBuilder for an evaluator ([here](https://github.com/0xPolygonMiden/air-script/blob/df7d2609fd596ef33541c50644a51c71c409f98f/ir/src/lib.rs#L121-L122)), why do we pass a clone of symbol table to it? Wouldn't it make sense to augment the passed-in symbol table to make it relevant for the specific evaluator?

We are passing a clone because there may be many evaluators in a row which use the same "starting" symbol table before adding their trace column information to it.

> Clear variables (though, there probably shouldn't be any to begin with).

There are no variables in the symbol table that is used to initialize the `ConstraintBuilder` for processing an evaluator, since they are only added when processing sections for `boundary_constraints`, `integrity_constraints` or `evaluators`

> Clear all TraceColumns symbols.

I had been wondering about this too. I think we should update the current handling so that only `Constants` and `PeriodicColumns` are added to the base symbol table before evaluator processing happens. `TraceColumns` and `PublicInputs` could be added after all evaluators are processed, before processing the constraint sections.

> Add evaluator-specific TraceColumns to the symbol table based on trace_params of the evaluator.

Based on discussions, I think the best thing to do here is to parse these as `TraceBinding` like we do with the trace column declarations. Then instead of having a new type of value for `Parameter(TraceParameterAccess)`, we'll just use the old `TraceElement(IndexedTraceAccess)` and do absolute indexing based on the trace defined in the evaluator signature.

Then when we call the evaluator, we'll do something like `evaluator.clone_with_offsets` to clone the graph and update each node containing a value of type `TraceElement(IndexedTraceAccess)` to have a new `TraceElement(IndexedTraceAccess)` with the correct offset. These offsets will be calculated according to the arguments with which the evaluator is called and passed into `clone_with_offsets` from within `process_integrity_stmt`

## Third point

> Third, it feels like ConstraintBuilder is now used for too many things (i.e., evaluators don't care about boundary constraints, and yet every ConstraintBuilder contains info about constrainted_boundaries etc.

I agree, but this presents challenges. In fact, I started out by trying to do what you proposed, but it was less clean.
Essentially - we actually want to:
1. consume the graph
2. consume the constraints
3. reset the symbol table to the one that was used to create the ConstraintBuilder that processed the evaluator
4. update the map of evaluators to include the one we just processed

Basically, before processing an evaluator we actually want a new clean ConstraintBuilder containing the base symbol table and a new evaluators map that includes the previously processed evaluators.

For the moment, I think it will be easier to take the approach I've taken here and maybe revisit after the other changes

tohrnii:
superseded by #241, #242 and #243

grjte:
Reopening this PR, since it is the current point of discussion for the design of the evaluators implementation

grjte:
This PR has now been updated to complete a simple initial implementation of evaluator functions according to the discussion above and with the benefit of a lot of refactoring.

Notes:
- when evaluator functions are called, the graph and the constraints are simply appended to the graph/constraints of the caller rather than merging them. I did this to keep things simpler. We can revisit this and optimize it in the future

## Errors:
Durrently, calling an evaluator function and passing only the auxiliary trace does not work, so I left these 2 tests "ignored". The update for this would take a bit more work, and I'm not convinced we want to do it, so I think it would be better to merge this PR first and then revisit that in a subsequent PR.

Based on the discussion in the refactor PR (https://github.com/0xPolygonMiden/air-script/pull/259), we may want to get rid of the main/aux keywords. In that case, the most straightforward thing to do with the evaluator parameters is what I suggested in that discussion. I think implementing this would be simpler and cleaner than following the old plan.

In this case, we would just use empty segments instead of keywords. Instead of doing `main:[...], aux: [...]` it could just be `[...], [...]`.

```
ev foo([a,b,c], [p]):
  ...

ev bar([], [p]):
  ...

ev bat([a, b]):
  ...

enf foo([x, y, z], [p])
enf bar([], [p])
enf bat([x, y])
```

ir/src/tests/evaluators.rs
```diff
@@ -1,10 +1,10 @@
 use super::{parse, AirIR};
 
-#[ignore]
 #[test]
 fn simple_evaluator() {
     let source = "
     ev advance_clock(main: [clk]):
+        let z = a + 1
```
tohrnii, line 7: Question: Shouldn't this be invalid but the test passes. I'm assuming that's what you were testing here. I think we currently don't validate the declaration unless we use the variable and this should be a bug. Did I miss something here or should we open an issue for this?

tohrnii (approved):
Thanks @grjte, looks good to me. I think there probably is a pre-existing bug that we should address in another PR.

ir/src/lib.rs
```diff
@@ -81,47 +88,63 @@ impl AirIR {
...
+        }
+
+        // process & validate the public inputs
+        symbol_table.insert_public_inputs(pub_input_decls)?;
 
-        // TODO: process evaluators
+        // process & validate the random value declarations
+        for rand_decls in rand_decls.into_iter() {
+            symbol_table.insert_random_values(rand_decls)?;
+        }
```
bobbinth, line 138: Maybe I'm not following something, but shouldn't this happen before we process evaluators? If we don't do that, how would evaluators have access to random values?

ir/src/lib.rs
```diff
@@ -81,47 +88,63 @@ impl AirIR {
...
                 }
                 ast::SourceSection::PeriodicColumns(columns) => {
                     // process & validate the periodic columns
                     symbol_table.insert_periodic_columns(columns)?;
                 }
                 ast::SourceSection::RandomValues(values) => {
-                    symbol_table.insert_random_values(values)?;
                     validator.exists("random_values");
+                    // save the random value declaration for later processing.
+                    rand_decls.push(values);
```
bobbinth, line 106: Why not process random values here? (similar to what we do with periodic columns and constants)

bobbinth (commented):
Thank you! Looks very clean and easy to follow. I left a few small comments inline. One thing I didn't quite understand is handling of random values.

ir/src/tests/evaluators.rs
grjte: Good point! I didn't even notice that actually. I think I just made a mistake. You're right though, we don't validate the declaration unless the variable is used. If you can open an issue, that would be great!

ir/src/lib.rs
grjte: You're right! It was because in our discussion before we only said to process the constants and periodic columns before the evaluators, and so that was what was in the todo list, and I didn't notice the bug

ir/src/constraint_builder/evaluator_functions.rs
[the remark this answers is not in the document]
grjte: Sure. We've been a bit inconsistent about it. We can switch to using evaluators everywhere

ir/src/tests/evaluators.rs
grjte: Although really, this is just dead code, so validating it is not a priority. @bitwalker what's your take on these situations?

ir/src/tests/evaluators.rs
```diff
@@ -24,7 +24,30 @@ fn simple_evaluator() {
     assert!(result.is_ok());
 }
 
-#[ignore]
+#[test]
+fn evaluator_with_variables() {
+    let source = "
+    ev advance_clock(main: [clk]):
+        let z = clk + 1
+        enf clk' = z
```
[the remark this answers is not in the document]
grjte: You're right that the ordering doesn't matter in terms of functionality. Do you prefer putting functions at the bottom as well? Currently we've been doing constants, pure functions, then evaluators. However, that's been the ordering in module definitions where there's no `integrity_constraints` section anyway. I think we'll avoid defining evaluators in the top level AIR def, so this point is possibly moot, except in tests like this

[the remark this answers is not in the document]
bobbinth: > Do you prefer putting functions at the bottom as well?

Probably yes, but this is a fairly minor point either way.

ir/src/tests/evaluators.rs
bitwalker: @grjte I'm very much in favor of performing semantic analysis on all code, regardless of whether it is dead or not, but we need richer diagnostics infra for that to be worthwhile IMO. Otherwise I suspect it is easy to accidentally assume that something is working a certain way (because compilation succeeds) when it actually isn't. When we can do the diagnostics refactoring we discussed briefly, that would be a good time to introduce some of these additional warnings/errors.

PR #228, opened 2023-03-22
refactor: refactor trace access

tohrnii:
This PR changes makes the following changes to the parser:
1.  `TraceCols` to `TraceBinding`.
2. `NamedTraceAccess` struct and enum variant in `Expression` to `TraceBindingAccess`.
3. `IndexedTraceAccess` enum variant in `Expression` to `TraceAccess`.


grjte:
I've been trying to leave the following comment at line 17 of `parser/src/ast/evaluator_function.rs`

> Based on the discussion in #227 and outside of github, this should be `Vec<TraceBinding>`. We want to treat it exactly like the way that we parse the declarations of trace columns. We'll use absolute indexing, and all nodes related to this in the graph will be stored as `TraceAccess` (formerly `IndexedTraceAccess`). Then when an evaluator function is called, we'll map each `TraceAccess` to a new `TraceAccess` depending on the absolute offsets of the arguments of the the EvaluatorFunctionCall

grjte:
A couple of general comments:
- the CI checks should be fixed
- the renaming of the types should probably be carried through to the IR in the same PR so that we don't have failures

parser/src/ast/evaluator_function.rs
```diff
@@ -1,27 +1,25 @@
-use super::{Identifier, IntegrityStmt, TraceCols};
+use super::{Identifier, IntegrityStmt, TraceBinding, TraceBindingAccess};
 
-/// Evaluator functions take descriptions of the main and auxiliary traces as input, and enforce
+/// Evaluator functions take column groups as parameters where each column group is a set of
+/// columns in the particular trace segment that are passed to the evaluator function, and enforce
```
grjte, line 4: nit: we should update the terminology in this comment from column groups to trace bindings

parser/src/ast/evaluator_function.rs
```diff
@@ -48,27 +41,23 @@ impl EvaluatorFunction {
...
     /// of the evaluator function.
-    pub fn into_parts(self) -> (String, Vec<TraceCols>, Vec<TraceCols>, Vec<IntegrityStmt>) {
-        (
-            self.name.into_name(),
-            self.main_trace,
-            self.aux_trace,
-            self.integrity_stmts,
-        )
+    pub fn into_parts(self) -> (String, Vec<TraceBinding>, Vec<IntegrityStmt>) {
+        (self.name.into_name(), self.params, self.integrity_stmts)
```
grjte, line 45: nit: maybe we should update the comment for this function as well.

parser/src/ast/mod.rs
```diff
@@ -62,31 +63,18 @@ pub enum SourceSection {
...
-pub struct TraceCols {
-    name: Identifier,
-    size: u64,
-}
-
-impl TraceCols {
-    pub fn new(name: Identifier, size: u64) -> Self {
-        Self { name, size }
-    }
+/// Given a vector of identifiers and their trace segment, returns a vector of trace bindings.
```
grjte, line 66: nit: we're actually given a vector of (identifier, size) pairs

parser/src/parser/tests/evaluator_functions.rs
```diff
@@ -1,8 +1,10 @@
+use air_script_core::{Range, TraceBinding};
+
```
grjte, line 2: nit: we can probably import these from `crate::ast` below, or at least remove the extra empty line

grjte (approved):
Looks good to me! I left a few minor nits inline regarding comments. They're non-blocking, but it would be great to update them before merging

Issue #209, opened 2023-03-07
write MidenVM constraints for the stack (general) and determine operation flag handling


grjte:
For the stack operations we need to determine a clean way of managing operation flags. That should be determined in this issue, which should be done before writing any other stack constraints.

The general constraints are currently implemented in Rust [here](https://github.com/0xPolygonMiden/miden-vm/blob/159a04f3d4648f77d664642409e521438378bd6c/air/src/stack/mod.rs#L148), and the operation flags handling is done [here](https://github.com/0xPolygonMiden/miden-vm/blob/main/air/src/stack/op_flags/mod.rs).

PR #239, opened 2023-03-31
Write general stack constraints in AirScript

Overcastan:
Addressing: #209 
This PR adds constraints for every VM operation.

### DIFF
--- a/constraints/miden-vm/stack.air
+++ b/constraints/miden-vm/stack.air
@@ -379,6 +379,10 @@ fn f_ctrl(b: vector[7], extra: scalar) -> scalar:
     return f_sjsl + f_errh + f_call(b, extra) + f_syscall(b, extra)
 
 
+fn binary_check_sub(elem: scalar) -> scalar:
+    return elem^2 - elem
+
+
 ### Helper evaluators #############################################################################
 
 # Enforces that the provided columns must be zero.
@@ -392,16 +396,16 @@ ev is_unchanged(main: [values[4]]):
     enf v' = v for v in values
 
 
-# Enforces that the provided columns must be binary.
-ev is_binary(main: [a]):
-    enf a^2 = a
+ev range_check_16bit()
 
 
-0######################################################################
+### Stack Air Constraints #########################################################################
 
 # Enforces the constraints on the stack.
 # TODO: add docs for columns
 # ev stack_constraints(main: [stack[16], bookkeeping[2], helper], aux: [op_bits[7], extra, hasher5]):
+
+# No stack shift operations
 
 ev noop(main: [s[16]], aux: [b[7]]):
     enf f_dup(b) * delta(elem) = 0 for elem in stack
@@ -424,7 +428,7 @@ ev incr(main: [s[16]], aux: [b[7]]):
     enf f_incr(b) * delta(s[i]) = 0 for i in 1..16
 
 ev not(main: [s[16]], aux: [b[7]]):
-    enf f_not(b) * is_binary([s[0]])
+    enf f_not(b) * binary_check_sub(s[0])
     enf f_not(b) * (s[0]' - !s[0]) = 0
     enf f_not(b) * delta(s[i]) = 0 for i in 1..16
 
@@ -474,7 +478,7 @@ ev advpopw(main: [s[16]], aux: [b[7]]):
     enf f_advpopw(b) * delta(s[i]) = 0 for i in 4..16
 
 ev expacc(main: [s[16]], aux: [b[7], helper]):
-    enf f_expacc(b) * is_binary(s[0]')
+    enf f_expacc(b) * binary_check_sub(s[0]')
     enf f_expacc(b) * (s[1]' - s[1]^2) = 0
     enf f_expacc(b) * (helper - ((s[1] - 1) * s[0]' + 1)) = 0
     enf f_expacc(b) * (s[2]' - s[2] * helper) = 0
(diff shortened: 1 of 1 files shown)


constraints/miden-vm/stack.air
```diff
@@ -0,0 +1,715 @@
...
+    enf f_assert(b) * (s[i]' - s[i + 1]) = 0 for i in 0..15
+
+ev eq(main: [s[16]], aux: [b[7], helper]):
+    enf f_eq(b) * (s[0]' * (s[0] - s[1])) = 0
+    enf f_eq(b) * (s[0]' - (1 - (s[0] - s[1]) * helper)) = 0
+    enf f_eq(b) * (s[i]' - s[i + 1]) = 0 for i in 1..15
+
+ev add(main: [s[16]], aux: [b[7]]):
+    enf f_add(b) * (s[0]' - (s[0] + s[1])) = 0
+    enf f_eq(b) * (s[i]' - s[i + 1]) = 0 for i in 1..15
```
bobbinth, line 581: I would not make evaluators aware of the flag values - this could be handled by the calling evaluator as a part of `match enf` statement. So, I would rewrite the evaluators in this file to work like that. For example, the evaluator for `add` operation could look like this:
```
ev add(main: [s[16]]):
  enf s[0]' = s[0] + s[1]
  enf s[i]' = s[i + 1] for i in 1..15
```
Also, do we already support computed indexes (i.e., `s[i + 1]`)? Or is this something planned?

bobbinth: And to provide more details about why I meant by "calling evaluator": I'm imagining that we'll have a vector of 88 mutually-exclusive flags and then we'll select based on that. If we had only 4 operations, this could look something like this:
```
# this returns a vector of mutually exclusive operation flags
let op_flags = compute_op_flags()
match enf
  op_add(stack_top) when op_flags[0]
  op_neg(stack_top) when op_flags[1]
  op_mul(stack_top) when op_flags[2]
  op_inv(stack_top) when op_flags[3]
```
In the above `stack_top` is a reference to the 16 columns representing 16 columns of the stack top.

constraints/miden-vm/stack.air
```diff
@@ -396,320 +493,396 @@ ev is_unchanged(main: [values[4]]):
...
+    enf is_binary([s[0]])
+    enf s[0]' = !s[0]
+    enf delta(s[i]) = 0 for i in 1..16
 
-ev fmpadd(main: [s[16]], aux: [b[7], fmp]):
-    enf f_fmpadd(b) * (s[0]' - s[0] - fmp) = 0
-    enf f_fmpadd(b) * delta(s[i]) = 0 for i in 1..16
+ev fmpadd(main: [s[16]], aux: [fmp]):
+    enf s[0]' = s[0] + fmp
+    enf delta(s[i]) = 0 for i in 1..16
```
bobbinth, line 610: `fmp` is a part of main trace (not auxiliary trace). So, this should look like this:
```
ev op_fmpadd(main: [s[16], fmp]):
  enf s[0]' = s[0] + fmp
  enf s[i]' = s[i] for i in 1..16
```

constraints/miden-vm/stack.air
```diff
@@ -396,320 +493,396 @@ ev is_unchanged(main: [values[4]]):
...
 # are unique for i/o ops?
-ev mload(main: [s[16]], aux: [b[7], helper[3], clk, b_chip]):
+ev mload(main: [s[16]], aux: [helper[3], clk, b_chip]):
     let v = alpha[5] * s[0]' + sum([alpha[i + 5] * helper[3 - i]'])
     let op_mem_read = 12
     let u_mem = alpha[0] + alpha[1] * op_mem_read + alpha[3] * s[0] + alpha[4] * clk + v
-    enf f_mload(b) * (b_chip' * u_mem - b_chip) = 0
-    enf f_mload(b) * delta(s[i]) = 0 for i in 1..16
+    enf b_chip' * u_mem = b_chip
+    enf delta(s[i]) = 0 for i in 1..16
```
bobbinth: Similar points to the previous comment: `helper` and `clk` columns are a part of the main trace (not auxiliary trace). Also, referencing random values should be done with `$` prefix. I would rewrite this like this:
```
ev op_mload(main: [s[16], helpers[3], clk], aux: [b_chip]):
    let v = $alpha[5] * s[0]' + sum([$alpha[i + 5] * helper[3 - i]']) # this is missing for part of the statement
    let op_mem_read = 12
    let u_mem = $alpha[0] + $alpha[1] * op_mem_read + $alpha[3] * s[0] + $alpha[4] * clk + v
    enf b_chip' * u_mem = b_chip
    enf s[i]' = s[i] for i in 1..16 # this is missing constraint against s[0]
```

constraints/miden-vm/stack.air
```diff
@@ -396,320 +493,396 @@ ev is_unchanged(main: [values[4]]):
...
 # Enforces the constraints on the stack.
 # TODO: add docs for columns
-# ev stack_constraints(main: [stack[16], bookkeeping[2], helper], aux: [op_bits[7], extra, hasher5]):
+ev stack_constraints(main: [stack[16], bookkeeping[2], h0], aux: [op_bits[7], extra, hasher5, fmp, helper[3], clk, b_chip]):
+    let op_flags = compute_op_flags(op_bits, extra)
+    match enf:
+        noop([stack]) when op_flags[0]
+        eqz([stack, h0]) when op_flags[1]
+        neg([stack]) when op_flags[2]
+        inv([stack]) when op_flags[3]
```
bobbinth, line 514: I would try to align `when` parts as much as possible to make this easier to read.

constraints/miden-vm/stack.air
Overcastan: Ah, indeed, my bad. And I also made the same mistake with saying `v'` in the decoder PR calling `..._next` flags.

constraints/miden-vm/stack.air
Overcastan: I'm not sure what constraint we should apply to the `s[0]`. For `MLOAD` we check memory access correctness and no change starting from position 1.

constraints/miden-vm/stack.air
```diff
@@ -0,0 +1,957 @@
...
+    let v_hi = 2^16 * helper[3] + helper[2]
+    let v_lo = 2^16 * helper[1] + helper[0]
+    enf (1 - m * (2^32 - 1 - v_hi)) * v_lo = 0
+
+
+### Stack Air Constraints #########################################################################
+
+# Enforces the constraints on the stack.
+# TODO: add docs for columns
+ev stack_constraints(main: [stack[16], bookkeeping[2], h0, helper[6], clk, fmp], aux: [op_bits[7], extra, hasher5, b_chip, b_range]):
```
bobbinth, line 512: A few comments on these:

- I would probably change `stack` to `stack_top`.
- I'd combine `bookkeeping` and `h0` into something like `stack_helpers`.
- I'd rename `helpers` to `op_helpers`.
- `op_bits`, `extra` are part of the main trace - not auxiliary trace. Also, I'd probably lump `extra` together with `op_bits`.
- What is `hasher5`?

Overall, this could loo like so:
```
ev stack_constraints(main: [stack_top[16], stack_helpers[2], op_helper[6], op_bits[8], clk, fmp], aux: [hasher5, b_chip, b_range]):
```

constraints/miden-vm/stack.air
```diff
@@ -0,0 +1,957 @@
...
+        u32split([stack, helper])              when op_flags[68]
+        u32assert2([stack, helper])            when op_flags[69]
+        u32add3([stack, helper])               when op_flags[70]
+        u32madd([stack, helper])               when op_flags[71]
+
+
+# No stack shift operations
+
+ev noop(main: [s[16]]):
+    enf elem' = elem for elem in s
```
bobbinth, line 599: Since we are doing it everywhere else, we can probably use indexed access here as well. For example:
```
ev noop(main: [s[16]]):
    enf s[i]' = s[i] for i in 0..16
```

constraints/miden-vm/stack.air
```diff
@@ -0,0 +1,957 @@
...
+    enf s[i]' = s[i + 1] for i in 1..15
+
+ev or(main: [s[16]]):
+    enf is_binary([s[i]]) = 0 for i in 0..2
+    enf s[0]' = s[1] + s[0] - s[1] * s[0]
+    enf s[i]' = s[i + 1] for i in 1..15
+
+ev u32and(main: [s[16]], aux: [b_chip]):
+    let op_u32and = 2
+    enf b_chip' * ($alpha[0] + $alpha[1] * op_u32and + $alpha[2] * s[0] + $alpha[3] * s[1] + $alpha[4] * s[0]') = b_chip
```
bobbinth, line 788: We'll probably need to put bus constraints into a separate module as they need to include both left and right sides of the bus, and here we are computing only the left side.

I'm thinking that we'd probably end up having each bus constraints being a separate module.

constraints/miden-vm/stack.air
```diff
@@ -0,0 +1,1015 @@
...
+    let f_errh = op_bits[6] & op_bits[5] & op_bits[4]
+
+    return f_sjsl + f_errh + f_call(op_bits) + f_syscall(op_bits)
+
+
+fn compute_op_flags(op_bits: vector[8]) -> vector[88]:
+    return [
+        f_noop(op_bits),
+        f_eqz(op_bits),
+        f_neg(op_bits),
```
bobbinth, line 386: I'm not sure putting each operation flag into a separate function gives us much in terms of modularity or readability. Maybe we should get rid of individual functions and just add some comments here? For example, this could look like:
```
fn compute_op_flags(op_bits: vector[8]) -> vector[88]:
    return [
        f_000(op_bits) & f_x0000(op_bits), # noop
        f_000(op_bits) & f_x0001(op_bits), # eqz
        f_000(op_bits) & f_x0010(op_bits), # neg
```

Overcastan: I agree. I thought that we would use each flag individually, but since we have the `compute_op_flags` function they lose their main purpose.

constraints/miden-vm/stack.air
```diff
@@ -0,0 +1,738 @@
...
+    enf b0_is_zero([op_bits])
+
+# TODO: add constraints
+ev pipe([s[16], op_bits[8], op_helpers[6]]):
+
+
+# Bus constraint is implemented in a separate file
+ev mstream([s[16], op_bits[8], op_helpers[6]]):
+    enf s[i]' = s[i] for i in 8..12
+    enf s[i]' = s[i] for i in 13..16
```
tohrnii, line 688: Should there also be another constraint to enforce that `s[12]' = s[12] + 2`?

Overcastan: Indeed! I mistakenly decided that this constraint will be part of the bus constraint, but it is not. Thank you!

tohrnii (changes requested):
Thank you @Overcastan, looks great! I've left a few small comments. I haven't reviewed the `u32` constraints yet.

constraints/miden-vm/stack.air
```diff
@@ -0,0 +1,738 @@
...
+ev mloadw([s[16], clk]):
+    enf s[i]' = s[i + 1] for i in 4..15
+
+# Bus constraint is implemented in a separate file
+ev mstore([s[16], clk]):
+    enf s[i]' = s[i + 1] for i in 0..15
+
+# Bus constraint is implemented in a separate file
+ev mstorew([s[16], clk]):
+    enf s[i]' = s[i + 1] for i in 0..15
```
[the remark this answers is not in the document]
Overcastan: I just forget to remove `clk` from input values when I was moving bus constraints to the separate file, so both `clk` and `ctx` will be used for bus constraints and are not needed here. Great catch, thank you!

constraints/miden-vm/stack.air
```diff
@@ -0,0 +1,739 @@
+mod StacktAir
```
tohrnii, line 1: nit: Typo

constraints/miden-vm/stack.air
```diff
@@ -0,0 +1,739 @@
...
+        op_bits[7] & op_bits[4] & !op_bits[3] & !op_bits[2],  # end
+        op_bits[7] & op_bits[4] & !op_bits[3] & op_bits[2],   # repeat
+        op_bits[7] & op_bits[4] & op_bits[3] & !op_bits[2],   # respan
+        op_bits[7] & op_bits[4] & op_bits[3] & op_bits[2],    # halt
+    ]
+
+
+### Helper evaluators #############################################################################
+
+# Enforces that the provided columns must be binary.
```
tohrnii, line 205: minor nit: Should this be "Enforces that the provided columns are binary."

tohrnii (approved):
Thank you. Looks good. Should the remaining constraints be added in the same PR or a follow up PR?

Overcastan:
@tohrnii I think it will be better to add remaining constraints in a separate PR, because I'll need some time to understand how they work and come up with some constraints.

Al-Kindi-0:
Superseded by work in #455

PR #241, opened 2023-04-03
refactor(IR): Refactor IR

tohrnii:


ir/src/constraint_builder/mod.rs
```diff
@@ -1,3 +1,5 @@
+use std::fmt;
+
```
grjte, line 2: Let's remove the extra line here and also import the items below from super instead of from `constraints`

ir/src/constraint_builder/mod.rs
```diff
@@ -9,60 +11,118 @@ use super::{
 };
 
 mod boundary_constraints;
+use air_script_core::TraceBinding;
```
grjte, line 14: let's import this from `super`

ir/src/constraint_builder/mod.rs
grjte, line 2: Let's remove the extra line here and also import the items below from super instead of from `constraints`

grjte (changes requested):
Let's keep this as a pure renaming refactor, so the following things should be removed:

- introduce ConstraintBuilderContext
- Refactors the processing of boundary and integrity statements

As a note, these changes should have ideally been done as 3 distinct commits anyway (renaming, intro context, stmt refactor), which would have made this adjustment easier. You can reset all the changes and then use `git add -p` to select only the renaming changes for the commit.

ir/src/constraint_builder/boundary_constraints.rs
```diff
@@ -72,7 +72,7 @@ impl ConstraintBuilder {
                );
 
                 // add its expression to the constraints graph.
-                let rhs = self.insert_expr(constraint.value())?;
+                let rhs = self.insert_expr(constraint.value().clone())?;
```
grjte, line 75: Since we're changing to a consume pattern, this clone shouldn't be necessary. Instead, at the top of this function we should have the following:

```
let (boundary, access, value) = constraint.into_parts();
```

Then this will just be:
```
let rhs = self.insert_expr(value)?;
```

ir/src/constraint_builder/expression.rs
```diff
@@ -114,7 +123,7 @@ impl ConstraintBuilder {
         rhs: &Expression,
     ) -> Result<NodeIndex, SemanticError> {
         // add base subexpression.
-        let lhs = self.insert_expr(lhs)?;
+        let lhs = self.insert_expr(lhs.clone())?;
```
grjte, line 126: Since we're changing to a consume pattern, the signature of this method should change, then this `clone` won't be necessary:

```
    fn insert_exp_op(
        &mut self,
        lhs: Expression,
        rhs: Expression,
```

ir/src/constraint_builder/integrity_constraints/mod.rs
```diff
@@ -25,10 +25,10 @@ impl ConstraintBuilder {
         match stmt {
             IntegrityStmt::Constraint(ConstraintType::Inline(constraint), _) => {
                 // add the left hand side expression to the graph.
-                let lhs = self.insert_expr(constraint.lhs())?;
+                let lhs = self.insert_expr(constraint.lhs().clone())?;
```
grjte, line 28: Since we're changing to a consume pattern, we should break the constraint into parts at the top of this case, then the `clone` won't be necessary.

```
                let (lhs, rhs) = constraint.into_parts();
                // add the left hand side expression to the graph.
                let lhs = self.insert_expr(lhs)?;
```

ir/src/symbol_table/mod.rs
```diff
@@ -25,7 +22,7 @@ pub use value::{ConstantValue, Value};
 
 /// SymbolTable for identifiers to track their types and information and enforce uniqueness of
 /// identifiers.
-#[derive(Default, Debug)]
+#[derive(Default, Debug, Clone)]
```
grjte, line 25: this shouldn't be required right now.

ir/src/symbol_table/mod.rs
```diff
@@ -230,7 +227,7 @@ impl SymbolTable {
         Ok(TraceAccess::new(
             columns.trace_segment(),
             columns.offset() + trace_access.col_offset(),
-            1,
+            columns.size(),
```
grjte, line 230: Shouldn't this size should be determined by the size of the `trace_access`?

grjte (changes requested):
Thanks @tohrnii, just a few more small things to avoid some unnecessary cloning.

ir/src/symbol_table/mod.rs
tohrnii: Thanks, fixed.

ir/src/constraint_builder/boundary_constraints.rs
```diff
@@ -71,8 +71,9 @@ impl ConstraintBuilder {
                    "The boundary constraint's domain should be {lhs_domain:?}, but the domain {domain:?} was inferred by the graph",
                );
 
+                let (_, _, value) = constraint.into_parts();
```
grjte, line 74: We should do the `into_parts` call at the top of the case and then use the access and boundary, instead of calling `constraint.access()` and `constraint.boundary()` like above

```
                let (boundary, access, value) = constraint.into_parts();
                let trace_access = self.symbol_table.get_trace_binding_access(&access)?;
                let domain = boundary.into();
```

PR #246, opened 2023-04-04
Remove usage of next row operator from functions

Overcastan:
Removed usage of next row operator `'` from functions.


grjte (approved):
Looks good, thank you!

Issue #205, opened 2023-03-07
write MidenVM constraints for the range checker 8-bit range checks virtual table (multiset check constraints)


grjte:
After we've written the main range checker constraints we need to add the multiset check constraints for the range checker's virtual table. The constraints are described [here](https://0xpolygonmiden.github.io/miden-vm/design/range.html#8-bit-range-checks-table) and implemented in Rust [here](https://github.com/0xPolygonMiden/miden-vm/blob/main/air/src/range.rs).

Al-Kindi-0:
Closed by #248  (although this not up-to-date with the current state of things)

PR #248, opened 2023-04-06
Write 8-bit range checks constraints in AirScript

Overcastan:
Addressing: #205 
This PR adds constraint for the 8-bit range checks table (multiset check constraint).


constraints/miden-vm/range_checker_8bit.air
```diff
@@ -0,0 +1,32 @@
...
+
+
+# Set to 1 when we include two copies of the value into the running product.
+fn f2(s0: scalar, s1: scalar) -> scalar:
+    return !s0 & s1
+
+
+# Set to 1 when we include four copies of the value into the running product.
+fn f3(s0: scalar, s1: scalar) -> scalar:
+    return s0 & s1
```
bobbinth, line 22: I would probably combine these into a single function which returns a vector with 4 flags. Maybe call it something like `get_multiplicity_flags`.

constraints/miden-vm/range_checker_8bit.air
```diff
@@ -0,0 +1,32 @@
...
+
+
+### 8-bit Range Checker Air Constraints #################################################################
+
+ev range_checker_8bit(main: [t, s0, s1, v, p0]):
```
bobbinth, line 27: `p0` should be a part of the auxiliary trace. I'd also rename this to just `range_checker`. So, this would look something like this:
```
ev range_checker_8bit(main: [t, s0, s1, v], aux: [p0]):
```

constraints/miden-vm/range_checker_8bit.air
```diff
@@ -0,0 +1,32 @@
...
+    # z represents how a row in the execution trace is reduced to a single value.
+    let z = ($alpha[0] + v)^4 * f3(s0, s1) + ($alpha[0] + v)^2 * f2(s0, s1) + ($alpha[0] + v) * f1(s0, s1) + f0(s0, s1)
```
bobbinth, line 29: With flags array, we can rewrite this as:
```
let val = $alpha[0] + v
let f = get_multiplicity_flags(s0, s1)
let z = val^4 * f[3] + val^2 * f[2] + val * f[1] + f[0]
```

bobbinth (changes requested):
Thank you! I left a couple of comments inline.

constraints/miden-vm/range_checker_8bit.air
Overcastan: But won't such a name interfere with the work of the evaluator of the main, already implemented range checker? Because it also has `range_checker` name.

tohrnii:
@Overcastan question: should these constraints not be part of the original range checker module?

Overcastan:
I'm not sure. These constraints indeed look like they could be placed together, although in many places we decide to move range checker constraints to the separate files. So, I'm not sure, but I can merge them if you think it'll be better. What do you think, @tohrnii, @bobbinth, @grjte?

constraints/miden-vm/range_checker_8bit.air
```diff
@@ -0,0 +1,23 @@
...
+# p[0] set to 1 when we don't include the value into the running product.
+# p[1] set to 1 when we include the value into the running product.
+# p[2] set to 1 when we include two copies of the value into the running product.
+# p[3] set to 1 when we include four copies of the value into the running product.
+fn get_multiplicity_flags(s0: scalar, s1: scalar) -> vector[4]:
+    return [!s0 & !s1, s0 & !s1, !s0 & s1, s0 & s1]
+
+### 8-bit Range Checker Air Constraints #################################################################
+
+ev range_checker_8bit(main: [t, s0, s1, v], aux: [p0]):
```
grjte, line 15: Let's add a comment above this evaluator. It can reuse some of the explanation from the [8-bit range checks table](https://0xpolygonmiden.github.io/miden-vm/design/range.html#8-bit-range-checks-table) section

grjte (changes requested):
@Overcastan this looks good to me. I just left 2 minor requests about updating the documentation comments. Also, I agree that this should be combined with the other range checker constraints. I think it will be clearer to have them all in the same module.

constraints/miden-vm/range_checker_8bit.air
grjte: Also, we can remove the `main` and `aux` keywords

constraints/miden-vm/range_checker_8bit.air
```diff
@@ -0,0 +1,23 @@
+mod RangeChecker8BitAir
+
+### Helper functions ##############################################################################
+
+# Returns array of mutually exclusive multiplicity flags.
+# p[0] set to 1 when we don't include the value into the running product.
+# p[1] set to 1 when we include the value into the running product.
+# p[2] set to 1 when we include two copies of the value into the running product.
+# p[3] set to 1 when we include four copies of the value into the running product.
```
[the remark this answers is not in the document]
Overcastan: Completely agree, I didn't even find why I used `p` instead of `f`, that's strange.

constraints/miden-vm/range_checker.air
```diff
@@ -0,0 +1,61 @@
+mod RangeCheckerAir
+
+### Helper functions ##############################################################################
+
+# Returns binary negation of the value.
+fn binary_not(e: scalar) -> scalar:
+    return 1 - e
```
grjte, line 7: nit: I know this was here before, but I find it strange to name this e, since e has a meaning and is never used as an indeterminate. I would probably use `v` or `x` instead

constraints/miden-vm/range_checker.air
```diff
@@ -0,0 +1,61 @@
...
+    enf t * binary_not(t') = 0
+
+    # Ensure that when column t flips, column v must equal 255
+    enf v = 255 when t' & !t
+
+    # Ensure that when column t flips, v' must be reset to 0
+    enf v' = 0 when t' & !t
+
+# Enforces the internal correctness of the 16-bit section of the Range Checker.
+ev range_checker_8bit([t, s0, s1, v], [p0]):
```
grjte, line 36: nit: Let's rename this evaluator to "virtual_table" and update the comment to "The virtual table enforces an 8-bit range check for each row transition in the 16-bit section of the range checker, which enforces its internal correctness."

constraints/miden-vm/range_checker.air
```diff
@@ -0,0 +1,61 @@
...
+
+
+### Helper evaluators #############################################################################
+
+# Enforces that column must be binary.
+ev is_binary([v]):
+    enf v^2 = v
+
+# Enforces correct transition from 8-bit to 16-bit section of the table.
+ev transition_8_to_16_bit_validity([t, v]):
```
grjte, line 25: nit: I think we can remove "_validity" from the end of this to shorten it a bit

grjte (changes requested):
Looks great, thank you! A couple final nits for clarity/succinctness in the comments & naming, then we'll be ready to merge

constraints/miden-vm/range_checker.air
Overcastan: I think we can simply get rid of this function and use `!` instead.

grjte (approved):
Looks good to me, thank you!

grjte (changes requested):
@Overcastan actually there are a couple errors remaining - please check the comments inline

constraints/miden-vm/range_checker.air
```diff
@@ -0,0 +1,58 @@
...
+
+### Range checker Air Constraints #################################################################
+
+ev range_checker([t, s0, s1, v], [p0]):
+    # Check selector flags are binary.
+    let selectors = [t, s0, s1]
+    enf is_binary([s]) for s in selectors
+
+    # Constrain the row transitions in the 8-bit section of the table.
+    enf (v' - v) * (v' - v - 1) = 0 when !t0'
```
bobbinth, line 52: Reference to `t0` does not seem to be right. Should it be `t`?

Also, I'd prefer a more descriptive comment - something which describes the intent of the constraint.

grjte (changes requested):
Thanks for fixing those issues @Overcastan. I think we just need the final fix Bobbin pointed out for `t0` -> `t`

constraints/miden-vm/range_checker.air
Overcastan: Yes, it should be `t`, not sure why it suddenly became `t0`, looks like I miss clicked. Thank you!

PR #253, opened 2023-04-13
Write chiplet module constraints in AirScript

Overcastan:
Addressing: #201 
This PR adds constraints for the chiplet module.

### DIFF
--- a/constraints/miden-vm/chiplets.air
+++ b/constraints/miden-vm/chiplets.air
@@ -1,0 +1,67 @@
+use bitwise::bitwise_chiplet
+use hash::hash_chiplet
+use memory::memory_chiplet
+
+mod ChipletsConstraintsAir
+
+### Helper evaluators #############################################################################
+
+# Enforces that the provided columns must be binary.
+ev is_binary(main: [a]):
+    enf a^2 = a
+
+# Enforces that constraints are only ever selected for one chiplet at a time.
+ev chiplet_constraints(main: [s[3], chip_columns[15]]):
+    # Variables for hash chiplet.
+    let hash_s_3 = [s[1], s[2], chip_columns[0]]
+    let hash_r = chip_columns[1]
+    let hash_h_12 = [chip_columns[2], chip_columns[3], chip_columns[4], 
+                     chip_columns[5], chip_columns[6], chip_columns[7], 
+                     chip_columns[8], chip_columns[9], chip_columns[10], 
+                     chip_columns[11], chip_columns[12], chip_columns[13]]
+    let hash_i = chip_columns[14]
+
+    # Variables for bitwise chiplet.
+    let bitwise_s = s[2]
+    let bitwise_a = chip_columns[0]
+    let bitwise_b = chip_columns[1]
+    let bitwise_a_limb_4 = [chip_columns[2], chip_columns[3], chip_columns[4], chip_columns[5]]
+    let bitwise_b_limb_4 = [chip_columns[6], chip_columns[7], chip_columns[8], chip_columns[9]]
+    let bitwise_zp = chip_columns[10]
+    let bitwise_z = chip_columns[11]
+
+    # Variables for memory chiplet.
+    let memory_s_2 = [chip_columns[0], chip_columns[1]]
+    let memory_ctx = chip_columns[2]
+    let memory_addr = chip_columns[3]
+    let memory_clk = chip_columns[4]
+    let memory_v_4 = [chip_columns[5], chip_columns[6], chip_columns[7], chip_columns[8]]
+    let memory_d_2 = [chip_columns[9], chip_columns[10]]
+    let memory_t = chip_columns[11]
+
+    match enf: 
+        hash_chiplet([hash_s_3, hash_r, hash_h_12, hash_i]) when !s[0]
+        bitwise_chiplet([bitwise_s, bitwise_a, bitwise_b, bitwise_a_limb_4, bitwise_b_limb_4, bitwise_zp, bitwise_z]) when s[0] & !s[1]
+        memory_chiplet([memory_s_2, memory_ctx, memory_addr, memory_clk, memory_v_4, memory_d_2, memory_t]) when s[0] & s[1] & !s[2]'
+
+
+# Enforces that the chiplet selector columns are set correctly.
+enf chiplet_selector_constraints(main: [s[3]]):
+    # Enforce that selectors are binary.
+    enf is_binary([s[0]])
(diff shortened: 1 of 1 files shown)


constraints/miden-vm/chiplets.air
```diff
@@ -0,0 +1,67 @@
...
+    # only change from 0 to 1.
+    enf s[0]' = s[0] when s[0]
+    enf s[1]' = s[1] when s[0] & s[1]
+    enf s[2]' = s[2] when s[0] & s[1] & s[2]
+
+
+### Chiplets Constraints ##########################################################################
+
+ev chiplets_constraints(main: [s[3], chip_columns[15]]):
+    enf chiplet_constraints([s, chip_columns])
```
grjte, line 65: very minor nit: I would use `chiplet_columns` or `chiplet_cols` instead

constraints/miden-vm/chiplets.air
```diff
@@ -0,0 +1,67 @@
...
+
+    enf chiplet_selector_constraints([s])
```
grjte, line 67: very minor nit: I would enforce the selector constraints before the other chiplet constraints, since that's the order we usually follow.

constraints/miden-vm/chiplets.air
```diff
@@ -0,0 +1,67 @@
...
+    let memory_addr = chip_columns[3]
+    let memory_clk = chip_columns[4]
+    let memory_v_4 = [chip_columns[5], chip_columns[6], chip_columns[7], chip_columns[8]]
+    let memory_d_2 = [chip_columns[9], chip_columns[10]]
+    let memory_t = chip_columns[11]
+
+    match enf: 
+        hash_chiplet([hash_s_3, hash_r, hash_h_12, hash_i]) when !s[0]
+        bitwise_chiplet([bitwise_s, bitwise_a, bitwise_b, bitwise_a_limb_4, bitwise_b_limb_4, bitwise_zp, bitwise_z]) when s[0] & !s[1]
+        memory_chiplet([memory_s_2, memory_ctx, memory_addr, memory_clk, memory_v_4, memory_d_2, memory_t]) when s[0] & s[1] & !s[2]'
```
grjte, line 45: We should be able to do something much simpler here. Instead of defining all these variables, we can just do the following:

```
match enf:
    hash_chiplet([s[1], s[2], chip_columns]) when 1s[0]
    bitwise_chiplet([s[2], chip_columns]) when s[0] & !s[1]
    memory_chiplet([chip_columns]) when s[0] & s[1] & !s[2]'
```

grjte (changes requested):
Thank you @Overcastan! This looks correct, but I think we can actually do something much simpler, as I proposed inline

constraints/miden-vm/chiplets.air
Overcastan: You mean that we will implement feature for evaluators so they would be able to rearrange input values?
So for `hash_chiplet` evaluator, for example, we could pass `[scalar, scalar, array[15]]` and evaluator will internally rearrange it to the `[array[3], scalar, array[12], scalar]`, which is declared in the evaluator definition, right? With the only condition that the input data lengths must match, of course.

constraints/miden-vm/chiplets.air
```diff
@@ -0,0 +1,67 @@
+use bitwise::bitwise_chiplet
+use hash::hash_chiplet
+use memory::memory_chiplet
```
tohrnii, line 3: Based on @grjte's comment in another PR, these should go below the module declaration.

constraints/miden-vm/chiplets.air
grjte: The evaluator doesn't rearrange anything really. When it is first processed, every constraint is added to the graph using trace accesses that are offset from zero according to how the parameters were defined. When an evaluator function call is processed, every SymbolAccess that is passed as an argument is processed into 1 or more trace accesses depending on how many columns it represents. This means that the number of columns in each segment should match between the defined parameters and the passed arguments, but the "shape" of these column groupings does not matter. Here's the [relevant code snippet](https://github.com/0xPolygonMiden/air-script/blob/next/ir/src/constraint_builder/evaluators.rs#L93).

constraints/miden-vm/chiplets.air
```diff
@@ -0,0 +1,40 @@
...
+    # only change from 0 to 1.
+    enf s[0]' = s[0] when s[0]
+    enf s[1]' = s[1] when s[0] & s[1]
+    enf s[2]' = s[2] when s[0] & s[1] & s[2]
+
+
+### Chiplets Constraints ##########################################################################
+
+ev chiplets_constraints(main: [s[3], chiplet_columns[15]]):
+    enf chiplet_selector_constraints([s])
```
grjte, line 38: nit: I would do a bit of renaming for more consistency with our other files

1. `chiplet_selector_constraints` -> `chiplet_selectors`
2. `chiplet_constraints` -> deleted
3. `chiplets_constraints` -> `chiplets`

grjte (changes requested):
It all looks correct now @Overcastan - I would just simplify a bit more and do some minor renaming as noted inline

constraints/miden-vm/chiplets.air
```diff
@@ -0,0 +1,34 @@
...
+use memory::memory_chiplet
+
+### Helper evaluators #############################################################################
+
+# Enforces that the provided columns must be binary.
+ev is_binary(main: [a]):
+    enf a^2 = a
+
+# Enforces that the chiplet selector columns are set correctly.
+enf chiplet_selectors(main: [s[3]]):
```
grjte, line 14: This should be `ev` not `enf`

constraints/miden-vm/chiplets.air
```diff
@@ -0,0 +1,34 @@
...
+    # Enforce that the chiplets are stacked correctly by restricting selector values so they can 
+    # only change from 0 to 1.
+    enf s[0]' = s[0] when s[0]
+    enf s[1]' = s[1] when s[0] & s[1]
+    enf s[2]' = s[2] when s[0] & s[1] & s[2]
+
+
+### Chiplets Constraints ##########################################################################
+
+ev chiplets(main: [s[3], chiplet_columns[15]]):
```
grjte, line 29: nit: it would be good to have a comment here.

grjte (changes requested):
Thanks @Overcastan, there's an error I just noticed, otherwise this is nearly ready.

grjte (approved):
Looks good, thank you!

PR #254, opened 2023-04-14
refactor: change access types to BindingAccess

grjte:
Currently, there are 5 different variants of `Expression` into which references to bound identifiers within variables or constraints are parsed - `Elem`, `VectorAccess`, `MatrixAccess`, `Rand`, and `TraceBindingAccess`. These are all handled in essentially the same way within the IR, so consolidating these into a single type allows us to simplify handling in the IR, especially of variables.

This PR does the following. Because of the nature of the change, it's hard to split it into a smaller change without errors.

- introduce `BindingAccess`
- replace `Elem`, `VectorAccess`, and `MatrixAccess` with `BindingAccess`
- update codegen handling for new `Value` types

Future items for discussion:
- `Expression::Rand` should probably be similarly replaced. This would be fairly straightforward.
- Ideally, we would consolidate `TraceBindingAccess` into `BindingAccess` as well. This would improve the code clarity significantly, since all trace bindings for the "current" row are already parsed as `BindingAccess`, so `TraceBindingAccess` only represents references to trace bindings in the "next" row or being accessed as a `Range`. This means that the handling within the IR of accessing trace bindings is split into 2 places. The complication with this is that we access ranges of the trace, so `AccessType` would need to change, and I'm not sure of the best way to adjust this.

### DIFF
--- a/air-script-core/src/expression.rs
+++ b/air-script-core/src/expression.rs
@@ -1,17 +1,11 @@
-use super::{Identifier, ListFolding, MatrixAccess, TraceAccess, TraceBindingAccess, VectorAccess};
+use super::{BindingAccess, Identifier, ListFolding, TraceAccess, TraceBindingAccess};
 
 /// Arithmetic expressions for evaluation of constraints.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub enum Expression {
     Const(u64),
-    /// Represents any named constant or variable.
-    Elem(Identifier),
-    /// Represents an element inside a constant or variable vector. [VectorAccess] contains the
-    /// name of the vector and the index of the element to access.
-    VectorAccess(VectorAccess),
-    /// Represents an element inside a constant or variable matrix. [MatrixAccess] contains the
-    /// name of the matrix and indices of the element to access.
-    MatrixAccess(MatrixAccess),
+    /// Represents a reference to all or part of a constant, variable, or trace binding.
+    BindingAccess(BindingAccess),
     TraceAccess(TraceAccess),
     TraceBindingAccess(TraceBindingAccess),
     /// Represents a random value provided by the verifier. The first inner value is the name of
--- a/ir/src/constraint_builder/expression.rs
+++ b/ir/src/constraint_builder/expression.rs
@@ -1,5 +1,5 @@
 use super::{
-    get_variable_expr, AccessType, ConstantValue, ConstraintBuilder, Expression, ListFolding,
+    get_variable_expr, AccessType, BindingAccess, ConstraintBuilder, Expression, ListFolding,
     NodeIndex, Operation, SemanticError, SymbolType, TraceAccess, TraceBindingAccess, Value,
 };
 
@@ -27,19 +27,12 @@ impl ConstraintBuilder {
             }
 
             // --- IDENTIFIER EXPRESSIONS ---------------------------------------------------------
-            Expression::Elem(ident) => self.insert_symbol_access(ident.name(), AccessType::Default),
+            Expression::BindingAccess(access) => self.insert_symbol_access(access),
             Expression::Rand(ident, index) => {
+                // TODO: replace Rand with BindingAccess in parser?
                 let access_type = AccessType::Vector(index);
-                self.insert_symbol_access(ident.name(), access_type)
-            }
-            Expression::VectorAccess(vector_access) => {
-                let access_type = AccessType::Vector(vector_access.idx());
-                self.insert_symbol_access(vector_access.name(), access_type)
-            }
-            Expression::MatrixAccess(matrix_access) => {
-                let access_type =
-                    AccessType::Matrix(matrix_access.row_idx(), matrix_access.col_idx());
-                self.insert_symbol_access(matrix_access.name(), access_type)
+                let access = BindingAccess::new(ident, access_type);
+                self.insert_symbol_access(access)
             }
             Expression::ListFolding(lf_type) => self.insert_list_folding(lf_type),
 
@@ -77,9 +70,7 @@ impl ConstraintBuilder {
     /// Inserts the specified constant value into the graph and returns the resulting expression
     /// details.
     fn insert_inline_constant(&mut self, value: u64) -> Result<NodeIndex, SemanticError> {
-        let node_index = self.insert_graph_node(Operation::Value(Value::Constant(
-            ConstantValue::Inline(value),
-        )));
+        let node_index = self.insert_graph_node(Operation::Value(Value::InlineConstant(value)));
 
         Ok(node_index)
     }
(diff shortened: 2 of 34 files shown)


ir/src/constraint_builder/expression.rs
```diff
@@ -139,27 +130,28 @@ impl ConstraintBuilder {
...
     /// # Errors
     /// Returns an error if the identifier is not present in the symbol table or is not a supported
     /// type.
-    fn insert_symbol_access(
+    fn insert_binding_access(
         &mut self,
-        name: &str,
-        access_type: AccessType,
+        binding_access: BindingAccess,
     ) -> Result<NodeIndex, SemanticError> {
```
bobbinth, line 141: Following my prior comments, I'd probably keep the old name and change the signature to look something like:
```Rust
fn insert_symbol_access(
    &mut self,
    symbol_access: SymbolAccess
) -> Result<NodeIndex, SemanticError>
```

ir/src/constraint_builder/variables.rs
```diff
@@ -1,73 +1,174 @@
...
+        // access into the expression at the specified index
+        AccessType::Vector(new_idx) => match expr {
+            Expression::BindingAccess(inner_binding) => match inner_binding.access_type() {
+                AccessType::Default => {
+                    let new_binding_access = BindingAccess::new(
+                        inner_binding.ident().clone(),
+                        AccessType::Vector(new_idx),
+                    );
+                    Ok(Expression::BindingAccess(new_binding_access))
+                }
```
bobbinth, line 129: I got lost in this a little. Does this mean that if we have something like:
```
let x = [1, 2, 3]
let c = x[2]
```
We replace `c = x[2]` with `c = 3`?

ir/src/constraint_builder/variables.rs
```diff
@@ -1,73 +1,174 @@
...
+        },
+        // access into the expression at the specified row and column indices
+        AccessType::Matrix(row_idx, col_idx) => match expr {
+            Expression::BindingAccess(inner_binding) => match inner_binding.access_type() {
+                AccessType::Default => {
+                    let new_binding_access = BindingAccess::new(
+                        inner_binding.ident().clone(),
+                        AccessType::Matrix(row_idx, col_idx),
+                    );
+                    Ok(Expression::BindingAccess(new_binding_access))
```
bobbinth, line 158: Similarly, I'm not sure I understood what's happening here.

ir/src/symbol_table/symbol_type.rs
```diff
@@ -4,7 +4,7 @@ use std::fmt::Display;
 #[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
 pub(crate) enum SymbolType {
     /// an identifier for a constant, containing its type and value
-    Constant(ConstantValueExpr),
+    ConstantBinding(ConstantValueExpr),
     /// an identifier for a binding to one or more trace columns, containing the trace binding
     /// information with its identifier, trace segment, size, and offset.
     TraceBinding(TraceBinding),
```
bobbinth, line 10: I'm wondering if a better name for `SymbolType` is actually `SymbolBinding` (since it actually defines what the symbol is bound to). I would potentially redefine `Symbol` struct as:
```Rust
pub(crate) struct Symbol {
    name: String,
    binding: SymbolBinding,
}
```
And then rename `SymbolType` into `SymbolBinding` and keep the enum variant names simple. Maybe something like:
```Rust
pub(crate) enum SymbolType {
    Constant(ConstantValueExpr),
    Trace(TraceBinding),
    PublicInput(usize),
    PeriodicColumn(usize, usize),
    Variable(VariableValueExpr),
    RandomValues(usize, usize),
}
```

bobbinth (commented):
Thank you! I think this improves things quite a lot! I left some thoughts/comments inline. It is possible that I didn't consider all the implications for some of the - so, let me know if I missed anything.

For some of them, it may make sense to address in future PRs not to complicate this PR even further.

ir/src/constraint_builder/variables.rs
```diff
@@ -1,73 +1,174 @@
...
+
+/// Accesses into a `BindingAccess` expression and returns a new `BindingAccess` of a higher
+/// dimension.
+///
+/// For example:
+/// - suppose the `expr` is a [BindingAccess] specifying that a binding `A` is being accessed as a
+///   vector at index 0 (i.e. the access_type is [AccessType::Vector(0)], representing `A[0]`),
+/// - and suppose the specified `access_type` is [AccessType::Vector(i)].
+/// - Then the resulting expression would be A[i][0], represented by a new [BindingAccess] with
+///   identifier `A` and [AccessType::Matrix(0, i)].
```
tohrnii, line 108: very minor nit: I'm not sure if these should be bullet points.

tohrnii (approved):
Thank you @grjte. I don't have much to add other than what @bobbinth pointed out. I just left a couple of very minor optional nits.

air-script-core/src/expression.rs
```diff
@@ -1,17 +1,11 @@
...
-    /// name of the vector and the index of the element to access.
-    VectorAccess(VectorAccess),
-    /// Represents an element inside a constant or variable matrix. [MatrixAccess] contains the
-    /// name of the matrix and indices of the element to access.
-    MatrixAccess(MatrixAccess),
+    /// Represents a reference to all or part of a constant, variable, or trace binding.
+    BindingAccess(BindingAccess),
     TraceAccess(TraceAccess),
     TraceBindingAccess(TraceBindingAccess),
     /// Represents a random value provided by the verifier. The first inner value is the name of
```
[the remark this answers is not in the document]
grjte: - I'm fine with renaming `BindingAccess` to `SymbolAccess`, but I think it would be cleaner to do it in a separate PR.
- I agree that I'd like to remove `Rand` and `TraceBindingAccess` and just use `BindingAccess` (or `SymbolAccess`), but I think this will require a lot of changes and should be addressed in a separate PR as well. Adding `offset` to the `BindingAccess` struct should resolve the complication I mentioned in the PR description.
- We shouldn't remove `TraceAccess`, as it serves a different function. It is the type used in Value nodes which are leaves in the `AlgebraicGraph`. However, we can remove it from use in the parser so it is only used in the IR. I would also do this in a separate PR.

ir/src/constraint_builder/variables.rs
grjte: No, it means we replace `c` where it is used by `x[2]`. In other words, if we had something like the following:

```
let x = [1, 2, 3]
let c = x[2]
enf c = 3
```

this step  turns it into:

```
let x = [1, 2, 3]
let c = x[2]
enf x[2] = 3
```

Then a subsequent processing turns it into `enf 3 = 3`

grjte:
Thank you both for the review! I have addressed the smaller points. I agree with the larger points, and appreciate the suggestions for addressing the consolidation of `Rand` and `TraceBindingAccess` into `BindingAccess`. As these adjustments will result in significant changes, I'm going to leave them for future refactoring PRs. I'm planning the following series of PRs:

- simple renaming (e.g. `BindingAccess` to `SymbolAccess` and `SymbolType` to `SymbolBinding`)
- replace `Rand`
- replace `TraceBindingAccess`
- revisit the variable handling for improving clarity again (will probably be required as part of replacing `TraceBindingAccess`)

PR #256, opened 2023-04-17
refactor: replace Rand with SymbolAccess

grjte:
This is a minimal refactor to remove `Rand` and replace it with `SymbolAccess`. We may want to do this a bit more cleanly in the parser, but that work will coincide with consolidating `TraceBindingAccess` into `SymbolAccess`, so I wanted to keep the changeset small for now.


bobbinth (approved):
Looks good! Thank you!

PR #257, opened 2023-04-18
feat(constraints): add rpo constraints

tohrnii:


constraints/miden-vm/hash.air
```diff
@@ -181,6 +225,33 @@ ev hasher_state(main: [s[3], h[12], i]):
         is_unchanged(h[j + 4]) for j in 0..4 when !b & f_absorb_node
         h[j + 8]' = h[j + 4] for j in 0..4 when b & f_absorb_node
 
+ev enforce_rpo_round(main: [h[12]]):
```
Overcastan, line 228: We don't need segment names any more, so here and in the other evaluators we should get rid of them. 

But maybe it will be better to remove them in a future PR, I'm not sure.

Overcastan (changes requested):
Thank you @tohrnii, looks good! I left just few small nits.

Overcastan:
About your question: in my opinion it will be indeed better to move RPO constraints to the separate file, since they have a lot of data that is needed just for RPO and require only `h` columns to be evaluated.

constraints/miden-vm/hash.air
tohrnii: Done in this PR.

Overcastan (approved):
Looks great to me, thank you @tohrnii!

constraints/miden-vm/rpo.air
```diff
@@ -0,0 +1,81 @@
...
+    
+    # apply mds
+    let step1_with_mds = apply_mds(step1_with_sbox)
+    
+    # add constants
+    let step1 = [step1_with_mds[i] + ark[i + 12] for i in 0..12]
+
+    # compute the state that should result from applying the inverse of the last operation of the
+    # RPO round to the next step of the computation.
+    let step2 = apply_sbox(h)
```
bobbinth, line 78: I believe the S-box needs to be applied to the next state, rather than the current state. So, this should be something like:
```
let step2 = apply_sbox(h')
```
But I'm not sure we support this syntax (and not sure we should). Alternatively, we could change this to:
```
let step2 = [s'^7 for s in h]
```
And then we could probably git rid of `apply_sbox` function and replace line 68 with:
```
let step1_with_sbox = [s^7 for s in step1_with_constants]
```

tohrnii: Thanks, fixed.

constraints/miden-vm/rpo.air
```diff
@@ -0,0 +1,80 @@
...
+    ark_17: [11614812331536766976, 14172757457833930752, 2054001340201038848, 17000774922218162176, 14619614108529063936, 7744142531772273664, 3485239601103661568, 0]
+    ark_18: [14871063686742261760, 707573103686350208, 12420704059284934656, 13533282547195531264, 1394813199588124416, 16135070735728404480, 9755891797164034048, 0]
+    ark_19: [10148237148793042944, 15453217512188186624, 355990932618543744, 525402848358706240, 4635111139507788800, 12290902521256030208, 15218148195153268736, 0]
+    ark_20: [4457428952329675776, 219777875004506016, 9071225051243524096, 16987541523062161408, 16217473952264204288, 12059913662657710080, 16460604813734957056, 0]
+    ark_21: [15590786458219171840, 17876696346199468032, 12766199826003447808, 5466806524462796800, 10782018226466330624, 16456018495793752064, 9643968136937730048, 0]
+    ark_22: [10063319113072093184, 17731621626449383424, 9045979173463557120, 14512769585918244864, 6844229992533661696, 4571485474751953408, 3611348709641382912, 0]
+    ark_23: [14200078843431360512, 2897136237748376064, 12934431667190679552, 10973956031244050432, 7446486531695178752, 17200392109565784064, 18256379591337758720, 0]
```
grjte, line 43: These round constants [don't look right](https://github.com/0xPolygonMiden/crypto/blob/52db23cd423e01a92eaf92abfd25aeef5b882c92/src/hash/rpo/mod.rs#L645). Also, I think for clarity it would be good to use names like `ark1_0`, ..., `ark1_11` and `ark2_0`, ..., `ark2_11`

constraints/miden-vm/rpo.air
```diff
@@ -0,0 +1,80 @@
...
+    [7, 6, 22, 21, 8, 7, 23, 8, 26, 13, 10, 9],
+    [9, 7, 6, 22, 21, 8, 7, 23, 8, 26, 13, 10],
+    [10, 9, 7, 6, 22, 21, 8, 7, 23, 8, 26, 13],
+    [13, 10, 9, 7, 6, 22, 21, 8, 7, 23, 8, 26],
+    [26, 13, 10, 9, 7, 6, 22, 21, 8, 7, 23, 8],
+    [8, 26, 13, 10, 9, 7, 6, 22, 21, 8, 7, 23],
+    [23, 8, 26, 13, 10, 9, 7, 6, 22, 21, 8, 7]
+]
+
+periodic_columns:
```
grjte, line 19: let's add a comment to explain the round constants and how they are laid out here

constraints/miden-vm/rpo.air
```diff
@@ -0,0 +1,80 @@
...
+
+fn apply_mds(state: vector[12]) -> vector[12]:
+    return [sum([state[i] * mds_row[i] for i in 0..12]) for mds_row in MDS]
+
+### RPO Air Constraints ##################################################################
+
+ev enforce_rpo_round([h[12]]):
+    let ark = [ark_0, ark_1, ark_2, ark_3, ark_4, ark_5, ark_6, ark_7, ark_8, ark_9, ark_10,
+        ark_11, ark_12, ark_13, ark_14, ark_15, ark_16, ark_17, ark_18, ark_19, ark_20, ark_21,
+        ark_22, ark_23]
```
grjte, line 55: For clarity, let's define a constant for `STATE_WIDTH` and use it everywhere instead of 12. I think it might also be clearer if we defined separate variables for `ark1` and `ark2`. We couldn't do that in the Rust implementation because of the way we get periodic values from Winterfell, but we can do it here.

constraints/miden-vm/rpo.air
grjte:
```
let step2 = apply_sbox(h')
```

@bobbinth with the recent refactor we can support this, and we have used it in other places. It is because the argument passed to the function call is now a `SymbolAccess`, which includes an offset specification. Why do you not think we should? I actually find this clearer to read.

grjte (changes requested):
Thanks @tohrnii, looks good overall. I left a few things inline

constraints/miden-vm/rpo.air
grjte:
```
let step2 = apply_sbox(h')
```

@bobbinth with the recent refactor we can support this, and we have used it in other places. It is because the argument passed to the function call is now a `SymbolAccess`, which includes an offset specification. Why do you not think we should? I actually find this clearer to read.

constraints/miden-vm/rpo.air
tohrnii: > These round constants don't look right

Can you please elaborate what you mean?

constraints/miden-vm/hash.air
```diff
@@ -198,7 +196,8 @@ ev hash_chiplet(main: [s[3], r, h[12], i]):
     enf node_index([s, i])
 
     ## Hasher state constraints ##
-    # TODO: apply RPO constraints to the hasher state
+    enf enforce_rpo_round([h]) when !cycle_row_0
```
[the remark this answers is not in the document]
tohrnii: You're right. Thanks, fixed.

constraints/miden-vm/rpo.air
bobbinth: My thinking was that it might introduce quite a bit of complexity - but if this is not too difficult to support, then it is fine.

constraints/miden-vm/hash.air
```diff
@@ -83,21 +83,20 @@ fn get_f_out(s: vector[3]) -> scalar:
 fn get_f_out_next(s: vector[3]) -> scalar:
     return cycle_row_6 & binary_not(s[0]') & binary_not(s[1]')
 
-
```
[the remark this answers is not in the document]
Overcastan: I fixed it in the hasher multiset check constraints PR ([here](https://github.com/0xPolygonMiden/air-script/pull/261))

Al-Kindi-0:
Superseded by work in #455

PR #258, opened 2023-04-18
replace `TraceBindingAccess` with `SymbolAccess`

grjte:
This PR consolidates `TraceBindingAccess` into `SymbolAccess`. It aims to be as straightforward as possible without losing correctness, which means that some things are not fully addressed yet.

- replaces `TraceBindingAccess` with `SymbolAccess` in the parser/IR/codegen
- simplifies access errors in the IR, since some changes were required by the `SymbolAccess` handling changes anyway

Remaining items to address:
- remove `TraceAccess` usage from parser. This will require some cleanup in the parser, so it will be done in a separate PR
- handling of `Slice` access type. Previously this was only allowed for trace columns. I don't think there's a good reason not to allow this for variables/constants, but it needs to be done carefully. Similarly, within the `ListComprehension` we need to be careful about this, but because the `ListComprehension` is still quite complicated I think this should be revisited separately.
- revisit expression handling to see if it can be further simplified
- revisit list comprehension/folding to see if they can be simplified
- once the IR changes are done, we need to do a thorough analysis to make sure that all required errors are being caught. I think it's likely there are things missing here or that things which were previously enforced by the parser are now being left for the IR (but not yet handled)


ir/src/symbol_table/symbol_binding.rs
```diff
@@ -23,13 +23,13 @@ pub(crate) enum SymbolBinding {
...
-                write!(f, "Trace in segment {}", binding.trace_segment())
+            Self::Constant(_) => write!(f, "ConstantBinding"),
+            Self::Trace(_) => {
+                write!(f, "TraceBinding")
             }
             Self::PublicInput(_) => write!(f, "PublicInput"),
             Self::PeriodicColumn(_, _) => write!(f, "PeriodicColumn"),
-            Self::Variable(_) => write!(f, "Variable"),
+            Self::Variable(_) => write!(f, "VariableBinding"),
             Self::RandomValues(_, _) => write!(f, "RandomValues"),
```
bobbinth, line 33: Question: is there a reason why we sometimes add "Binding" at the end and sometimes don't?

ir/src/symbol_table/symbol.rs
```diff
@@ -49,121 +49,189 @@ impl Symbol {
...
         match access_type {
             AccessType::Default => {
                 if binding.size() != 1 {
-                    return Err(SemanticError::invalid_trace_binding_access(self.name()));
+                    return Err(SemanticError::invalid_access_type(self, &access_type));
                 }
                 let trace_segment = binding.trace_segment();
                 let trace_access =
-                    TraceAccess::new(trace_segment, binding.offset(), binding.size(), row_offset);
+                    TraceAccess::new(trace_segment, binding.offset(), binding.size(), offset);
```
bobbinth, line 221: Two different offsets are a bit confusing here. Maybe renaming `offset` variable here into `row_offset` would help a bit? Though it still may not be clear that `binding.offset()` is a column offset.

ir/src/symbol_table/mod.rs
```diff
@@ -210,28 +206,49 @@ impl SymbolTable {
...
+                };
+                Ok(TraceAccess::new(
+                    columns.trace_segment(),
+                    col_offset,
+                    1,
+                    symbol_access.offset(),
+                ))
+            }
+            _ => Err(SemanticError::not_a_trace_column_identifier(symbol)),
+        }
```
bobbinth, line 251: nit: I wonder if it would be possible to reduce nesting here a bit. Since we are interested only in a single variant, maybe we can do something like this upfront:
```Rust
let columns = match symbol.binding() {
    SymbolBinding::Trace(columns) => columns,
    _ => return Err(SemanticError::not_a_trace_column_identifier(symbol)),
};
```

bobbinth (approved):
Looks good! Thank you! I left a couple of non-blocking comments inline.

air-script-core/src/access.rs
```diff
@@ -22,8 +23,9 @@ impl Display for AccessType {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Self::Default => write!(f, "direct reference by name"),
-            Self::Vector(_) => write!(f, "vector"),
-            Self::Matrix(_, _) => write!(f, "matrix"),
+            Self::Slice(range) => write!(f, "slice at range {range:?}"),
```
tohrnii, line 26: minor nit: should it be "in range"? And we can also use display here instead.

tohrnii (approved):
Thanks @grjte, looks great.

Overcastan (approved):

ir/src/symbol_table/symbol_binding.rs
grjte: I think it's clearer in the errors if it has binding in it, but we don't ever refer to the other ones as bindings (even though they are).

PR #259, opened 2023-04-18
refactor: remove TraceAccess from the parser

grjte:
This PR removes references to `TraceAccess` from the parser. 

## Note:
An interesting side effect of this is that there is now an easy way we could approach supporting additional trace segments. Instead of having "main" and "aux" baked in as keywords, we could have the order of rows represent the trace segments and let the names associated with the segments be anything just like the random values array can be. This would allow us to generalize the grammar a bit. It would look something like this:

```
trace_columns:
    segment_0: [a, b, c]
    segment_1: [p0, p1]
    segment_2: [x, y, z]
```

and of course in the above we could still use `main` and `aux` instead of `segment_0` and `segment_1` respectively.

Interested to hear thoughts on this.


bobbinth:
> An interesting side effect of this is that there is now an easy way we could approach supporting additional trace segments. Instead of having "main" and "aux" baked in as keywords, we could have the order of rows represent the trace segments and let the names associated with the segments be anything just like the random values array can be.

Yes, that's how I was thinking about it as well. We should probably remove requirement that `main` and `aux` are keywords at some point and just rely on ordering. The only part which I'm not sure here about is handling of evaluator parameters.

parser/src/parser/grammar.lalrpop
```diff
@@ -515,8 +517,12 @@ Range: Range = {
     <start: Num_u64> ".." <end: Num_u64> => Range::new(start as usize, end as usize)
 }
 
-RandArrayAccess: Identifier = {
-    "$" <n:identifier> => Identifier(format!("${n}"))
+DeclIdentRef: Identifier = {
+    <n:decl_ident_ref> => Identifier(n.to_string())
```
tohrnii, line 521: minor nit: I was wondering if we should change the name from `n` here and elsewhere to something else because it sounds like it could be referring to a number. In some places we use `ident`, so we could use that. This could be done in another PR but I was just thinking about changing this earlier but forgot to.

tohrnii (approved):
Thanks @grjte, looks good to me.

bobbinth (approved):
Looks good! Thank you!

One question: should this also remove `TraceAccess` from `Expression` enum (i.e., [here](https://github.com/0xPolygonMiden/air-script/blob/grjte-refactor-3/air-script-core/src/expression.rs#L5))? Or is this for a subsequent PR?

ir/src/symbol_table/mod.rs
```diff
@@ -103,20 +103,22 @@ impl SymbolTable {
...
-        let (name, num_values, bindings) = rand_values.into_parts();
+        let (num_values, bindings) = rand_values.into_parts();
 
-        let mut offset = 0;
-        // add the name of the random values array to the symbol table
-        self.insert_symbol(
-            format!("${name}"),
-            SymbolBinding::RandomValues(offset, num_values as usize),
-        )?;
+        println!("rand values: {:?}", &bindings);
```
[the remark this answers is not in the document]
grjte: Oops, I thought I managed to keep that out of the commit. Thanks!

parser/src/parser/grammar.lalrpop
grjte: Never mind, I changed it now. There were only a couple places :)

grjte:
> One question: should this also remove `TraceAccess` from `Expression` enum (i.e., [here](https://github.com/0xPolygonMiden/air-script/blob/grjte-refactor-3/air-script-core/src/expression.rs#L5))? Or is this for a subsequent PR?

I had been planning only to remove it from the parser for now, but this was a small change & the only thing remaining for this refactor, so I made the adjustment now

PR #260, opened 2023-04-20
Remove keywords from evaluator function declaration

grjte:
This PR removes the keywords `main` and `aux` from evaluator function declarations. Instead, it parses the trace segments in order and accepts empty trace segments as long as the final trace segment is not empty. This enables the definition of evaluator functions which only use the trace for the auxiliary segment, which completes support for #124 .

This fixes the error from the previous evaluator function implementation where evaluators could not accept only an auxiliary trace. It also enables unlimited trace segments. A check should be added to the IR to ensure that no columns are passed to the wrong trace segment. This will ensure that evaluator functions cannot use trace segments that haven't been defined.

Example of the syntax that is now supported:

```
ev foo([a,b,c], [p]):
  ...

ev bar([], [p]):
  ...

ev bat([a, b]):
  ...

enf foo([x, y, z], [p])
enf bar([], [p])
enf bat([x, y])
```

### DIFF
--- a/ir/src/tests/evaluators.rs
+++ b/ir/src/tests/evaluators.rs
@@ -74,7 +74,6 @@ fn evaluator_with_main_and_aux_cols() {
     assert!(result.is_ok());
 }
 
-#[ignore]
 #[test]
 fn ev_call_with_aux_only() {
     let source = "
@@ -92,7 +91,7 @@ fn ev_call_with_aux_only() {
         enf clk.first = 0
     
     integrity_constraints:
-        enf enforce_a([a, b])";
+        enf enforce_a([], [a, b])";
 
     let parsed = parse(source).expect("Parsing failed");
     let result = AirIR::new(parsed);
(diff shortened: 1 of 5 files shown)


bitwalker (approved):
LGTM. I do have one question though: why must the last trace segment specifically be non-empty? It seems to me that there must be at least one non-empty trace segment, but that it could be in any position in the argument list. I'm guessing that I'm missing something, but wanted to make sure I raised the question just in case.

tohrnii:
@bitwalker I think the idea is that evaluators don't need to take all the segments as arguments, so if the last segment passed has no columns relevant to the evaluator, there's no need to pass that segment in the first place.

tohrnii (approved):
Thanks @grjte, looks good to me.

bobbinth (approved):
Looks great! Thank you!

Overcastan (approved):

grjte:
> @bitwalker I think the idea is that evaluators don't need to take all the segments as arguments, so if the last segment passed has no columns relevant to the evaluator, there's no need to pass that segment in the first place.

@bitwalker yes, it's exactly as @tohrnii said. Additionally, at least one segment should be non-empty, so requiring that the final segment be non-empty kills 2 birds with one stone.

Of course, we could just allow both things and it would still work, because these evaluators wouldn't be able to do anything. They would essentially just be dead code.

Issue #200, opened 2023-03-07
chiplets: hasher (includes multiset checks for virtual table)


grjte:

Al-Kindi-0:
Superseded by work in #455

PR #261, opened 2023-04-20
Write hasher multiset check constraints in AirScript

Overcastan:
Addressing: #200
This PR adds constraints for the hash chiplet bus and hash chiplet sibling table.


constraints/miden-vm/hash.air
```diff
@@ -34,27 +35,38 @@ fn get_f_mv(s: vector[3]) -> scalar:
 fn get_f_mu(s: vector[3]) -> scalar:
     return cycle_row_0 & s[0] & s[1] & s[2]
 
+# Set to 1 when selector flags are (0,0,0) on rows which are 1 less than a multiple of 8. This is 
+# flag of the instruction that returns the result of the currently running computation.
```
grjte, line 39: nit: Since it's not exactly an instruction, I would change the second sentence to "This flag indicates that the result of the currently running computation is returned."

constraints/miden-vm/hash.air
```diff
@@ -34,27 +35,38 @@ fn get_f_mv(s: vector[3]) -> scalar:
...
+fn get_f_hout(s: vector[3]) -> scalar:
+    return cycle_row_7 & !s[0] & !s[1] & !s[2]
+
+
+# Set to 1 when selector flags are (0,0,1) on rows which are 1 less than a multiple of 8. This is 
+# flag of the instruction that returns the whole hasher state.
```
grjte, line 45: nit: I would change the second sentence to "This flag indicates that the entire hasher state is returned."

constraints/miden-vm/hash.air
```diff
@@ -65,23 +77,9 @@ fn get_f_mua(s: vector[3]) -> scalar:
...
-#
 # Flag f_out is set to 1 when either f_hout = 1 or f_sout = 1 in the current row.
 fn get_f_out(s: vector[3]) -> scalar:
-    return cycle_row_7 & binary_not(s[0]) & binary_not(s[1])
-
-
-# Flag f_out_next is set to 1 when either f_hout = 1 or f_sout = 1 in the next row.
-fn get_f_out_next(s: vector[3]) -> scalar:
-    return cycle_row_6 & binary_not(s[0]') & binary_not(s[1]')
+    return cycle_row_7 & !s[0] & !s[1]
```
grjte, line 82: We can't actually make this change because these flags use different periodic columns

constraints/miden-vm/hash.air
```diff
@@ -98,8 +96,10 @@ ev is_unchanged(main: [column]):
 
 # Enforce selector columns constraints
 ev selector_columns(main: [s[3]]):
+    let s_next = [s_i' for s_i in s]
+
```
grjte, line 100: This variable shouldn't be necessary - we can just pass `s'` instead.

constraints/miden-vm/hash.air
```diff
@@ -202,4 +202,67 @@ ev hash_chiplet(main: [s[3], r, h[12], i]):
     enf hasher_state([s, h, i])
 
     # Multiset check constraints
-    # TODO: Apply multiset check constraints
\ No newline at end of file
+    # TODO: Apply multiset check constraints
+
+
+# Enforce constraints for the hash chiplet bus.
+ev hash_chiplet_bus_constraints(main: [s[3], r, h[12], i], aux: [p0]):
```
grjte, line 209: nit: let's rename this to `chiplet_bus` as we did elsewhere. It's the same bus which is shared among the chiplets

constraints/miden-vm/hash.air
```diff
@@ -202,4 +202,67 @@ ev hash_chiplet(main: [s[3], r, h[12], i]):
...
+    let h_next = [h[i]' for i in 0..12]
```
grjte, line 210: This isn't necessary. We can just pass `h'`

constraints/miden-vm/hash.air
```diff
@@ -202,4 +202,67 @@ ev hash_chiplet(main: [s[3], r, h[12], i]):
...
+    # v_all represents the entire hasher state.
+    let v_all = v_h + v_a + v_b + v_c
+
+    # b is the value of the bit which is discarded during shift by one bit to the right.
+    let b = i - 2 * i'
+
+    # v_leaf represents the leaf of the path.
+    let v_leaf = v_h + (1 - b) * v_b + b * v_d
+
+    # v_abp represents delta between the last 8 elements of the hasher state during absorbing a new
```
grjte, line 238: nit: "absorption" instead of "absorbing"

constraints/miden-vm/hash.air
```diff
@@ -202,4 +202,67 @@ ev hash_chiplet(main: [s[3], r, h[12], i]):
...
+
+    # v_leaf represents the leaf of the path.
+    let v_leaf = v_h + (1 - b) * v_b + b * v_d
+
+    # v_abp represents delta between the last 8 elements of the hasher state during absorbing a new 
+    # set of elements into the state while computing a linear hash.
+    let v_abp = v_h + v_b_next + v_c_next - (v_b + v_c)
+
+    # v_res represents the second word of the hasher state.
+    let v_res = v_h + v_b
```
grjte, line 243: nit: This isn't quite the second word of the hasher state, since it includes the header. I would probably say something like: "v_res represents returning the hash result, taken from the second word of the hasher state"

constraints/miden-vm/hash.air
```diff
@@ -202,4 +202,67 @@ ev hash_chiplet(main: [s[3], r, h[12], i]):
...
+
+    # Enforce constraints for updating column p0.
+    p0' = p0 * ((get_f_bp(s) + get_f_sout(s)) * v_all + (get_f_mp(s) + get_f_mv(s) + get_f_mu(s)) * 
+        v_leaf + get_f_abp(s) * v_abp + get_f_hout(s) * v_res + 1 - 
+        (get_f_bp(s) + get_f_mp(s) + get_f_mv(s) + get_f_mu(s) + get_f_abp(s) + get_f_out(s)))
```
grjte, line 248: nit: almost all of these flags are computed twice, which I would probably avoid. It's not a big deal, though, since it will be optimized by the compiler anyway.

Let's also add a comment with the constraint degree

constraints/miden-vm/hash.air
```diff
@@ -202,4 +202,67 @@ ev hash_chiplet(main: [s[3], r, h[12], i]):
...
+
+
+# Enforce constraints for the hash chiplet sibling table.
```
grjte, line 251: nit: I would change this to: "Enforce constraints for the hash chiplet sibling table, which is used to keep track of sibling nodes used during Merkle root update computations." I would probably even add the next bit from the docs with the 2 rules that need to be enforced.

constraints/miden-vm/hash.air
```diff
@@ -202,4 +202,67 @@ ev hash_chiplet(main: [s[3], r, h[12], i]):
     enf hasher_state([s, h, i])
 
     # Multiset check constraints
-    # TODO: Apply multiset check constraints
\ No newline at end of file
+    # TODO: Apply multiset check constraints
+
```
grjte: 1. The new helper evaluators should go above in the helper evaluator section so that the `hash_chiplet` evaluator is still the last thing in the file as we've done in other places. 
2. We should add calls to the two new evaluators here and remove this TODO

constraints/miden-vm/hash.air
```diff
@@ -202,4 +202,67 @@ ev hash_chiplet(main: [s[3], r, h[12], i]):
...
+    let v_b = sum([$alpha[j + 4] * h[j] for j in 4..8])
+    let v_c = sum([$alpha[j + 4] * h[j] for j in 8..12])
+
+    # v_sibling represents an entry in the sibling table.
+    let v_sibling = $alpha[0] + $alpha[3] * i + b * v_b + (1 - b) * v_c
+
+    # Enforce constraint for updating p1.
+    enf p1' * ((get_f_mv(s) + get_f_mva(s)) * v_sibling + 1 - (get_f_mv(s) + get_f_mva(s))) = 
+        p1 * ((get_f_mu(s) + get_f_mua(s)) * v_sibling + 1 - (get_f_mu(s) + get_f_mua(s)))
```
grjte, line 264: Let's add a comment with the constraint degree

constraints/miden-vm/hash.air
```diff
@@ -202,4 +202,67 @@ ev hash_chiplet(main: [s[3], r, h[12], i]):
...
+
+    # Enforce that computation of the old Merkle root is immediately followed by the computation of 
+    # the new Merkle root.
+    enf (get_f_bp(s) + get_f_mp(s) + get_f_mv(s) * (1 - p1)) = 0
```
grjte, line 268: The paren is in the wrong place here. This should be:

```
enf (get_f_bp(s) + get_f_mp(s) + get_f_mv(s)) * (1 - p1) = 0
```

Let's also add a comment with the constraint degree.

grjte (changes requested):
Thanks @Overcastan! I left my comments inline

constraints/miden-vm/hash.air
```diff
@@ -110,23 +120,29 @@ ev selector_columns(main: [s[3]]):
...
 
     # Enforce that selector columns are binary.
+    # Constraint degree: 2
     enf is_binary([selector]) for selector in s
 
     # Enforce that unless f_out = 1 or f_out' = 1, the values in columns s[1] and s[2] are copied 
     # over to the nex row.
+    # Constraints degree: 7
     enf is_unchanged([s[1]]) when !f_out & !f_out_next
     enf is_unchanged([s[2]]) when !f_out & !f_out_next
```
Overcastan, line 130: I think there is a typo in the [docs](https://0xpolygonmiden.github.io/miden-vm/design/chiplets/hasher.html#selector-columns-constraints) regarding the degree of these constraints (imho it should be 7, not 9). The `f_out` flag has degree 3, not 4, since it is computed using only periodic column and two selectors (as opposed to `f_hout` and `f_sout` flags, which use three selectors)

constraints/miden-vm/hash.air
```diff
@@ -144,19 +160,24 @@ ev node_index(main: [s[3], i]):
     let f_an = f_mp + f_mv + f_mu + f_mpa + f_mva + f_mua
 
     # Enforce that b is binary only when a new node is absorbed into the hasher state.
+    # Constraint degree: 6
     enf f_an * (b^2 - b) = 0
 
     # Enforce that when a computation is finished i = 0.
+    # Constraint degree: 4
     enf f_out * i = 0
```
Overcastan, line 168: Similar comment to the one above: I think the degree of this constraint should be 4, not 5.

constraints/miden-vm/hash.air
```diff
@@ -260,9 +286,37 @@ ev hash_sibling_table_constraints(main: [s[3], r, h[12], i], aux: [p1]):
...
+### Hash Chiplet Air Constraints ##################################################################
+
+# Enforces the constraints on the hash chiplet, given the columns of the hash execution trace.
+#
+# Max constraint degree: 7
+ev hash_chiplet([s[3], r, h[12], i]):
+    ## Row address constraint ##
+    # TODO: Apply row address constraints:
+    # 1. Boundary constraint `enf r.first = 1`
+    # 2. Transition constraint. It requires chiplets module's selector flag s0.
```
Overcastan: I think we can add this transition constraint, I'm just not sure where we should place `chip_s0` chiplet module's selector flag -- to the first (`main`) segment, or to the second (`aux`).

Al-Kindi-0:
Superseded by work in #455

Issue #203, opened 2023-03-07
write MidenVM constraints for the chiplets bus (multiset check constraints)


grjte:

Al-Kindi-0:
Superseded by work in #455

PR #262, opened 2023-04-21
Write Bitwise chiplet bus constraints in AirScript

Overcastan:
Partially addressing: #203
This PR adds bitwise chiplet bus constraints and removes keywords from evaluator function declaration.

### DIFF
--- a/constraints/miden-vm/bitwise.air
+++ b/constraints/miden-vm/bitwise.air
@@ -1,31 +1,37 @@
 mod BitwiseAir
 
-### Declarations ##################################################################################
+### Constants and periodic columns ################################################################
 
 periodic_columns:
     k0: [1, 0, 0, 0, 0, 0, 0, 0]
     k1: [1, 1, 1, 1, 1, 1, 1, 0]
 
-### Helper functions/evaluators ###################################################################
 
-# Enforces that column must be binary.
-ev is_binary(main: [a]):
-    enf a^2 - a = 0
+### Helper functions ##############################################################################
 
 # Returns value aggregated from limbs.
 fn aggregate(limb: vector[4]) -> scalar:
     return sum([2^i * a for (i, a) in (0..4, limb)])
 
+
+### Helper evaluators #############################################################################
+
+# Enforces that column must be binary.
+ev is_binary([a]):
+    enf a^2 = a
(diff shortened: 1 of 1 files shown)


constraints/miden-vm/bitwise.air
```diff
@@ -70,10 +77,28 @@ ev output_aggregation(main: [s, a, b, a_limb[4], b_limb[4], zp, z]):
...
         z = zp * 16 + a_and_b when !s
 
+
+# Enforces the constraints for bitwise chiplet bus.
+ev bus_constraints([a, b, z], [b_chip]):
+    # u_i represents reduction of a, b abd z into a single value for row i.
+    let u_i = $alpha[1] * a + $alpha[2] * b + $alpha[3] * z
+
+    # m represents inversion of k1 periodic column.
+    let m = 1 - k1
```
grjte, line 87: nit: it might be clearer to use a `binary_not` function like we do in other modules

constraints/miden-vm/bitwise.air
```diff
@@ -70,10 +77,28 @@ ev output_aggregation(main: [s, a, b, a_limb[4], b_limb[4], zp, z]):
         z = zp * 16 + a_xor_b when s
         z = zp * 16 + a_and_b when !s
 
+
+# Enforces the constraints for bitwise chiplet bus.
+ev bus_constraints([a, b, z], [b_chip]):
```
grjte, line 82: A few small nits here:
1. We don't use "constraints" in the other names, so I'd probably prefer to just call this `bus` or `bitwise_chiplet_bus`
2. There's actually only one constraint, so the comment can say "Enforces the constraint for the bitwise chiplet bus"
3. Let's include the degree of the constraint in this comment as well.

grjte (changes requested):
Thanks @Overcastan! This looks great overall. I just left a few very minor nits

constraints/miden-vm/bitwise.air
tohrnii: I was thinking that now we have `not` operator, we can avoid defining `binary_not` functions here and in other constraints.

grjte (approved):
Looks good, thank you!

PR #263, opened 2023-04-21
Write Memory chiplet bus constraints in AirScript

Overcastan:
Partially addressing: #203 
This PR adds memory chiplet bus constraints and removes keywords from evaluator function declaration.


grjte (changes requested):
Thank you @Overcastan! I think this looks correct. I just left a few minor nits inline

grjte (approved):

PR #264, opened 2023-04-24
fn parser

tohrnii:


ir/src/constraint_builder/expression.rs
```diff
@@ -50,6 +50,7 @@ impl ConstraintBuilder {
                 Ok(node_index)
             }
             Expression::Exp(lhs, rhs) => self.insert_exp_op(*lhs, *rhs),
+            Expression::FunctionCall(_, _) => todo!()
```
grjte, line 53: Let's remove this and handle it in a future PR

grjte (commented):
Thanks @tohrnii, I've taken a look at everything. Overall, I think we should simplify things and take smaller steps to make this easier for ourselves.

A few general points:

- Let’s handle just function definitions first and do function calls in a separate PR. There’s some other complexity there, and I don’t think the way it’s being addressed here is the way we want to do it
- It looks like this is not up to date with the latest version. I would start a new branch and cherry-pick your `feat(parser): add support for functions` commit or start from scratch and just copy in the things that you definitely want to keep like tests, etc.
- Maybe we shouldn’t allow returning tuples for now, to simplify this and get a working version completed. However, if we _are_ going to allow returning multiple values (e.g. a tuple), then there should be tests for this to make it easier to reason about
- It seems like it would make our lives a lot easier if we only allowed returning a variable rather than allowing expressions. I don’t think there’s a problem with enforcing this, at least for a first version.
- In the grammar, I think we’ve complicated things a lot by separating `BoundaryExpr` from `IntegrityExpr`. This forces us to have separation of `BoundaryVariableValue` and `VariableValueExpr` (the equivalent, but for variables in integrity constraints). It’s nearly 70 lines of almost-duplicated code. The only difference is that in the integrity expressions we allow `&`, `|`, and `!`, and in the boundary constraints we don’t, and the IntegrityAtom contains `function_call`. I think we should unify all of these and find another way to deal with this. If we really don’t want to allow logical operators in boundary constraints (up for debate), then we should add new expression types for these, and then we can enforce in the IR that boundary constraints don’t contain them. Similar logic for handling the function calls, but again, I wouldn’t add anything related to function calls yet. This will make the grammar much simpler and easier to reason about, and we’d have just one way of parsing variables.


Given the number of changes I’ve suggested above, I recommend dropping the current changes in the grammar file to take a fresh crack at it the simplified version suggested above, starting with a refactor PR to consolidate the variables for boundary and integrity constraint sections as discussed above.

tohrnii:
Thanks @grjte, I'll open smaller PRs based on your suggestion.

> It seems like it would make our lives a lot easier if we only allowed returning a variable rather than allowing expressions. I don’t think there’s a problem with enforcing this, at least for a first version.

I agree. I was considering this as well to keep things simpler for first version.

> In the grammar, I think we’ve complicated things a lot by separating BoundaryExpr from IntegrityExpr. This forces us to have separation of BoundaryVariableValue and VariableValueExpr (the equivalent, but for variables in integrity constraints). It’s nearly 70 lines of almost-duplicated code. The only difference is that in the integrity expressions we allow &, |, and !, and in the boundary constraints we don’t, and the IntegrityAtom contains function_call. I think we should unify all of these and find another way to deal with this. If we really don’t want to allow logical operators in boundary constraints (up for debate), then we should add new expression types for these, and then we can enforce in the IR that boundary constraints don’t contain them. Similar logic for handling the function calls, but again, I wouldn’t add anything related to function calls yet. This will make the grammar much simpler and easier to reason about, and we’d have just one way of parsing variables.

I agree with unifying `BoundaryExpr` and `IntegrityExpr`. Maybe we should do this first before tackling this.

grjte:
> I agree with unifying BoundaryExpr and IntegrityExpr. Maybe we should do this first before tackling this and move validation to IR for things that they don't have in common.

Yes, I agree that this is the best first step, as I mentioned in my previous comment. I'd start with a refactor PR to consolidate the variables for boundary and integrity constraint sections as discussed above (adding error checks in the IR if needed). Let's do that before returning to any of the function-related changes

Issue #236, opened 2023-03-29
Refactor constraint comprehension


tohrnii:
1. Combine `Constraint` and `ConstraintComprehension` variants for both integrity and boundary constraints.
```
pub enum IntegrityStmt {
    Constraint(ConstraintType, Option<ConstraintComprehensionContext>, Option<Expression>),
    Variable(Variable),
}
```

2. Modify the ordering of the syntax for constraint comprehension with selectors. E.g.
```
enf x = a + b for x in c when s[0] & s[1]
```

_Originally proposed by @grjte in https://github.com/0xPolygonMiden/air-script/pull/189_

grjte:
Closed by #269

PR #269, opened 2023-04-25
refactor(parser): refactor constraint comprehension grammar

tohrnii:
Addressing #236.

- Combines `Constraint` and `ConstraintComprehension` variants.
- Re-orders constraint comprehension with selectors statements to place selectors after comprehension (i.e., `enf x = a + b for x in c when s[0] & s[1]`)

### DIFF
--- a/parser/src/ast/boundary_constraints.rs
+++ b/parser/src/ast/boundary_constraints.rs
@@ -8,7 +8,7 @@ use std::fmt::Display;
 
 #[derive(Debug, Eq, PartialEq)]
 pub enum BoundaryStmt {
-    Constraint(BoundaryConstraint, Option<ComprehensionContext>),
+    Constraint(BoundaryConstraint),
     VariableBinding(VariableBinding),
 }
 
@@ -18,14 +18,21 @@ pub struct BoundaryConstraint {
     access: SymbolAccess,
     boundary: Boundary,
     value: Expression,
+    comprehension_context: Option<ComprehensionContext>,
 }
 
 impl BoundaryConstraint {
-    pub fn new(access: SymbolAccess, boundary: Boundary, value: Expression) -> Self {
+    pub fn new(
+        access: SymbolAccess,
+        boundary: Boundary,
+        value: Expression,
+        comprehension_context: Option<ComprehensionContext>,
+    ) -> Self {
         Self {
             access,
             boundary,
             value,
+            comprehension_context,
         }
     }
 
@@ -42,8 +49,24 @@ impl BoundaryConstraint {
         &self.value
     }
 
-    pub fn into_parts(self) -> (Boundary, SymbolAccess, Expression) {
-        (self.boundary, self.access, self.value)
+    pub fn comprehension_context(&self) -> Option<&ComprehensionContext> {
+        self.comprehension_context.as_ref()
+    }
+
+    pub fn into_parts(
+        self,
+    ) -> (
+        Boundary,
+        SymbolAccess,
+        Expression,
+        Option<ComprehensionContext>,
+    ) {
+        (
+            self.boundary,
+            self.access,
+            self.value,
+            self.comprehension_context,
+        )
     }
 }
 
--- a/parser/src/ast/boundary_constraints.rs
+++ b/parser/src/ast/boundary_constraints.rs
@@ -8,8 +8,7 @@ use std::fmt::Display;
 
 #[derive(Debug, Eq, PartialEq)]
 pub enum BoundaryStmt {
-    Constraint(BoundaryConstraint),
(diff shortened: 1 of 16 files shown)


parser/src/ast/integrity_constraints.rs
```diff
@@ -7,8 +7,11 @@ use super::{EvaluatorFunctionCall, Expression, VariableBinding};
 
 #[derive(Debug, Eq, PartialEq)]
 pub enum IntegrityStmt {
-    Constraint(ConstraintType, Option<Expression>),
-    ConstraintComprehension(ConstraintType, Option<Expression>, ComprehensionContext),
+    Constraint(
+        ConstraintType,
+        Option<Expression>,
+        Option<ComprehensionContext>,
+    ),
```
grjte, line 14: I think maybe we should make a new struct to hold this. Maybe something like the following:

```
#[derive(Debug, Eq, PartialEq)]
pub enum IntegrityStmt {
    Constraint(IntegrityConstraint),
    VariableBinding(VariableBinding),
}

#[derive(Debug, Eq, PartialEq)]
pub struct IntegrityConstraint {
    constraint_expr: ConstraintExpr,
    selectors: Option<Expression>,
    comprehension_context: Option<ComprehensionContext>,
}

#[derive(Debug, Eq, PartialEq)]
enum ConstraintExpr {
    Inline(InlineConstraintExpr),
    Evaluator(EvaluatorFunctionCall),
}

/// Stores the expression corresponding to the integrity constraint.
#[derive(Debug, Eq, PartialEq, Clone)]
struct InlineConstraintExpr {
    lhs: Expression,
    rhs: Expression,
}
```

parser/src/ast/boundary_constraints.rs
```diff
@@ -8,8 +8,7 @@ use std::fmt::Display;
 
 #[derive(Debug, Eq, PartialEq)]
 pub enum BoundaryStmt {
-    Constraint(BoundaryConstraint),
-    ConstraintComprehension(BoundaryConstraint, ComprehensionContext),
+    Constraint(BoundaryConstraint, Option<ComprehensionContext>),
```
grjte, line 11: I think maybe we should put the comprehension context into the boundary constraint, like this:

```
#[derive(Debug, Eq, PartialEq)]
pub enum BoundaryStmt {
    Constraint(BoundaryConstraint),
    VariableBinding(VariableBinding),
}

/// Stores the expression corresponding to the boundary constraint.
#[derive(Debug, Eq, PartialEq)]
pub struct BoundaryConstraint {
    access: SymbolAccess,
    boundary: Boundary,
    value: Expression,
    comprehension_context: Option<ComprehensionContext>,
}
```

grjte (changes requested):
Thanks @tohrnii! I do prefer having just one `Constraint` variant in the `Stmt` enums, but I think we could simplify things more by reducing to just one contained item inside the `Constraint`. See my comments inline and lmk what you think

parser/src/ast/integrity_constraints.rs
tohrnii: I was considering doing this as this does look cleaner but decided against it since `IntegrityConstraint` seems like it should be a single constraint and constraint comprehension seems like syntactic sugar to write multiple constraints in a single statement. Although I don't have a strong opinion on this so happy to change it if you prefer what you suggested instead.

grjte: I think taking an approach like this will make the IR cleaner, so I would prefer to change it

>constraint comprehension seems like syntactic sugar to write multiple constraints in a single statement

Maybe this can be addressed just by using different names for the structs above. For example, instead of `IntegrityConstraint` it could be `ConstraintStmt` or something like that

tohrnii: I went with the structure you suggested.

grjte (approved):
Thanks @tohrnii, LGTM! Just one comment inline for an optional change, so I'll leave merging to you

parser/src/parser/grammar.lalrpop
```diff
@@ -316,29 +314,23 @@ IntegrityStmtGroup: Vec<IntegrityStmt> = {
...
-                return vec![IntegrityStmt::ConstraintComprehension(constraint_type, selectors, comprehension)],
-            _ => unreachable!() // this should never happen
-        }
 }
 
 IntegrityConstraintExpr: IntegrityStmt = {
+    <constraint: IntegrityConstraintExprValue>
+        <comprehension: ConstraintComprehension<IntegrityExpr>?>
+        <selectors: WithSelectors?> =>
+            IntegrityStmt::Constraint(IntegrityConstraint::new(constraint, selectors, comprehension)),
```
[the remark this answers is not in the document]
tohrnii: Made the change

PR #270, opened 2023-04-25
feat(ir): add support for selectors

tohrnii:
Partly addressing #125.
Adds support for selectors to IR.

Note: This doesn't implement selectors with evaluators yet. That will be done in a subsequent PR.


grjte:
@tohrnii this looks good to me overall, but changes to #269 will affect this, so I've put it into draft mode for now. Let's finish that one and then revisit this.

grjte (approved):
looks good to me, thanks!

PR #272, opened 2023-04-25
Write stack multiset check constraints in AirScript

Overcastan:
Partially addressing: #203
This PR adds multiset check constraints for stack operations and removes keywords from evaluator function declaration.


constraints/miden-vm/stack.air
```diff
@@ -203,461 +218,566 @@ fn compute_op_flags(op_bits: vector[8]) -> vector[88]:
...
+ev u32and_bus_constraint([stack_top[16]], [b_chip]):
+    # and_req_val represents bitwise operation request value.
+    let and_req_val = $alpha[0] + 
+                      $alpha[1] * BITWISE_AND + 
+                      $alpha[2] * stack_top[0] + 
+                      $alpha[3] * stack_top[1] + 
+                      $alpha[4] * stack_top[0]'
+
+    # Enforce u32and bus constraint.
+    enf b_chip' * and_req_val = b_chip
```
bobbinth, line 250: I don't think we can write bus constraints this way. This captures only one half of the constraint (the request part) but we need to capture both parts (request and response) in the same constraint.

What we should probably do is to create a separate file for each bus and describe bus constraints there fully.

Al-Kindi-0:
Superseded by work in #455

PR #276, opened 2023-04-26
Enum for section exhaustive check

hackaugusto:
An enum has a few advantages:

- it auto completes
- the compiler checks the pattern is exhaustive
- is more efficient (matching on a string does a `bcmp` for each arm, not super important here though)


grjte (approved):
Looks great, thank you!

PR #278, opened 2023-04-28
fix(ir): fix evaluator constraints insertion

tohrnii:


ir/src/constraint_builder/evaluators.rs
```diff
@@ -46,14 +46,18 @@ impl ConstraintBuilder {
...
             // extend the constraint graph with all of the nodes from the evaluator graph
             self.graph.extend(ev_call_graph);
 
             // add the roots of the evaluator function call's constraints
-            self.integrity_constraints.append(&mut constraints);
+            self.integrity_constraints
+                .resize_with(constraints.len(), Vec::new);
+            for (segment, segment_constraints) in constraints.iter().enumerate() {
+                self.integrity_constraints[segment].extend(segment_constraints.clone());
+            }
```
grjte, line 60: you should be able to use `into_iter` here and avoid cloning

grjte (changes requested):
Thanks for catching this @tohrnii. Just one comment

grjte (approved):
Looks good, thanks

Issue #280, opened 2023-04-28
Code generated for evaluator constraints is incorrect


tohrnii:
Consider the following constraints:

```
ev check_binary([x]):
    enf x^2 - x = 0
    
integrity_constraints:
    # Enforce that selector must be binary
    enf check_binary([s])

    # Enforce that selector should stay the same throughout the cycle.
    enf k1 * (s' - s) = 0

    # Enforce that the input is decomposed into valid bits
    enf check_binary([a_limb[0]])
    enf check_binary([a_limb[1]])
    enf check_binary([a_limb[2]])
    enf check_binary([a_limb[3]])
    enf check_binary([b_limb[0]])
    enf check_binary([b_limb[1]])
    enf check_binary([b_limb[2]])
    enf check_binary([b_limb[3]])
```

This should generate:

```
result[0] = main_current[0].exp(E::PositiveInteger::from(2_u64)) - main_current[0] - E::ZERO;
result[1] = periodic_values[1] * (main_next[0] - main_current[0]) - E::ZERO;
result[2] = main_current[3].exp(E::PositiveInteger::from(2_u64)) - main_current[3] - E::ZERO;
result[3] = main_current[4].exp(E::PositiveInteger::from(2_u64)) - main_current[4] - E::ZERO;
result[4] = main_current[5].exp(E::PositiveInteger::from(2_u64)) - main_current[5] - E::ZERO;
result[5] = main_current[6].exp(E::PositiveInteger::from(2_u64)) - main_current[6] - E::ZERO;
result[6] = main_current[7].exp(E::PositiveInteger::from(2_u64)) - main_current[7] - E::ZERO;
result[7] = main_current[8].exp(E::PositiveInteger::from(2_u64)) - main_current[8] - E::ZERO;
result[8] = main_current[9].exp(E::PositiveInteger::from(2_u64)) - main_current[9] - E::ZERO;
result[9] = main_current[10].exp(E::PositiveInteger::from(2_u64)) - main_current[10] - E::ZERO;
```

However, it generates:

```
result[0] = main_current[13] - E::ZERO - main_current[0];
result[1] = periodic_values[1] * (main_next[0] - main_current[0]) - E::ZERO;
result[2] = main_current[13] - E::ZERO - main_current[0];
result[3] = main_current[13] - E::ZERO - main_current[0];
result[4] = main_current[13] - E::ZERO - main_current[0];
result[5] = main_current[13] - E::ZERO - main_current[0];
[...]

PR #282, opened 2023-04-28
fix graph cloning when processing evaluator function calls

grjte:
This PR addresses #280. When evaluator graphs were being cloned, the node index references weren't being updated to reflect the offset of the node indices.


tohrnii (approved):
Thanks @grjte, looks good.

PR #283, opened 2023-04-28
feat(ir): add support for selectors with evaluators

tohrnii:
Partly addressing #125. Adds support for selectors with evaluator calls. Also includes a tiny fix for a bug introduced in a previous PR.

### DIFF
--- a/air-script/tests/selectors/selectors_with_evaluators.air
+++ b/air-script/tests/selectors/selectors_with_evaluators.air
@@ -1,0 +1,25 @@
+def SelectorsAir
+
+ev next_is_zero([clk]):
+    enf clk' = 0
+
+ev is_unchanged([clk, s]):
+    enf clk' = clk when s
+
+ev next_is_one([clk]):
+    enf clk' = 1
+
+trace_columns:
+    main: [s[3], clk]
+    
+public_inputs:
+    stack_inputs: [16]
+
+boundary_constraints:
+    enf clk.first = 0
+
+integrity_constraints:
+    enf next_is_zero([clk]) when s[0] & !s[1]
+    match enf:
+        is_unchanged([clk, s[0]]) when s[1] & s[2]
+        next_is_one([clk]) when !s[1] & !s[2]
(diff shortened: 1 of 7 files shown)


air-script/tests/selectors/selectors_with_evaluators.air
```diff
@@ -0,0 +1,25 @@
+def SelectorsAir
+
+ev is_next_zero([clk]):
+    enf clk' = 0
+
+ev is_unchanged([clk, s0, s1]):
+    enf clk' = clk when s0 & s1
+
+ev is_next_one([clk]):
+    enf clk' = 1
```
grjte, line 10: nit: I would do `next_is_*`

ir/src/constraint_builder/evaluators.rs
```diff
@@ -52,11 +54,25 @@ impl ConstraintBuilder {
...
-                .resize_with(constraints.len(), Vec::new);
             for (segment, segment_constraints) in constraints.into_iter().enumerate() {
-                self.integrity_constraints[segment].extend(segment_constraints);
+                if let Some(selectors) = selectors {
+                    for constraint in segment_constraints {
+                        let constraint_with_selectors = self
+                            .insert_graph_node(Operation::Mul(*constraint.node_index(), selectors));
+                        self.integrity_constraints[segment].push(ConstraintRoot::new(
+                            constraint_with_selectors,
+                            constraint.domain(),
```
grjte, line 70: This domain could be wrong, for example if the constraint's domain was previously `EveryRow` but one of the selectors is in the next row. We need to provide the domain for the newly created graph node

ir/src/tests/selectors.rs
```diff
@@ -62,3 +62,104 @@ fn multiconstraint_selectors() {
...
+        stack_inputs: [16]
+
+    boundary_constraints:
+        enf clk.first = 0
+
+    integrity_constraints:
+        enf is_next_zero([clk]) when s[0] & !s[1]
+        match enf:
+            is_unchanged([clk, s[0], s[1]])
+            is_next_one([clk]) when !s[0] & !s[1]";
```
grjte, line 160: The point of `match` as I understand it is to only enforce one of these things, depending on the selectors that are specified. Therefore it makes sense to use it when the selectors for different cases are mutually exclusive. In this case, the first evaluator function (which has no selectors) will always be enforced, and the second one never will be, so this doesn't really make sense.

grjte (approved):
Thanks @tohrnii, this looks good to me now, and I'm approving it.

The only thing is that I would really like to see a test where one of the selectors is in the next row as noted in my last review. Can you do a follow-up PR with this test?

Issue #281, opened 2023-04-28
Problem using column groups in evaluator signature


tohrnii:
Using column groups in evaluator signature returns an error.

```
ev evaluatorfn([selectors[3], clk]):
    enf (clk' - clk) * selectors[0] * selectors[1] * selectors[2] = 0
    
trace_columns:
    main: [s[3], clk]
    
public_inputs:
    stack_inputs: [16]
    
boundary_constraints:
    enf clk.first = 0
    
integrity_constraints:
    enf evaluatorfn([s, clk])
```

This panics with `index out of bounds: the len is 2 but the index is 3`

tohrnii:
Closed by #284

PR #284, opened 2023-04-28
Fix conversion of `SymbolAccess` to `TraceAccess` for multi-column references

grjte:
This addresses #281. It adds a new test and fixes the issue with converting column group `SymbolAccess` references to `TraceAccess`


tohrnii (approved):
Thanks @grjte, looks good.

PR #287, opened 2023-04-29
feat(ir): add support for constraint comprehension with evaluators

tohrnii:


ir/src/validation/error.rs
```diff
@@ -2,6 +2,7 @@ use super::{
     AccessType, ConstrainedBoundary, ConstraintDomain, Symbol, SymbolBinding, TraceAccess,
     TraceSegment, MIN_CYCLE_LENGTH,
 };
+use air_script_core::Expression;
```
grjte, line 5: nit: we should probably follow the same pattern we use elsewhere of importing from super

grjte (approved):
Thanks @tohrnii, this looks correct. I just left one minor nit inline. 

As a note - I find the method names a bit confusing, and the comprehension handling feels a bit complex in comparison to how a lot of the rest of the IR is handled now. I don't think we should make a change now, but I think it might be worth revisiting all of the comprehension handling to see if there's a way we can simplify/clarify a bit. I didn't make any significant updates to these sections during the recent refactors, so there may be opportunities for improvement. What do you think?

tohrnii:
> As a note - I find the method names a bit confusing, and the comprehension handling feels a bit complex in comparison to how a lot of the rest of the IR is handled now. I don't think we should make a change now, but I think it might be worth revisiting all of the comprehension handling to see if there's a way we can simplify/clarify a bit. I didn't make any significant updates to these sections during the recent refactors, so there may be opportunities for improvement. What do you think?

Makes sense. Added #292 to track this.

PR #288, opened 2023-04-29
fix(ir): throw an error on segment mismatch in evaluators

tohrnii:


ir/src/validation/error.rs
```diff
@@ -157,6 +157,15 @@ impl SemanticError {
         ))
     }
 
+    pub(crate) fn evaluator_trace_segment_mismatch(
+        trace_segment: u8,
+        required_segment: usize,
+    ) -> Self {
+        SemanticError::InvalidUsage(format!(
+            "Evaluator function expected column in trace segment {required_segment}, but was given column in trace segment {trace_segment}.",
```
grjte, line 165: nit: this should be `binding` instead of `column`, since it could actually be a group of multiple columns

ir/src/constraint_builder/evaluators.rs
```diff
@@ -78,11 +78,18 @@ impl ConstraintBuilder {
     ) -> Result<Vec<Vec<usize>>, SemanticError> {
         // get the offsets for the trace binding accesses as a vector of indices
         let mut offsets: Vec<Vec<usize>> = Vec::new();
-        for segment in args.iter() {
+        for (segment, segment_trace_binding_access) in args.iter().enumerate() {
```
grjte, line 81: nit: I would prefer the name `segment_binding_accesses` instead of `segment_trace_binding_access`, since it makes it clearer that it's the set of bindings for the segment

grjte (approved):
Looks good, thanks @tohrnii! I left 2 terminology/naming nits inline

PR #295, opened 2023-05-12
deps: structopt -> clap

hackaugusto:
`structopt` is no longer in active development and is in maintenance mode. The recommendation is to use `clap` instead [ref](https://github.com/TeXitoi/structopt#maintenance)

### DIFF
--- a/air-script/src/cli/transpile.rs
+++ b/air-script/src/cli/transpile.rs
@@ -1,3 +1,4 @@
+use clap::Args;
 use std::{fs, path::PathBuf, sync::Arc};
 
 use codegen_winter::CodeGenerator;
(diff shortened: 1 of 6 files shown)


tohrnii (approved):
Thanks. Looks good to me.

air-script/src/cli/transpile.rs
```diff
@@ -1,39 +1,29 @@
...
+#[derive(Args)]
+pub struct Transpile {
+    input: PathBuf,
+
+    #[arg(
+        short,
+        long,
+        help = "Output filename, defaults the input file with the .rs extension"
+    )]
+    output: Option<PathBuf>,
```
bobbinth, line 17: minor nit: the "help" on line 15 should probably say "... default to the ..." instead of "defaults the".

Also, it seems like we've lost some info here - i.e., `structop` would take doc comments and convert them into command hints (i.e., for the input file it would say "Path to input file" etc.). Could we not do something similar with `clap`?

hackaugusto: Updated it, now it looks like this:

```
AirScript language compiler

Usage: airc <COMMAND>

Commands:
  transpile
  help       Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
```

bobbinth: Could we also add the description of the `transpile` command? I think it was "Transpile AirScript source code to Rust targeting Winterfell" previously.

grjte (dismissed):
Looks good, thanks! The CLI portion of the AirScript docs should be updated as well. Also, I agree with Bobbin's comment about adding a description for the `transpile` command.

One other note - maybe we should make an issue to make the same change to the MidenVM CLI

air-script/src/cli/transpile.rs
hackaugusto: changed:

```
AirScript language compiler

Usage: airc <COMMAND>

Commands:
  transpile  Transpile AirScript source code to Rust targeting Winterfell
  help       Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
```

bobbinth (approved):
Looks good! Thank you!

PR #296, opened 2023-05-12
masm: initial implementation of the backend

hackaugusto:
- Initial implementation of a visitor pattern
- The visitor is used to walk through the AirIR, using the same order as currently used by the winterfell backend
- Masm is emitted in the walk order, which is enough to generate functional but inefficient code - Biggest source of inneficiencies is loading individual Felts from memory


codegen/masm/src/codegen.rs
```diff
@@ -0,0 +1,273 @@
...
+        &mut self,
+        _constraint: &'ast ConstraintRoot,
+        _trace_segment: u8,
+    ) -> Result<Self::Value, Self::Error> {
+        Ok(()) // TODO
+    }
+
+    fn visit_air(&mut self, ir: &'ast AirIR) -> Result<Self::Value, Self::Error> {
+        self.ir = Some(ir);
+        walk_constant_bindings(self, ir)?;
```
hackaugusto, line 82: I tried not to reinvent the wheel for this. I went over 3 approaches `rustc_ast`, `serde`, and the [rust patterns](https://rust-unofficial.github.io/patterns/patterns/behavioural/visitor.html). This is basically doing the same as those references.

codegen/masm/tests/test_basic_arithmetic.rs
```diff
@@ -0,0 +1,212 @@
...
+        Kernel::new(&[]),
+        StackInputs::new(vec![]),
+        AdviceInputs::default().into(),
+    );
+    let program_outputs = process.execute(&program).expect("execution failed");
+    let result_stack = program_outputs.stack();
+
+    // results are in stack-order
+    #[rustfmt::skip]
+    let expected = [
```
hackaugusto, line 65: The next thing to do is to load the delta from memory and multiply these elements to produce the numerator. Then add support for the aux (current code is assuming a Felt, supports for ext2 should be straight forward to add)

codegen/masm/src/codegen.rs
```diff
@@ -0,0 +1,273 @@
...
+};
+use std::borrow::Cow;
+use std::collections::btree_map::{BTreeMap, Entry};
+
+pub struct CodegenConfig {
+    pub ood_frame_address: u64,
+}
+
+pub struct Codegen<'ast> {
+    ins: Vec<Cow<'ast, str>>,
```
hackaugusto, line 19: Using `Vec<Cow<_>>` instead of `Vec<String>` because the majority of cases an allocation is not necessary. Using `Cow` allows to refer to the `&'static str` and go around memory allocations. This is neat since most of what this backend does is pushing strings to this vector.

codegen/masm/tests/test_basic_arithmetic.rs
```diff
@@ -0,0 +1,227 @@
...
+    let code = format!(
+        "
+        begin
+            push.{}          # frame address
+            push.0.0.{b}.{a} # data
+            dup.4 mem_storew # save data to memory
+            dropw drop       # clean stack
+
+            # [generated code]
+            {}
```
hackaugusto, line 61: The generated code isn't pretty:

```
            padw mem_loadw.4294965000 movdn.3 drop drop drop padw mem_loadw.4294965000 movdn.3 drop drop drop add push.0 sub padw mem_loadw.4294965000 movdn.3 drop drop drop padw mem_loadw.4294965000 movdn.3 drop drop drop sub push.0 sub padw mem_loadw.4294965000 movdn.3 drop drop drop padw mem_loadw.4294965000 movdn.3 drop drop drop mul push.0 sub padw mem_loadw.4294965000 movdn.3 drop drop drop exp.3 push.0 sub padw mem_loadw.4294965000 drop movdn.2 drop drop padw mem_loadw.4294965000 movdn.3 drop drop drop add push.0 sub padw mem_loadw.4294965000 drop movdn.2 drop drop padw mem_loadw.4294965000 movdn.3 drop drop drop sub push.0 sub padw mem_loadw.4294965000 drop movdn.2 drop drop padw mem_loadw.4294965000 movdn.3 drop drop drop mul push.0 sub
```

AFAICT there is no easy way to go from constraint root to the air expression, which would be nice to emit as comments. There is also no easy way to go backwards from column index to variable name, so the results ends up being the blob above.

Each value is loaded individually since it was agreed at this stage we don't want to have optimizations (and that would need something akin to the middle IR, which we decided not to do).

codegen/masm/src/codegen.rs
```diff
@@ -0,0 +1,273 @@
...
+                        self.ins.push("drop".into());
+                        self.ins.push("swap".into());
+                        self.ins.push("drop".into());
+                    }
+                    (MAIN_TRACE, 3) => {
+                        self.ins.push("drop".into());
+                        self.ins.push("drop".into());
+                        self.ins.push("drop".into());
+                    }
+                    _ => {
```
hackaugusto, line 261: question: is every trace besides `main` using `ext2`? (as in, should I use `(AUX_TRACE, ...)` or `(_, ...)` when implementing ext2 support?)

bobbinth: All elements in the OOD frame are in the extension field (including the `main` segment values).

hackaugusto: Ah, I misunderstood that. Thanks for the clarification!

codegen/masm/tests/test_basic_arithmetic.rs
bobbinth: A few thoughts on this:

1. We should probably emit the code as several procedures. At the very least, I'd have 3 procedures: one for each type of constraint. For example, `eval_boundary_first`, `eval_boundary_last`, and `eval_transition`. These 3 procedures could be executed sequentially by a wrapper procedure and then the results could be combined by a boilerplate procedure. Having this structure would make the results a bit easier to debug.
2. I'd separate the emitted code for each constraint by a newline. Also, even if we can't link back to the original expression, I'd probably still emit comments to separate one constraint from another (i.e., `# constraint 1`, `# constraint 2` etc.).
3. There is no need to optimize at this point, but I think once we have the general structure working, we'd probably need to spend some time on optimizations to get the cycle count of Miden VM constraints to (ideally) under 30K cycles.

codegen/masm/tests/test_basic_arithmetic.rs
```diff
@@ -0,0 +1,227 @@
...
+def Add
+
+trace_columns:
+    main: [a, b]
+
+public_inputs:
+    stack_inputs: [16]
+
+boundary_constraints:
+    enf a.first = 0
```
bobbinth, line 32: I'd probably add a constraint against the last step as well, and maybe something which involves public inputs.

codegen/masm/src/codegen.rs
```diff
@@ -0,0 +1,273 @@
...
+                        let scalar = mat
+                            .get(*x)
+                            .and_then(|v| v.get(*y))
+                            .ok_or(CodegenError::InvalidAccessType)?;
+                        self.ins.push(format!("push.{scalar}").into());
+                    }
+                    _ => return Err(CodegenError::InvalidAccessType),
+                },
+                Entry::Vacant(_) => return Err(CodegenError::UnknowConstant),
+            },
```
bobbinth, line 213: One of the things we could do in the IR as an optimization pass is to get rid of all bound constants by converting them into inline constants. Once we do that, this could would not be necessary.

codegen/masm/src/codegen.rs
bobbinth: I think this approach is fine for now - let's get all pieces in place first. But once this is done, I think the next step would be to build an optimizer which reduces the number of cycles. I'm thinking we could build a component which given an algebraic expression (and location of operands in memory) can build a Miden VM program which compute this expression in the minimum number of cycles.

The expressions we deal with are usually small enough that we may even be able to do this using an exhaustive search (rather than doing some kind of clever dynamic programming things). In the end, we don't really care if it takes and hour (or even many hours) to generate the code as this would be done relatively infrequently.

hackaugusto: I agree that an Middle IR would be useful, it makes optimization and translation easier to perform. But it is probably better to have this algebraic representation independent of stack positioning, and the lowering of the representation to Masm would take care of instruction scheduling and spill off.

Regarding optimization (minimizing the number of cycles):

- That requires a cost model, since we only emit instructions and not operations, and instructions have different costs (in terms of cycles and size inside a span block). Organizing the work here with our existing compiler effort would be useful, maybe adding the cost model in the `miden` repository, since that is also useful for the VM implementation and possibly for advanced VM users too.
- As for the optimization itself, I think a naive search would not be ideal, since problems like instruction scheduling are NP-hard. Exploiting symmetries (like fundamental arithmetic properties like commutative property of addition), would reduce the search space and make the problem more tractable. We also don't need to be clever, we can just reuse battle tested algorithms like state space search, graph coloring, or any strategies that are employed by optimizing compilers already.

With that said, we did remove that concern when aligning on this implementation. So I have not done any work towards that.

codegen/masm/tests/test_basic_arithmetic.rs
hackaugusto: 1. Agreed. This only implements the translation of AirIR to Masm code, which should be the same for transition and boundary constraints. Wrapping this in a procedure has been agreed upon and should be easy to add on top of this.
2. Okay, I will change the vector to instead of having instructions only, it will contain MASM fragments, to include the comments you suggested.
3. Left some comments on optimization on a previous comment.

codegen/masm/tests/test_basic_arithmetic.rs
hackaugusto: Agreed. Note that this is not the complete backend, I'm trying to open a PR to get input early on. This constraint is currently *not* evaluated, since the ATM the backend is only generating the MASM for the transition constraints. Currently this is only here to satisfy the Air parser, once the boundary constraint is implemented I will add more complex tests.

codegen/masm/src/codegen.rs
bobbinth: Yep - this is more me thinking about the future. We'd probably need some thing like a separate component which would take an algebraic expression and emit optimal MASM code for computing it. This component could be invoked from methods like `visit_integrity_constraint()`, `visit_boundary_constraint()` etc. Internally, it would contain operation cost model (which could be shared with Miden IR in the future), a way keep current location of each operand (whither on the stack or in memory) etc.

codegen/masm/tests/test_basic_arithmetic.rs
hackaugusto: Updated the codegen, now it adds a comment:

```
 padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop ext2add push.0.0 ext2sub # constraint 0
 padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop ext2sub push.0.0 ext2sub # constraint 1
 padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop ext2mul push.0.0 ext2sub # constraint 2
 padw mem_loadw.4294965000 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop ext2add push.0.0 ext2sub # constraint 3
 padw mem_loadw.4294965000 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop ext2sub push.0.0 ext2sub # constraint 4
 padw mem_loadw.4294965000 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop ext2mul push.0.0 ext2sub # constraint 5
```

codegen/masm/src/codegen.rs
hackaugusto: code is updated to use quadratic extension fields

codegen/masm/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+pub mod visitor;
+
+use codegen::{Codegen, CodegenConfig, CodegenError};
+use ir::AirIR;
+use visitor::AirVisitor;
+
+// CODEGEN
+// ================================================================================================
+
+/// Given a [AirIR] generates code to evaluate the boundary and transition constrains in Masm.
```
grjte, line 12: nit: typo in "constraints"

codegen/masm/src/codegen.rs
```diff
@@ -0,0 +1,370 @@
...
+        Codegen {
+            pieces: vec![],
+            constants: BTreeMap::new(),
+            transition_contraint: 0,
+            ir: None,
+            config,
+        }
+    }
+
+    /// Emits the Masm code corresponding to the [AirIR] afert visiting it.
```
grjte, line 43: nit: typo in "after"

codegen/masm/tests/test_basic_arithmetic.rs
```diff
@@ -0,0 +1,443 @@
...
+    let a_prime = a;
+    let b_prime = b;
+    let stack = to_stack_order(&[a, a_prime, b, b_prime]);
+    let code = format!(
+        "
+        begin
+            push.{}                                       # frame address
+            push.{}.{}.{}.{} dup.4 add.0 mem_storew dropw # save a and a'
+            push.{}.{}.{}.{} dup.4 add.1 mem_storew dropw # save b and b'
+            drop                                          # clean frame address
```
grjte, line 51: Non-blocking comment: it might be nice to have a test util in the future to generate this.

codegen/masm/src/codegen.rs
hackaugusto: I'm going to mark this as resolved to make the PR comments cleaner, feel free to change it if needed :)

codegen/masm/tests/test_basic_arithmetic.rs
hackaugusto: I changed the codegen to introduce nested control flow and add procedures, now it looks like this:

```
proc.evaluate_transitions
    exec.cache_periodic_polys
    padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop padw mem_loadw.500000000 drop drop ext2mul push.0 push.0 ext2sub # constraint 0 for main
end # END PROC evaluate_transitions
```

codegen/masm/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+pub fn code_gen(air: &AirIR) -> Result<String, CodegenError> {
+    let config = CodegenConfig {
+        ood_frame_address: constants::OOD_FRAME_ADDRESS,
+        aux_rand_address: constants::AUX_RAND_ELEM_PTR,
+    };
+    let mut codegen = Codegen::new(config);
+    codegen.visit_air(air)?;
+    Ok(codegen.to_masm())
+}
```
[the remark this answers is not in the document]
hackaugusto: changed

bobbinth (commented):
Looks great! Thank you! I left a few comments inline - some of them could probably be addressed in future PRs.

One thing that would be nice to see (but probably in a future PR) is the final output file. This would make it easier to review the structure holistically.

codegen/masm/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+// ================================================================================================
+
+/// Given a [AirIR] generates code to evaluate the boundary and transition constraints in Masm.
+pub fn code_gen(air: &AirIR) -> Result<String, CodegenError> {
+    let config = CodegenConfig {
+        trace_len_address: constants::TRACE_LEN_ADDRESS,
+        z_address: constants::Z_ADDRESS,
+        ood_frame_address: constants::OOD_FRAME_ADDRESS,
+        ood_aux_frame_address: constants::OOD_AUX_FRAME_ADDRESS,
+        compositional_coef_address: constants::COMPOSITIONAL_COEF_ADDRESS,
```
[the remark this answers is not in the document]
hackaugusto: fixed

codegen/masm/src/constants.rs
```diff
@@ -0,0 +1,24 @@
+pub const MAIN_TRACE: u8 = 0;
+pub const AUX_TRACE: u8 = 1;
+
+// https://github.com/0xPolygonMiden/miden-vm/blob/next/stdlib/asm/crypto/stark/verifier.masm#L21
+pub const COMPOSITIONAL_COEF_ADDRESS: u64 = 4294966016;
```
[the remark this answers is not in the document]
hackaugusto: fixed

codegen/masm/tests/test_constants.rs
```diff
@@ -0,0 +1,83 @@
...
+    let program_outputs = process.execute(&program).expect("execution failed");
+    let result_stack = program_outputs.stack();
+
+    // results are in stack-order
+    #[rustfmt::skip]
+    let expected = to_stack_order(&[
+        QuadExtension::<Felt>::ZERO,
+        QuadExtension::<Felt>::ZERO,
+        QuadExtension::<Felt>::ZERO,
+    ]);
```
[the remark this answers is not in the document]
hackaugusto: that is what I did for the tests on `test_basic_arithmetic`, the idea here was to have a different scenario that is more aligned with the happy case. Let me know if I should change / add tests.

codegen/masm/src/config.rs
```diff
@@ -0,0 +1,35 @@
+pub struct CodegenConfig {
+    // Memory location of the trace length using the following format:
+    //
+    //      [trace_len_address] => [0, 0, 0, trace_len]
```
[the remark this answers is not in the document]
hackaugusto: I believe in memory order it would be `[trace_len, 0, 0, 0]`, but in stack-order it is as documented. AFAIK the memory order is not observable by the code, so this documents what the code should expect, in particular, if using `mem_loadw` the result is as above.

codegen/masm/src/codegen.rs
```diff
@@ -0,0 +1,657 @@
...
+/// Converts a column ID to an element position.
+///
+/// The element position is then used to determine the memory location to load, and which elements
+/// of the word must be kept.
+///
+/// The position is used for both the cached value of `z^exp`, which is the point at which the
+/// polynomial equivalent to the periodic column is evaluated, and the result of evaluating the
+/// periodic polynomial at that point. Both values are stored in a unique memory location, the
+/// memomry is expected to be $[0, 0, q_1, q_0]$, this uses additional memory but simplifies code
+/// generation and requires fewer cycles to compute the result, because packing values is not
```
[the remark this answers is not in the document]
hackaugusto: > Is (q0, q1) evaluations of the periodic column polynomial at z^exp?

That is the format in memory for the `z^exp` and for the evaluation. Basically I'm trying to say there is one element per memory location.

> Are we storing the evaluations in the lower part of the word so that they are at the top of the stack when we load them from memory? If so, shouldn't it be [0, 0, q_0, q_1]?

I documented stack order, so the order in there is how the code sees it after the load.

> The text says that we are storing both z^exp and the evaluations in the memory address. So, would the contents of memory address look like [zexp_0, zexp_1, q_0, q_1]?

Both are store in memory, but in different memory ranges. Both are stored as `[0,0,q_1,q_0]`

codegen/masm/src/config.rs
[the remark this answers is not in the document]
bobbinth: I think for documentation purposes we should not use stack order. For example, when we refer to `mem[0][0]` we refer to the first element at memory address 0. This would be the element loaded by `mem_load` instruction. Listing it last makes things a bit more confusing.

I think it is OK to list memory values in the "natural" order because we know that semantics of `mem_loadw` instruction is to put elements onto the stack in the "stack order".

codegen/masm/tests/test_constants.rs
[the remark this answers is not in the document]
bobbinth: I think for this PR it is fine as is - but eventually, we probably want to have more robust tests (similar to what I mentioned in https://github.com/0xPolygonMiden/air-script/pull/296#discussion_r1203538746).

codegen/masm/src/codegen.rs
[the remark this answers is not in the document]
bobbinth: I guess what confused me here is mentioning of `z^exp`. If it was something like "Values of periodic columns are stored at distinct memory addresses such that each value occupies a single memory word with the two most significant word elements set to zeros (i.e., [q0, q1, 0, 0])" - I would have understood it.

codegen/masm/tests/test_constants.rs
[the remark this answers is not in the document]
hackaugusto: Totally agree, I just don't want to increase the size of this PR since there are more coming as follow ups. The goal is to implement the requests, but with smaller PRs so that it is easier to review.

codegen/masm/src/config.rs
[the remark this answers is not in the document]
bobbinth: The way I think about this is in terms of least significant vs. most significant elements. In a word, we usually say that the least significant element is `word[0]` and the most significant is `word[3]`. On the stack, we say that the least significant element is in the deeper in the stack than than the most significant.

We use the same with non-word values too. For example, when representing a `u64` value using two elements, the element with least significant bits is deeper in the stack than the element with the most significant bits. And when we store it in memory the list significant element ends up in `word[0]`.

I'd prefer to keep this way of looking at things as it would be consistent with how we do it in other places.

bobbinth (approved):
Thank you! Looks good!

Let's update the docs to describe to in-memory words in their natural order (not stack order), and I think we can merge. This way, we can add functionality/address issues in subsequent PRs.

codegen/masm/src/config.rs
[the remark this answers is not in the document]
hackaugusto: updated

codegen/masm/src/codegen.rs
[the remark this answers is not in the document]
hackaugusto: updated

PR #299, opened 2023-05-25
codegen: add tests and cli for masm

hackaugusto:
This adds Masm codegen to the CLI and tests for all the existing cases in `air-script/tests`


hackaugusto:
For the time being the tests random and selectors are skipped because the IR has some non-determinism in its construction. After the front-end rework is merged I will re-visit these.

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -0,0 +1,57 @@
...
+    padw mem_loadw.4294965001 drop drop padw mem_loadw.4294965002 movdn.3 movdn.3 drop drop padw mem_loadw.4294965000 drop drop ext2add ext2sub # constraint 1 for main
+    padw mem_loadw.4294965002 movdn.3 movdn.3 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop padw mem_loadw.4294965001 movdn.3 movdn.3 drop drop ext2add ext2sub # constraint 2 for main
+    padw mem_loadw.4294965072 drop drop padw mem_loadw.4294965072 movdn.3 movdn.3 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop ext2add padw mem_loadw.4294965001 movdn.3 movdn.3 drop drop ext2add padw mem_loadw.4294965000 drop drop ext2add ext2mul ext2sub # constraint 0 for aux
+    padw mem_loadw.4294965073 movdn.3 movdn.3 drop drop padw mem_loadw.4294965073 drop drop padw mem_loadw.4294965002 movdn.3 movdn.3 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop ext2add ext2mul ext2sub # constraint 1 for aux
+end # END PROC compute_evaluate_transitions
+
+proc.evaluate_transitions
+    exec.cache_periodic_polys
+    exec.compute_evaluate_transitions
```
bobbinth, line 30: A few nits for the future regarding naming and file organization. I'm thinking that the main exported function which looks something like this:
```
export.evaluate_constraints
  # same as the current cache_periodic_polys
  exec.compute_periodic_values

  # evaluates transition constraints (the result is an aggregated numerator)
  exec.evaluate_transition_constraints

  # evaluates boundary constraints (the result is an aggregated numerator)
  exec.evaluate_boundary_constraints

  # divides numerators by their respective divisors and sums up the evaluations
  exec.combine_constraints
end
```

One other questions: seems like right now we are doing evaluation and aggregation separately. I think this fine for now, but in the future (in a more optimized version), the aggregation result should probably just live on the stack. We may even add a specialized instruction to the VM to make this aggregation simpler.

bobbinth (changes requested):
Thank you! Looks good! I left a few comments inline - some of them are for the generated MASM files. These are not for this PR but more like references for the future.

Also, I didn't really review all generated MASM files. I'll try to do that at a later date.

air-script/src/cli/transpile.rs
```diff
@@ -16,9 +21,16 @@ pub struct Transpile {
     #[arg(
         short,
         long,
-        help = "Output filename, default to the input file with the .rs extension"
+        help = "Output filename, default to the input file with the .rs extension for Winterfell or .masm for MASM"
```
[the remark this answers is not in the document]
hackaugusto: thanks!

air-script/src/lib.rs
```diff
@@ -1,11 +1,6 @@
 // EXPORTS
 // ================================================================================================
-
-/// AirScript parse method to generate an AST from AirScript source files
-pub use parser::parse;
-
-/// AirScript intermediate representation
+pub use codegen_masm::CodeGenerator as MasmCodeGenerator;
```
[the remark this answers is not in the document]
hackaugusto: added

air-script/src/cli/transpile.rs
```diff
@@ -57,10 +61,27 @@ impl Transpile {
         let ir = ir.unwrap();
 
         // generate Rust code targeting Winterfell
-        let codegen = CodeGenerator::new(&ir);
+        let (code, extension) = match self.target.unwrap_or(Target::Winterfell) {
```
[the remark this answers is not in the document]
hackaugusto: done

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -0,0 +1,57 @@
...
+    # Clean stack
+    dropw
+end # END PROC cache_z_exp
+
+proc.cache_periodic_polys
+    exec.cache_z_exp
+end # END PROC cache_periodic_polys
+
+proc.compute_evaluate_transitions
+    padw mem_loadw.4294965000 drop drop padw mem_loadw.4294965001 movdn.3 movdn.3 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop padw mem_loadw.4294965001 movdn.3 movdn.3 drop drop ext2mul padw mem_loadw.4294965002 movdn.3 movdn.3 drop drop ext2mul ext2add ext2sub # constraint 0 for main
```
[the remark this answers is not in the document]
hackaugusto: changed

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -0,0 +1,57 @@
...
+    padw mem_loadw.4294967291 drop drop
+    # => [z_1, z_0, num_of_cycles, trace_len, ...]
+    # => [z_1^2, z_0^2, num_of_cycles', trace_len, ...]
+    # Clean stack
+    dropw
+end # END PROC cache_z_exp
+
+proc.cache_periodic_polys
+    exec.cache_z_exp
+end # END PROC cache_periodic_polys
```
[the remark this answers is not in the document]
hackaugusto: changed

air-script/tests/aux_trace/aux_trace.masm
hackaugusto: Is there are reason why you want to renamed the procedures?

I used `compute_` for procedures that leave data on the stack, and `cache_` for procedures that save the data to memory.

> One other note: seems like right now we are doing evaluation and aggregation separately. I think this fine for now, but in the future (in a more optimized version), the aggregation result should probably just live on the stack. We may even add a specialized instruction to the VM to make this aggregation simpler.

This was done for testing purposes, otherwise we can't test the intermediate values

air-script/tests/main.rs
```diff
@@ -1,191 +1,327 @@
...
+    expected.assert_eq(&generated_winterfell);
 
-    let generated_air = Test::new("tests/random_values/random_values_bindings.air".to_string())
-        .transpile()
-        .unwrap();
+    // let generated_masm = Test::new("tests/random_values/random_values_simple.air".to_string())
+    //     .transpile(Target::Masm)
+    //     .unwrap();
+    // let expected = expect_file!["random_values/random_values.masm"];
+    // expected.assert_eq(&generated_masm);
```
grjte, line 212: nit: I think it would be good to add a `TODO` comment about the non-determinism so it's clear why this is commented out. Same thing in the other affected tests below

air-script/tests/main.rs
```diff
@@ -1,191 +1,327 @@
...
     let expected = expect_file!["aux_trace/aux_trace.rs"];
-    expected.assert_eq(&generated_air);
+    expected.assert_eq(&generated_winterfell);
+
+    let generated_masm = Test::new("tests/aux_trace/aux_trace.air".to_string())
+        .transpile(Target::Masm)
+        .unwrap();
+
+    let expected = expect_file!["aux_trace/aux_trace.masm"];
+    expected.assert_eq(&generated_masm);
```
grjte, line 23: I'm not convinced that we should have both of these in the same test (and similar for the rest of the tests in this file). It implies that if we added more backends (e.g. the IR we'll share with PIL or something targeting another zkVM) we would test them all here too. I would probably advocate a different test module for each backend. I assume you considered this though, so it'd be good to hear your thinking

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -0,0 +1,42 @@
...
+    movdn.2 movdn.2
+    # Accumulate the 4-th constraint into the numerator
+    padw mem_loadw.4294966018 movdn.3 movdn.3 drop drop ext2mul ext2add # accumulate the contraint into the numerator
+    # Save the accumulator
+    movdn.2 movdn.2
+    # Accumulate the 3-th constraint into the numerator
+    padw mem_loadw.4294966017 drop drop ext2mul ext2add # accumulate the contraint into the numerator
+    # Save the accumulator
+    movdn.2 movdn.2
+    # Accumulate the 2-th constraint into the numerator
```
grjte, line 28: minor nit: we could avoid the strange English by just saying "Accumulate constraint {i} into the numerator" like in `compute_evaluate_transitions`

grjte (commented):
Looks good to me, thanks! Just a few small non-blocking comments

air-script/tests/main.rs
hackaugusto: I haven't thought much about it, will add a new module 👍

air-script/tests/aux_trace/aux_trace.masm
hackaugusto: changed.

air-script/tests/main.rs
hackaugusto: added

air-script/tests/main.rs
hackaugusto: I split the tests into 2 separate modules, because each module is a different binary (these are integration tests), the compiler was complaining the other option was not being used (i.e. the winterfell tests didn't generate masm code, so the `Target::Masm` was never used on that compilation unit).

To make the compiler happy, I made `main.rs` import both `winterfell` and `masm`, so all variants of the enum are used, and no warning is emitted. To avoid having the compilation for both tests being done twice, I moved the code into a module named `codegen`.

To summarize:

- `codegen/winterfell.rs` run the winterfell codegen tests
- `codegen/masm.rs` runs the masm
- all tests are inside `codegen` to satisfy the compiler

air-script/src/cli/transpile.rs
```diff
@@ -57,10 +61,27 @@ impl Transpile {
...
+            Target::Winterfell => (codegen_winter::CodeGenerator::new(&ir).generate(), "rs"),
+            Target::Masm => (
+                codegen_masm::CodeGenerator::new(&ir, codegen_masm::CodegenConfig::default())
+                    .generate()
+                    .expect("code generation failed"),
```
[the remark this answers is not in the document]
bobbinth: Yeah - not sure. I does seem cleaner to handle the errors somewhere downstream (even if it is just printing an error message) - but we can keep it as is as well.

air-script/tests/aux_trace/aux_trace.masm
bobbinth: > Is there are reason why you want to renamed the procedures?
> 
> I used `compute_` for procedures that leave data on the stack, and `cache_` for procedures that save the data to memory.

It communicates the purpose of procedures a bit better I think. Plus, we'll probably have most of these procedures (if not all) write the results to memory rather than leave them on the stack (with `combine_constraints` being potentially the only exception).

> This was done for testing purposes, otherwise we can't test the intermediate values

I think we could still do it by setting random coefficients for specific constraints to zeros. Basically, if we have 3 constraints, and we want to test that the 2nd one evaluates correctly, we could set random coefficient for that constraint to `ONE` and for the other two constraints to `ZERO`. This does assume that there are independent tests which make sure constraint aggregation works correctly.

bobbinth (approved):
Looks good! Thank you!

PR #300, opened 2023-05-29
masm codegen: optimized codegen for numerator aggregation

hackaugusto:
instead of multiplying the coefficients on `evaluate_transitions`, this performs the multiplications on `compute_evaluate_transitions`, which means there is no longer a need to manipulate the accumulator on the stack, saving a few cycles.

requested here: https://github.com/0xPolygonMiden/air-script/pull/299#discussion_r1207190470

### DIFF
--- a/air-script/tests/periodic_columns/periodic_columns.masm
+++ b/air-script/tests/periodic_columns/periodic_columns.masm
@@ -125,23 +125,19 @@ end # END PROC cache_periodic_polys
 proc.compute_evaluate_transitions
     # constraint 0 for main
     padw mem_loadw.500000000 drop drop padw mem_loadw.4294965001 movdn.3 movdn.3 drop drop padw mem_loadw.4294965002 movdn.3 movdn.3 drop drop ext2add ext2mul push.0 push.0 ext2sub
+    # Multiply by the composition coefficient
+    padw mem_loadw.4294966016 movdn.3 movdn.3 drop drop ext2mul
     # constraint 1 for main
     padw mem_loadw.500000001 drop drop padw mem_loadw.4294965000 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop ext2sub ext2mul push.0 push.0 ext2sub
+    # Multiply by the composition coefficient
+    padw mem_loadw.4294966016 drop drop ext2mul
 end # END PROC compute_evaluate_transitions
 
 proc.evaluate_transitions
     exec.cache_periodic_polys
     exec.compute_evaluate_transitions
-    # Compute the numerator of the constraint polynomial
-    push.0 push.0
-    # Save the accumulator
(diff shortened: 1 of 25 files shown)


air-script/tests/periodic_columns/periodic_columns.masm
```diff
@@ -125,23 +125,19 @@ end # END PROC cache_periodic_polys
 proc.compute_evaluate_transitions
     # constraint 0 for main
     padw mem_loadw.500000000 drop drop padw mem_loadw.4294965001 movdn.3 movdn.3 drop drop padw mem_loadw.4294965002 movdn.3 movdn.3 drop drop ext2add ext2mul push.0 push.0 ext2sub
+    # Multiply by the composition coefficient
+    padw mem_loadw.4294966016 movdn.3 movdn.3 drop drop ext2mul
     # constraint 1 for main
     padw mem_loadw.500000001 drop drop padw mem_loadw.4294965000 drop drop padw mem_loadw.4294965000 movdn.3 movdn.3 drop drop ext2sub ext2mul push.0 push.0 ext2sub
+    # Multiply by the composition coefficient
+    padw mem_loadw.4294966016 drop drop ext2mul
```
bobbinth, line 133: A couple of questions/comments:

1. Is `4294966016` the address for random coefficients used to compose constraint evaluations? If so, we should use distinct values here. That is, each constraint should be multiplied by its own coefficient (so, the address should probably be incremented by 1 for each constraint).
2. I think this is good for now, but eventually, we might optimize it even further by having a dedicated instruction which can load the coefficient, multiply by it, and aggregate the result in a single VM cycle.

hackaugusto: > we should use distinct values here. That is, each constraint should be multiplied by its own coefficient (so, the address should probably be incremented by 1 for each constraint).

These are distinct values, they are loaded here:

https://github.com/0xPolygonMiden/air-script/pull/300/files#diff-2fe2575b239a9080f57590f8d184b6ad5335fc1e316f22a4659c236e17f234ffR384-R390

The snippet above is loading the ext2 element `(a, b)` from `[a, b, c, d]`, the second is loading `(c, d)` from the same address.

hackaugusto: > I think this is good for now, but eventually, we might optimize it even further by having a dedicated instruction which can load the coefficient, multiply by it, and aggregate the result in a single VM cycle.

https://github.com/0xPolygonMiden/miden-vm/issues/913

bobbinth: > The snippet above is loading the ext2 element `(a, b)` from `[a, b, c, d]`, the second is loading `(c, d)` from the same address.

Ah - makes sense!

bobbinth (approved):
Looks good! Thank you!

PR #302, opened 2023-05-30
katex: udpate config

hackaugusto:
`mdbook-katex 0.5.3` no longer accepts the configuration option `output.katex`, update it to `preprocessor.katex` [ref](https://github.com/lzanini/mdbook-katex#katex-options).

Note: For local runs updating katex may be necessary `cargo install mdbook mdbook-katex`

### DIFF
--- a/docs/book.toml
+++ b/docs/book.toml
@@ -5,11 +5,11 @@ multilingual = false
 src = "src"
 title = "AirScript by Polygon Miden"
 
-[output.katex]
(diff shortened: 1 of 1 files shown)


bobbinth (approved):
Looks good! Thank you! I did something similar for Miden VM [docs](https://github.com/0xPolygonMiden/miden-vm/blob/next/docs/book.toml).

PR #305, opened 2023-06-06
feat: add utilities for pretty-printing ast

bitwalker:
This PR adds support for pretty printing the AST, useful for troubleshooting and reviewing the results of various transformations.

NOTE: This is based on the branch from #297 and is part of a chain of PRs that implement modules, as well as a number of refactorings/improvements to the AirScript compiler toolchain. This is the 3rd PR in that chain. It must be merged into #304 or merged into #297 after #304.

### DIFF
--- a/parser/src/ast/mod.rs
+++ b/parser/src/ast/mod.rs
@@ -1,4 +1,5 @@
 mod declarations;
+mod display;
 mod errors;
 mod expression;
 mod module;
@@ -7,6 +8,7 @@ mod trace;
 pub mod visit;
 
 pub use self::declarations::*;
+pub(crate) use self::display::*;
 pub use self::errors::*;
 pub use self::expression::*;
 pub use self::module::*;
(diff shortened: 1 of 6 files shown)


parser/src/ast/mod.rs
```diff
@@ -288,6 +290,90 @@ impl PartialEq for Program {
...
+            }
+            f.write_str("\n")?;
+        }
+
+        if !self.constants.is_empty() {
+            for (qid, constant) in self.constants.iter() {
+                if qid.module == self.name {
+                    writeln!(f, "    const {} = {}", &qid.item, &constant.value)?;
+                } else {
+                    writeln!(f, "    const {} = {}", qid, &constant.value)?;
```
grjte, line 337: We don't indent constant declarations in the language, so we probably shouldn't do it here either

grjte (changes requested):
Thank you! Just a few very minor points

parser/src/ast/mod.rs
bitwalker: Ah yep, not sure why I indented these

bitwalker:
I've updated the PR to address your requested changes, let me know if anything else catches your eye!

grjte (approved):
LGTM thanks!

PR #307, opened 2023-06-06
feat: support more expressiveness in the ast

bitwalker:
This commit introduces the following changes to the AST:

1. Change `Expr::Vector` to contain `Expr` elements rather than just `ScalarExpr` elements. This is necessary in order to properly represent the semantics of vectors in AirScript
2. Introduce `Statement::EnforceIf` for use during translation to represent individual scalar constraints with runtime selectors. We do not use this variant until inlining is performed to retain the benefits of the previous representation using `EnforceAll`
3. Introduce `Statement::Expr` for use during translation to represent statements used in an expression context. Specifically, when inlining is performed on a let-bound expression, the inlined expression may require the use of let-bound variables, so we need a way to represent such statements during the inlining transformation.

NOTE: This is based on the branch from #297 and is part of a chain of PRs that implement modules, as well as a number of refactorings/improvements to the AirScript compiler toolchain. This is the 5th PR in that chain. It must be merged into #306 or merged into #297 after #306.

### DIFF
--- a/parser/src/ast/expression.rs
+++ b/parser/src/ast/expression.rs
@@ -54,6 +54,12 @@ impl PartialEq<&str> for Identifier {
         self.0.item == *other
     }
 }
+impl PartialEq<&Identifier> for Identifier {
+    #[inline]
+    fn eq(&self, other: &&Self) -> bool {
+        self == *other
+    }
+}
 impl fmt::Debug for Identifier {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_tuple("Identifier")
@@ -138,9 +144,26 @@ impl QualifiedIdentifier {
         self.item
     }
 
+    /// Returns the name of the item in its [Symbol] form
+    #[inline]
+    pub fn name(&self) -> Symbol {
+        self.as_ref().name()
+    }
+
     /// Returns true if this identifier refers to a known builtin function
     pub fn is_builtin(&self) -> bool {
-        matches!(self.as_ref().as_str(), "sum" | "prod" if self.module.name() == "$builtin")
+        use crate::symbols;
+
+        if self.module.name() == "$builtin" {
+            match self.item {
+                NamespacedIdentifier::Function(id) => {
+                    matches!(id.name(), symbols::Sum | symbols::Prod)
+                }
+                _ => false,
+            }
+        } else {
+            false
+        }
     }
 }
 impl AsRef<Identifier> for QualifiedIdentifier {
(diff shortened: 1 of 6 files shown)


parser/src/ast/expression.rs
```diff
@@ -621,6 +674,156 @@ impl SymbolAccess {
...
+                        ty: Some(Type::Felt),
+                        ..self.clone()
+                    }),
+                    Type::Matrix(rows, _) if idx >= rows => {
+                        Err(InvalidAccessError::IndexOutOfBounds)
+                    }
+                    Type::Matrix(_, cols) => Ok(Self {
+                        access_type: AccessType::Index(idx),
+                        ty: Some(Type::Vector(cols)),
+                        ..self.clone()
```
grjte, line 705: nit: This function is really long and repeats very similar logic a few times. Is there any way to improve readability? Also, I think there's some logic that's repeated from the other `access` function. Can anything be consolidated?

parser/src/ast/trace.rs
```diff
@@ -208,8 +208,8 @@ impl<'a> fmt::Debug for FormatConstrainedFlags<'a> {
 pub struct TraceBinding {
     #[span]
     pub span: SourceSpan,
-    /// The name of this binding
-    pub name: Identifier,
+    /// The name of this binding, if applicable
+    pub name: Option<Identifier>,
```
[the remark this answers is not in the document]
bitwalker: When it is used in a type context - i.e. when it is used with `BindingType::TraceColumn` or `BindingType::TraceParam`, it is not always the case that we know (or care) what the name of the actual binding is. In most cases in which it is used though, a `TraceBinding` has a name, so it's really just the fact that we're reusing the `TraceBinding` type for more than one purpose that necessitates the optional name here.

parser/src/ast/expression.rs
bitwalker: I'm not sure there is a meaningful way to break this up - we can't really escape the quadratic complexity of the function. Each combination of base access type, nested access type (relative to the type of the value produced by the base access) produces a unique result. 

Functions like this are annoying, since we must enumerate all of the combinations explicitly, but there also isn't really a better way to do it, at least in Rust with two enums involved. We could forcibly break it up by yanking parts out into new free functions, but I'm not sure that's actually easier to read. Better a long function that handles all of the cases clearly (and exposes any inherent patterns), than a short function that hides those details behind more abstraction - but that's a double-edged sword to be sure. Any thoughts on how you'd tackle it?

grjte: Hm - my preference does tend to be using well-named helper functions and keeping the main function smaller, but I'd like to defer to @bobbinth on this again as well, since it's really a question of team style (similar to the `unwrap` discussion)

bobbinth: I would agree with @grjte - would be better to have smaller, well-named helper functions which handle individual cases separately. There could be situations when this is not desirable - e.g., when there is a lot of context that needs to be passed into the helper functions. But if context is relatively small, I find that it usually improves readability (and reduces the amount of nesting).

For example, in this specific case, one way to do this could be to have a separate helper function for each initial match branch. Maybe something like this:
```Rust
match self.access_type.clone() {
    AccessType::Default => access_default(ty),
    AccessType::Slice(range) => access_slice(ty, range),
    AccessType::Index(base_idx) => access_index(ty, base_idx),
    AccessType::Matrix(_, _) => match access_type {
        AccessType::Default => Ok(self.clone()),
        _ => Err(InvalidAccessError::IndexIntoScalar),
    }
}
```
And then individual functions `access_default()`, `access_slice()`, `access_index()` (names could be different too) would be much easier to digest.

PR #308, opened 2023-06-06
fix: use btreemap for more deterministic ast diffs

bitwalker:
As mentioned in the title, this PR swaps out the use of `HashMap` in the AST for `BTreeMap`, so that diffs in test assertions do not contain spurious differences related to iteration order. In general it should make compilation more deterministic, which is of particular value in generated test artifacts which must match previous outputs.

NOTE: This is based on the branch from #297 and is part of a chain of PRs that implement modules

### DIFF
--- a/parser/src/symbols.rs
+++ b/parser/src/symbols.rs
@@ -49,7 +49,7 @@ impl SymbolTable {
 unsafe impl Sync for SymbolTable {}
 
 /// A symbol is an interned string.
-#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
 pub struct Symbol(SymbolIndex);
 
 impl Symbol {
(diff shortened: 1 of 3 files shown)


parser/src/symbols.rs
```diff
@@ -24,13 +24,19 @@ pub mod predefined {
     pub const Sum: Symbol = Symbol::new(3);
     /// The symbol `prod`
     pub const Prod: Symbol = Symbol::new(4);
+    /// The symbol `program_hash`
+    pub const ProgramHash: Symbol = Symbol::new(5);
+    /// The symbol `stack_inputs`
+    pub const StackInputs: Symbol = Symbol::new(6);
```
grjte, line 30: These things are specific to MidenVM and shouldn't be part of AirScript. They are examples of specific public inputs. I don't remember seeing this before, so hopefully I didn't miss it, but if they're anywhere else (other than tests), they should be removed.

grjte (changes requested):
Thanks @bitwalker! Looks good mostly, with one comment left inline

parser/src/symbols.rs
bitwalker: I've gone ahead and made the `Symbol` implementation of `Ord` use the underlying string value, if we ever get into optimization in the future, we can revisit this then.

PR #309, opened 2023-06-06
chore: clean up semantic analysis/constant propagation

bitwalker:
This PR refactors the semantic analysis and constant propagation passes to factor out some unnecessarily large functions, split up the `sema` module better, and provide a better data structure for handling various types of bindings based on lexical scope.

NOTE: This is based on the branch from #297 and is part of a chain of PRs that implement modules, as well as a number of refactorings/improvements to the AirScript compiler toolchain. This is the 7th PR in that chain. It should be merged into #308 or merged into #297 after #308.

### DIFF
--- a/parser/src/sema/binding_type.rs
+++ b/parser/src/sema/binding_type.rs
@@ -1,0 +1,236 @@
+use crate::ast::{AccessType, FunctionType, InvalidAccessError, RandBinding, TraceBinding, Type};
+use std::fmt;
+
+/// This type provides type and contextual information about a binding,
+/// i.e. not only does it tell us the type of a binding, but what type
+/// of value was bound. This is used during analysis to check whether a
+/// particular access is valid for the context it is in, as well as to
+/// propagate type information while retaining information about where
+/// the type was derived from.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BindingType {
+    /// A local variable whose value is not an alias of a global/module declaration
+    Local(Type),
+    /// A local variable that aliases a global/module declaration
+    Alias(Box<BindingType>),
+    /// A direct reference to a constant declaration
+    Constant(Type),
+    /// A type associated with a function signature
+    ///
+    /// The result type is None if the function is an evaluator
+    Function(FunctionType),
+    /// A function parameter corresponding to trace columns
+    TraceParam(TraceBinding),
+    /// A direct reference to one or more contiguous trace columns
+    TraceColumn(TraceBinding),
+    /// A potentially non-contiguous set of trace columns
+    Vector(Vec<BindingType>),
+    /// A direct reference to a random value binding
+    RandomValue(RandBinding),
+    /// A direct reference to a public input
+    PublicInput(Type),
+    /// A direct reference to a periodic column
+    PeriodicColumn(usize),
+}
+impl BindingType {
+    /// Get the value type of this binding, if applicable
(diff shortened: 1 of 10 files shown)


parser/src/sema/binding_type.rs
```diff
@@ -0,0 +1,236 @@
...
+                        mid_and_rest.extend_from_slice(rest);
+                        (first, Some(Self::Vector(mid_and_rest)))
+                    }
+                }
+            }
+            invalid => panic!("invalid trace column(s) binding type: {:#?}", invalid),
+        }
+    }
+
+    pub fn access(&self, access_type: AccessType) -> Result<Self, InvalidAccessError> {
```
grjte, line 184: nit: it would be good to have a comment on this function

parser/src/sema/binding_type.rs
```diff
@@ -0,0 +1,236 @@
+use std::fmt;
+
+use crate::ast::{AccessType, FunctionType, InvalidAccessError, RandBinding, TraceBinding, Type};
```
grjte, line 3: nit: usually we avoid having extra lines between use statements

parser/src/sema/scope.rs
```diff
@@ -0,0 +1,146 @@
...
+    borrow::Borrow,
+    collections::HashMap,
+    hash::Hash,
+    ops::{Index, IndexMut},
+};
+
+pub type Env<K, V> = Box<HashMap<K, V>>;
+
+#[derive(Clone)]
+pub enum LexicalScope<K, V> {
```
grjte, line 11: nit: as a code style thing, we usually put headers above sections of code to separate things like constants, types, and the main code in the file, like the following:

```
// TYPES
//========================== (continued to a width of 100 chars)
```

This separation is useful for reading through files. It would be helpful especially to have some comments on the `LexicalScope`.

You can see examples in the [miden-vm repo](https://github.com/0xPolygonMiden/miden-vm/blob/main/processor/src/lib.rs)

grjte (approved):
Thank you, I think this organization is easier to read through. I left a few nits inline related to team code style

parser/src/sema/scope.rs
bitwalker: I think that can be helpful when there is a large number of top-level items in a single file, but I'm not sure how that applies here. This file defines a single type (`LexicalScope`) and a single type alias used to aid readability within the file, so having a pair of section comments to visually separate and identify them just seems like noise to me.

Of course I'll add them if you would prefer to have them rather than not, but it does feel out of place in Rust code IMHO. Generally you don't group declarations by type in Rust, but by domain. This is why you can define multiple `impl` blocks for a type (to group together declarations for a type which are used for a common purpose), and why you can inline declarations within a function body (e.g. constants, newtypes, etc.). In any case, let me know what convention you'd prefer I use in this repo, and I'll make a pass through and update all the modules I've touched so things are consistent.

bitwalker: FYI - I'm going to merge this as-is into the `bitwalker/modules` branch, since both this PR and the next have been signed off, but just let me know about your preference here and I can apply a fixup to that branch instead

grjte: This was really intended as more of a general note about the code style we've used as a team across other repositories, and I don't have a strong opinion about it here. I think the main reason I ended up noting it here is that there wasn't any comment on the `LexicalScope` struct, and we do in general try to comment most things.

PR #311, opened 2023-06-06
feat: implement lowering from ast to air

bitwalker:
This PR implements the rest of the changes related to the modules implementation started in #297. Specifically, this PR implements the translation pass from the AST to AIR, cleans up the `air-ir` crate, updates the tests, and updates the downstream `codegen/winterfell` and `air-script` crates.

This PR, being the last in the chain started by #297, can be retargeted to `next` and merged to merge all of the modularity-related PRs:

* #297 
* #304 
* #305 
* #306 
* #307 
* #308 
* #309 
* #310 

Otherwise, each of the PRs above must be retargeted+merged starting with #297. Since all of those PRs do not independently compile until this one, I'd recommend conducting your review in each PR, then when we're done with review, closing all but this one and retargeting it against `next`, rebasing against any recent changes, and merging when ready.


bitwalker:
NOTE: Due to commits in other PRs, we'll want to merge individual PRs into the `bitwalker/modules` branch (i.e. #297), and then merge that PR at the end, rather than merging this PR, otherwise it requires a bunch of rebasing to keep the branches clean and that's not really worth it IMO. I'll stay on top of that though as PRs get merged.

air-script/tests/list_comprehension/list_comprehension.rs
```diff
@@ -86,9 +86,9 @@ impl Air for ListComprehensionAir {
         let main_next = main_frame.next();
         let aux_current = aux_frame.current();
         let aux_next = aux_frame.next();
-        result[0] = aux_current[0] - E::from(main_current[0]) * E::from(2_u64).exp(E::PositiveInteger::from(3_u64)) * aux_current[7];
+        result[0] = aux_current[0] - E::from(main_current[0]) * E::from(8_u64) * aux_current[7];
         result[1] = aux_current[0] - E::from(main_current[0]) * (aux_next[4] - aux_next[8]);
-        result[2] = aux_current[2] - E::from(main_current[0]) * (aux_current[5] - aux_current[10]);
+        result[2] = aux_current[2] - E::from(main_current[0]) * (aux_current[5] - aux_current[9]);
```
grjte, line 91: I think this change is incorrect, but please check, in case I've made a mistake.

This comes from the following AirScript code:

```
trace_columns:
    main: [clk, fmp[2], ctx]
    aux: [a, b[3], c[4], d[4]]

...

    let diff_slice_iterables = [x - y for (x, y) in (c[0..2], d[1..3])]
    enf b[1] = clk * diff_slice_iterables[1]
```

`diff_slice_iterables[1]` should correspond to `c[1] - d[2]`, which is `aux_current[5] - aux_current[10]`

air-script/tests/main.rs
```diff
@@ -155,6 +155,7 @@ fn list_folding() {
 }
 
 #[test]
+#[ignore]
```
grjte, line 158: Sorry if I missed this explanation somewhere - why is this test being ignored?

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -1,17 +1,18 @@
...
 };
 
+use super::{Codegen, ElemType, Impl};
+
 // HELPERS TO GENERATE THE WINTERFELL BOUNDARY CONSTRAINT METHODS
 // ================================================================================================
 
 /// Adds an implementation of the "get_assertions" method to the referenced Air implementation
-/// based on the data in the provided AirIR.
+/// based on the data in the provided Air.
```
grjte, line 13: nit: I think this actually becomes a bit confusing now, because we're using "Air" to talk about two different things (the Air struct implementation that is being output, and the IR that is being referenced). Can we add a qualifier to one of them? Same comment with `add_fn_get_aux_assertions` below

codegen/winterfell/src/air/transition_constraints.rs
```diff
@@ -1,11 +1,13 @@
...
+
+use super::{Codegen, ElemType, Impl};
 
 // HELPERS TO GENERATE THE WINTERFELL TRANSITION CONSTRAINT METHODS
 // ================================================================================================
 
 /// Adds an implementation of the "evaluate_transition" method to the referenced Air implementation
-/// based on the data in the provided AirIR.
-pub(super) fn add_fn_evaluate_transition(impl_ref: &mut Impl, ir: &AirIR) {
+/// based on the data in the provided Air.
```
grjte, line 9: nit: same issue mentioned previously with having multiple meanings of "Air" in the same sentence. This also occurs in the comment for `add_fn_evaluate_aux_transition` below

air-script/tests/main.rs
bitwalker: It uses the `match` syntax which was removed, a follow-on PR will re-add it with the syntax changes we discussed and the test will be re-enabled at that time

air-script/tests/list_comprehension/list_comprehension.rs
bitwalker: Good catch! I'm not sure why I missed that when reviewing this test, but I can only assume I miscalculated the correct index in my head. In any case, see #3a05695 for the fix

grjte (approved):
Looks good, thank you! I left a few small comments inline

ir/src/tests/variables.rs
```diff
@@ -223,14 +211,13 @@ fn err_ic_variable_access_before_declaration() {
         enf clk' = clk + a
         let a = 1";
 
-    let parsed = parse(source).expect("Parsing failed");
-    let result = AirIR::new(parsed);
-    assert!(result.is_err());
+    expect_diagnostic(source, "expected one of: '\"enf\"', '\"let\"'");
```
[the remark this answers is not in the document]
bitwalker: This error is generated by LALRPOP, not us, so we don't really control errors of this type (this is one of the reasons why hand-written recursive descent parsers are preferable to parser generators, but I digress). We _may_ be able to modify the grammar to allow catching this specific case but I'd have to experiment a bit. 

IIRC the reason why we're leaning on the parser to catch this is due to a combination of factors:

* Statement blocks cannot be empty without making the grammar ambiguous due to the lack of block delimiters (e.g. `{`/`}`),
* If we have a grammar rule to catch a dangling let (e.g. just the `let x = expr` part), then a parse error in what would have been the body of that `let` expression will produce even worse diagnostic output. Specifically, it will fail to parse the "correct" non-terminal rule for `let` expressions, and fall back to the dangling-let rule, which will succeed because that part of the expression is valid. This will produce a dangling-let error even though there isn't one. If we make that type of error a "soft" error, parsing will fail immediately after that on the next token anyway, with an unexpected token error, as it is highly unlikely that the next token will be valid at that point in the parse tree.

The typical solution here is to relax the grammar and catch validation errors during semantic analysis, but that is only possible when relaxing the grammar doesn't make it ambiguous, which I believe is the issue here.

ir/src/tests/variables.rs
```diff
@@ -165,14 +152,15 @@ fn err_ic_variables_vector_with_inlined_vector() {
         let d = [a[0], [3, 4]]
         enf clk' = d[0][0]";
 
-    parse(source).expect_err("Parsing failed");
+    expect_diagnostic(source, "expected one of");
```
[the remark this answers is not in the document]
bitwalker: I decided against explicitly listing all of the tokens expected at this point, since there are many, and I'm not sure if the ordering is guaranteed to remain stable. Since the previous assertion was basically not checking what caused the error at all, this assertion at least catches the fact that an unexpected token error was raised.

That said, I'm happy to change the assertion to cover all the expected tokens if you think that's valuable here.

PR #312, opened 2023-06-07
Move chiplet bus constraints to the separate file. 

Overcastan:
This small PR removes chiplet bus constraints from the `memory.air` and `bitwise.air` files. In the next PR these constraints will be implemented in a separate file along with the rest of chiplet bus constraints.


bobbinth (approved):
Looks good! Thank you!

PR #316, opened 2023-06-18
masm backend: add integrity constraint divisor

hackaugusto:
This is a big PR, it introduces the constraint divisor for the integrity constraints. I'm going to open the divisor for the boundary constraint on a follow up since this PR is already pretty big.

The changes on this PR include:

- Syncing with the Miden's stdlib verifier constants
- Optimized the caching of $z^{power}$.
   - The stdlib code was updated to keep a copy of the `log(trace_len)`, and the procedure was changed to use that as a counter (with the trick of negating the value and using `add.1` to save a cycle per iteration)
   - The previous version of the procedure only computed the powers needed for the periodic columns. The procedure was extended to also compute $z^{tracelen}$ which is used in the numerator for the integrity constraint divisor.
   - The cached values are saved into groups, similar to how it is done in the winterfell codebase, this removed a few duplicated `mem_storew`s.
- Added a optimized procedure to get the exemption points.
  - Currently only `2` exemption points are supported by the AirScript compiler. This means the root-of-unity and the exemption points can be precomputed and cached by the compiler. The procedure uses a binary search with a chain of `if-else`s to load the correct precompiled value to the stack.
  - I looked into 2 other alternatives. 1. computing the value at runtime 2. creating a in memory lookup table. Among the alternatives the binary search is the most efficient.
  - The version of the binary search is unbiased. This means that most of the look ups will take the same number of cycles. It is possible to make the search tree unbalanced, and bias the search to be faster for some pre-defined powers. I have not done this but it should be too much work if the target ranges are determined.
- Added a small improvement to the order in which the boundary constraints are evaluated.
   - This is a partial fix for https://github.com/0xPolygonMiden/air-script/issues/315 , but there is some work left to do. Namely the constraint builder needs to sort the values by column, or the information needs to be forwarded to allow sorting in the backend, currently it is likely the wrong composition coefficients are loaded for the boundary constraints.
- Added code to compute the divisor for the integrity constraints.


air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -1,57 +1,189 @@
...
+                end # END else
+            end # END else
+        end # END else
+    end # END else
+end # END PROC get_exemptions_points
+
+proc.cache_z_exp
+    padw mem_loadw.4294903304 drop drop
+    # => [z_1, z_0, ...]
+    # Exponentiate z trace_len times
```
hackaugusto, line 94: The point $z^{tracelen-exemptions}$ for the boundary constraint divisor will be computed by dividing this value twice. This is because this procedure only performs a power-of-two exponentiation (useful to reuse results when period columns are used).

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -1,57 +1,189 @@
+proc.get_exemptions_points
+    mem_load.4294903306
+    # => [trace_len, ...]
+    dup.0 push.131064 u32checked_and neq.0 # Included(16)..Included(3)
+    if.true
```
hackaugusto, line 5: This is a bit intimidating at first, but it is the most performant variant I found. It is effectively a binary search over `trace_len`, it relies on the fact the `trace_len` is always a power-of-two and a u32.

Each branch uses a mask to bisect the trace length to its right size. This is only efficient because of the native support for binary `and`.

The test code exercises every branch by computing the expected value and calling it for the 29 valid input values.

codegen/masm/src/codegen.rs
```diff
@@ -114,15 +200,41 @@ fn quadratic_element_square(writer: &mut Writer, n: u32) {
 
 impl<'ast> CodeGenerator<'ast> {
     pub fn new(ir: &'ast AirIR, config: CodegenConfig) -> CodeGenerator<'ast> {
+        // remove duplicates and sort period lengths in descending order, since larger periods will
+        // give smaller number of cycles (which means a smaller number of exponentiations)
+        let mut periods: Vec<usize> = ir.periodic_columns().iter().map(|e| e.len()).collect();
```
hackaugusto, line 205: Initially some of this code was in the visit methods. But that was starting to create implicit required ordering to the visited fields. To avoid having bugs, I moved the simple code to the constructor, and left the visitor for the iteration that uses the algebraic graph.

codegen/masm/src/codegen.rs
bobbinth, line 1: This file has gotten rather big. We usually try to break files up after 500 - 600 lines. Is there a good way to do it here?

bobbinth (commented):
Not a real review yet - just a coupe of questions/comments.

codegen/masm/src/codegen.rs
hackaugusto: Agreed, I would do that in another PR though, since this is big already, is that okay?

codegen/masm/src/codegen.rs
```diff
@@ -279,14 +509,153 @@ impl<'ast> CodeGenerator<'ast> {
...
+    /// Note:
+    /// - The value `g^{trace_len-1}` is used only as an exemption point for the denominator of
+    /// integrity constraints divisor
+    /// - The value `g^{trace_len-2}` is used both as an exemption point the integrity constraints
+    /// divisor, as well as the offset point for the boundary constraints divisor.
+    fn gen_get_exemptions_points(&mut self) -> Result<(), CodegenError> {
+        // Notes:
+        // - Computing the exemption points on the fly would require 1 exponentiation to find the
+        // root-of-unity from the two-adacity, followed by another exponetiation to compute the
+        // two-to-last exemption point, and a multiplication to compute the last exemption point.
```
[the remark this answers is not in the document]
Al-Kindi-0: Clarification: hardcode means exactly what is done in `proc.get_exemptions_points` but with constants that are defined somewhere else e.g. the constants file.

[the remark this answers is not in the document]
hackaugusto: > Finally, from the trace domain generators, one can get the values at the exception points by at most a division and multiplication. This last claim follows from the following identity:

That would require computing the inverse `g^(-1)`, correct? How expensive is that? I haven't look at it, but if we can get it with anything less than an exponentiation it should be win.

hackaugusto:
splitting this up into smaller PRs.

codegen/masm/src/codegen.rs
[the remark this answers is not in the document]
hackaugusto: The code is now on PR https://github.com/0xPolygonMiden/air-script/pull/320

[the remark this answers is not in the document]
Al-Kindi-0: I think the last comment is what I was trying to get at

PR #317, opened 2023-06-20
Masm boundary constraint divisor

hackaugusto:
Changes from this PR:

- Split the code from the `codegen.rs` file, since it was getting too big.
- Split the procedure to compute the boundary constraints. Instead of having a single procedure to compute all the boundary constraints, now there is one procedure per boundary constraint group. This makes it easier to apply the denominator
- Computes the denominator for each boundary constraint group, and aggregates the result.


hackaugusto:
Closing this PR for the moment. I'm doing this because this is based on another PR that is being split up because of the size, once the underlying PRs are reviewed/merged I will open a new updated version of this.

PR #318, opened 2023-06-20
masm: updated constant to match miden vm changes

hackaugusto:
Synchronizes the constants of the MASM backend with the Miden VM.


bobbinth (approved):
Looks good! Thank you!

PR #319, opened 2023-06-20
Update z^exp caching

hackaugusto:
Review/Merge after #318 (this needs the LOG2 address added by 318)

```
    masm: caching `z^trace_len`

    - The value `z^trace_len` is used in the divisor of the transition
      constraints. This commit extends `cache_z_exp` to compute that value and
      cache the result along size the other points used by the periodic
      columns.
    - This also updates the code to use `log` of the trace_len instead of
      the absolute value. This allowed for some small optimizations and
      reduce stack manipulation.
    - The store/load of the cache value is also optimized, now the periodic
      columns are grouped by their cycle length, and only a single item per
      group is stored. In contrast to the previous implementation that
      stored one element per column, which could include duplicates.
```

### DIFF
--- a/air-script/tests/aux_trace/aux_trace.masm
+++ b/air-script/tests/aux_trace/aux_trace.masm
@@ -1,3 +1,21 @@
+proc.cache_z_exp
+    padw mem_loadw.4294903304 drop drop
+    # => [z_1, z_0, ...]
+    # Exponentiate z trace_len times
+    mem_load.4294903307 neg
+    # => [count, z_1, z_0, ...] where count = -log2(trace_len)
+    dup.0 neq.0
+    while.true
+        movdn.2 dup.1 dup.1 ext2mul
+        # => [(z_1, z_0)^n, i, ...]
+        movup.2 add.1 dup.0 neq.0
+        # => [b, i+1, (z_1, z_0)^n, ...]
+    end # END while
+    push.0 mem_storew.500000100 # z^trace_len
+    # => [0, 0, (z_1, z_0)^trace_len, ...]
+    dropw # Clean stack
+end # END PROC cache_z_exp
+
 proc.compute_evaluate_integrity_constraints
     # integrity constraint 0 for main
     padw mem_loadw.4294900000 drop drop padw mem_loadw.4294900001 movdn.3 movdn.3 drop drop padw mem_loadw.4294900000 movdn.3 movdn.3 drop drop padw mem_loadw.4294900001 movdn.3 movdn.3 drop drop ext2mul padw mem_loadw.4294900002 movdn.3 movdn.3 drop drop ext2mul ext2add ext2sub
--- a/air-script/tests/binary/binary.masm
+++ b/air-script/tests/binary/binary.masm
@@ -1,3 +1,21 @@
+proc.cache_z_exp
+    padw mem_loadw.4294903304 drop drop
+    # => [z_1, z_0, ...]
+    # Exponentiate z trace_len times
+    mem_load.4294903307 neg
+    # => [count, z_1, z_0, ...] where count = -log2(trace_len)
+    dup.0 neq.0
+    while.true
+        movdn.2 dup.1 dup.1 ext2mul
+        # => [(z_1, z_0)^n, i, ...]
+        movup.2 add.1 dup.0 neq.0
+        # => [b, i+1, (z_1, z_0)^n, ...]
+    end # END while
+    push.0 mem_storew.500000100 # z^trace_len
+    # => [0, 0, (z_1, z_0)^trace_len, ...]
+    dropw # Clean stack
+end # END PROC cache_z_exp
+
 proc.compute_evaluate_integrity_constraints
     # integrity constraint 0 for main
     padw mem_loadw.4294900000 movdn.3 movdn.3 drop drop
--- a/air-script/tests/bitwise/bitwise.masm
+++ b/air-script/tests/bitwise/bitwise.masm
@@ -1,38 +1,36 @@
 proc.cache_z_exp
-    # Load trace length
-    mem_load.4294903306
-    # => [trace_len, ...]
-    # Push initial num_of_cycles
-    push.1
-    # => [num_of_cycles, trace_len, ...]
-    # Load Z
     padw mem_loadw.4294903304 drop drop
-    # => [z_1, z_0, num_of_cycles, trace_len, ...]
-    # Compute exponentiations based on the number of cycles for a period of 8
-    dup.3 div.8
-    # => [num_of_cycles', z_1, z_0, num_of_cycles, trace_len, ...]
-    # Update next num_of_cycles and compute number of iterations
-    movup.3 dup.1 movdn.4 div
-    # => [i, z_1, z_0, num_of_cycles', trace_len, ...]
-    # Exponentiate the existing `z**num_of_cycles` an additional `i` times
-    dup.0 neq.1
+    # => [z_1, z_0, ...]
+    # Find number exponentiations required to get for a period of length 8
+    mem_load.4294903307 neg add.3
+    # => [count, z_1, z_0, ...] where count = -log2(trace_len) + 3
+    # Exponentiate z
+    dup.0 neq.0
     while.true
         movdn.2 dup.1 dup.1 ext2mul
-        # => [z_1^2, z_0^2, i, num_of_cycles', trace_len, ...]
-        movup.2 div.2 dup.0 neq.1
-        # => [b, i+1, z_1^2, z_0^2, num_of_cycles', trace_len, ...]
+        # => [(z_1, z_0)^n, i, ...]
+        movup.2 add.1 dup.0 neq.0
+        # => [b, i+1, (z_1, z_0)^n, ...]
     end # END while
-    drop push.0 push.0 mem_storew.500000100 # Save the exponentiation of z for column 0
-    mem_storew.500000101 # Save the exponentiation of z for column 1
-    # => [0, 0, z_1^2, z_0^2, num_of_cycles', trace_len, ...]
+    push.0 mem_storew.500000100 # z^8
+    # => [0, 0, (z_1, z_0)^n, ...] where n = trace_len-8
     drop drop
-    # => [z_1^2, z_0^2, num_of_cycles', trace_len, ...]
-    # Clean stack
-    dropw
+    # Exponentiate z 8 times, until trace_len
+    push.18446744069414584318
+    # => [count, (z_1, z_0)^n, ...] where count=-3 , n=trace_len-3
+    dup.0 neq.0
+    while.true
+        movdn.2 dup.1 dup.1 ext2mul
+        # => [(z_1, z_0)^n, i, ...]
+        movup.2 add.1 dup.0 neq.0
+        # => [b, i+1, (z_1, z_0)^n, ...]
+    end # END while
+    push.0 mem_storew.500000101 # z^trace_len
+    # => [0, 0, (z_1, z_0)^trace_len, ...]
+    dropw # Clean stack
 end # END PROC cache_z_exp
 
 proc.cache_periodic_polys
-    exec.cache_z_exp
     # periodic column 0
     padw mem_loadw.500000100 drop drop
     # => [z_exp_1, z_exp_0, ...]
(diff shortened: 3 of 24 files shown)


air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -1,3 +1,22 @@
+proc.cache_z_exp
+    padw mem_loadw.4294903304 drop drop
+    # => [z_1, z_0, ...]
+    # Exponentiate z trace_len times
+    mem_load.4294903307 neg dup.0 neq.0
+    while.true
+        movdn.2 dup.1 dup.1 ext2mul
+        # => [z_1^2, z_0^2, i, ...]
```
bobbinth, line 8: minor nit on notation: technically, the values on the stack are not `[z_1^2, z_0^2, ...]`, but more like `[(z_0, z_1)^2, ...]`.

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -1,3 +1,22 @@
+proc.cache_z_exp
+    padw mem_loadw.4294903304 drop drop
+    # => [z_1, z_0, ...]
+    # Exponentiate z trace_len times
+    mem_load.4294903307 neg dup.0 neq.0
```
bobbinth, line 5: Here, we are loading not `trace_len` but `log2_trace_len`, right? I wonder if it make sense to mention this in a comment.

codegen/masm/src/codegen.rs
```diff
@@ -136,111 +153,168 @@ impl<'ast> CodeGenerator<'ast> {
...
+    /// quadratic extension elements, and this is an expensive operation.
     ///
-    /// The algorithm computes the exponents from smallest to largest, using the previous result as
-    /// a cache value.
+    /// The generated code is optimized to perform the fewest number of exponentiations, this is
+    /// achieved by observing that periodic columns and trace length are both powers-of-two, since
+    /// the exponent is defined as `exp = trace_len / periodic_column_len`, all exponents are
+    /// themselves powers-of-two. This allows the results to be computed from smallest to largest,
+    /// re-using the intermediary values.
     fn gen_cache_z_exp(&mut self) -> Result<(), CodegenError> {
```
bobbinth, line 167: Should we move this function into a separate file? It could still be a part of the `CodeGenerator` impl - defined in its own file. It actually may make sense to create a file for all "boilerplate" code generation (e.g., `boilerplate.rs`).

bobbinth (approved):
Looks great! Thank you! I left a couple of non-blocking nits inline.

One other thing (and this could be more for the future): I wonder if we should emit doc comments to procedures like `cache_z_exp`. This could make it a bit easier to review the generated MASM files.

air-script/tests/aux_trace/aux_trace.masm
Al-Kindi-0: minor nit: the same is done [here](https://github.com/0xPolygonMiden/miden-vm/blob/4576337fbc8702364e01d1a505453f033048a5eb/stdlib/asm/crypto/stark/random_coin.masm#L422), so we should probably leave a TODO to avoid duplication

air-script/tests/binary/binary.masm
```diff
@@ -1,3 +1,22 @@
...
+        # => [b, i+1, z_1^2, z_0^2, ...]
+    end # END while
+    push.0 mem_storew.500000100 # z^exp for trace_len
+    # => [0, 0, z_1^2, z_0^2, ...]
+    drop drop
+    # => [z_1^2, z_0^2, num_of_cycles', -log2(trace_len), ...]
+    # Clean stack
+    dropw
+end # END PROC cache_z_exp
+
```
Al-Kindi-0, line 19: curious why the implementation of `cache_z_exp` in two places

air-script/tests/bitwise/bitwise.masm
```diff
@@ -1,38 +1,37 @@
...
+    while.true
+        movdn.2 dup.1 dup.1 ext2mul
+        # => [z_1^2, z_0^2, i, ...]
+        movup.2 add.1 dup.0 neq.0
+        # => [b, i+1, z_1^2, z_0^2, ...]
+    end # END while
+    push.0 mem_storew.500000101 # z^exp for trace_len
+    # => [0, 0, z_1^2, z_0^2, ...]
+    drop drop
+    # => [z_1^2, z_0^2, num_of_cycles', -log2(trace_len), ...]
```
Al-Kindi-0, line 29: I am a bit lost here, what are we computing?

air-script/tests/bitwise/bitwise.masm
```diff
@@ -1,38 +1,37 @@
...
     # Clean stack
     dropw
 end # END PROC cache_z_exp
 
 proc.cache_periodic_polys
-    exec.cache_z_exp
     # periodic column 0
     padw mem_loadw.500000100 drop drop
     # => [z_exp_1, z_exp_0, ...]
```
Al-Kindi-0, line 37: nit: if one of the elements is in the base field, then `ext2mul` can be substituted by two base field multiplications. This is basically the `<Felt as ExtensibleField<2>>::mul_base` method

codegen/masm/src/codegen.rs
```diff
@@ -58,15 +64,18 @@ pub struct CodeGenerator<'ast> {
...
-/// of the word must be kept. Values of periodic columns are stored at distinct memory addresses
-/// such that each value occupies a single memory word with the two most significant word elements
-/// set to zeros (i.e., [q0, q1, 0, 0])
-fn periodic_column_to_target_el(column: u32) -> u32 {
-    // each period column has its own address, and the element is in the lower half
-    (column * 2) + 1
+/// Periodic columns are grouped based on their length, this is done so that only a single z value
+/// needs to be cached per group. The grouping is based on unique lengths, sorted from highest to
+/// lowest. Given a periodic group, this function will return a memory offset, which can be used to
+/// load the correspoding z value.
```
Al-Kindi-0, line 72: minor nit: corresponding

codegen/masm/src/codegen.rs
```diff
@@ -58,15 +64,18 @@ pub struct CodeGenerator<'ast> {
...
+fn periodic_group_to_memory_offset(group: u32) -> u32 {
+    // Each memory address contains a single quadraic element, this makes the code to store/load the
```
Al-Kindi-0, line 74: minor nit: element of a quadratic field extension

codegen/masm/src/codegen.rs
```diff
@@ -136,111 +153,168 @@ impl<'ast> CodeGenerator<'ast> {
...
+    /// This procedure exists because the VM doesn't have native instructions for exponentiation of
+    /// quadratic extension elements, and this is an expensive operation.
     ///
-    /// The algorithm computes the exponents from smallest to largest, using the previous result as
-    /// a cache value.
+    /// The generated code is optimized to perform the fewest number of exponentiations, this is
+    /// achieved by observing that periodic columns and trace length are both powers-of-two, since
+    /// the exponent is defined as `exp = trace_len / periodic_column_len`, all exponents are
+    /// themselves powers-of-two. This allows the results to be computed from smallest to largest,
+    /// re-using the intermediary values.
```
Al-Kindi-0, line 166: I think this is the answer to my question above. Basically, we compute `z^n` from `z^m`, where `n` and `m` are two periodic lengths sorted in increasing order, using `log2(n-m)` multiplications?
Can we add a comment to the generated procedures to clarify this (or the correct reason if the above is wrong)?

codegen/masm/src/codegen.rs
```diff
@@ -136,111 +153,168 @@ impl<'ast> CodeGenerator<'ast> {
...
-                divisor
-            ));
-            self.writer.dup(3);
-            self.writer.div(Some(*divisor));
-            self.writer
-                .header("=> [num_of_cycles', z_1, z_0, num_of_cycles, trace_len, ...]");
+        self.load_z();
+        self.writer.header("=> [z_1, z_0, ...]");
+
+        // The loop below needs to mutable borrow the codegen, so take the field for the iteration
```
Al-Kindi-0, line 181: minor nit: mutably

Al-Kindi-0 (changes requested):
Looks great, thank you @hackaugusto !
Minor nits from my part.

air-script/tests/aux_trace/aux_trace.masm
hackaugusto: > minor nit on notation

changed

> the same is done [here](https://github.com/0xPolygonMiden/miden-vm/blob/4576337fbc8702364e01d1a505453f033048a5eb/stdlib/asm/crypto/stark/random_coin.masm#L422)

this is slightly different. Depending on the Air definition this will compute intermediary values used to evaluate the periodic column polynomial. Here is an [example](https://github.com/0xPolygonMiden/air-script/blob/714fe2f5bbccddcc1cfad861858c788253bbeb82/air-script/tests/bitwise/bitwise.masm#L4-L15).

I created an issue on the miden-vm repo https://github.com/0xPolygonMiden/miden-vm/issues/952

air-script/tests/aux_trace/aux_trace.masm
hackaugusto: Added a comment for the stack state:

```
    # Exponentiate z trace_len times
    mem_load.4294903307 neg
    # => [count, z_1, z_0, ...] where count = -log2(trace_len)
```

codegen/masm/src/codegen.rs
hackaugusto: This was requested on a previous PR https://github.com/0xPolygonMiden/air-script/pull/316#discussion_r1234892268 , I have done that, but in a commit downstream. I would prefer to not do any code reorganization now since that is not the goal of this PR and that will save me work on conflicts later on.

air-script/tests/binary/binary.masm
hackaugusto: This file has only one `cache_z_exp` procedure, each file gets its own version because this is a generated procedure by the backend.

air-script/tests/bitwise/bitwise.masm
hackaugusto: The `num_of_cycles', -log2(trace_len)` was an outdated state of the stack, sorry about that, I have updated the code to show the correct stack instead.

codegen/masm/src/codegen.rs
bobbinth: Yep - sounds good!

air-script/tests/bitwise/bitwise.masm
hackaugusto: I'm not sure I follow:

1. Are you suggesting something like `[(q1, q0), b, ...]` where `(q1, q0)` is an quadratic extension, and `b` a base field, that we should use `dup.2 mul swap movup.2 mul swap` instead of `ext2mul`? That would be `6` cycles against `5` from `movup.2 push.0 ext2mul`
2. This procedure should be using only quadratic elements, it corresponds only to the [polynomial evaluation](https://github.com/facebook/winterfell/blob/main/verifier/src/evaluator.rs#L34). Unless I made a mistake that should be only extension field elements, no?

codegen/masm/src/codegen.rs
hackaugusto: Thanks! Fixed.

codegen/masm/src/codegen.rs
hackaugusto: fixed :)

codegen/masm/src/codegen.rs
hackaugusto: > I think this is the answer to my question above. Basically, we compute z^n from z^m, where n and m are two periodic lengths sorted in increasing order, using log2(n-m) multiplications?

Exactly.

> Can we add a comment to the generated procedures to clarify this (or the correct reason if the above is wrong)?

I updated the comments of the generated code, let me know if that is enough (I have done small changes, I want to know if you want a more verbose text)

codegen/masm/src/codegen.rs
hackaugusto: changed

codegen/masm/src/codegen.rs
Al-Kindi-0: Looks good to me, thank you!

air-script/tests/bitwise/bitwise.masm
Al-Kindi-0: I thought that avoiding `ext2mul` will be better but as you showed it is worse than your solution. Thank you for the clarification.
Yes, the evaluation point is an element of the quadratic extension field.

PR #320, opened 2023-06-20
masm: added procedure to load exemption points

hackaugusto:
The generated procedure takes the approach of pre-computing the values during compilations, and then searching for the correct exemption points during runtime, doing a binary search on the runtime trace length to find the correct points.

The alternative of computing the points during runtime or creating a lookup table would have worked, but both approaches would be sligthly slower in comparison.

The data is kept in the stack for 1. testability 2. performance, since mem load/stores are no longer necessary.

### DIFF
--- a/codegen/masm/src/codegen.rs
+++ b/codegen/masm/src/codegen.rs
@@ -4,8 +4,9 @@ use air_ir::{
 };
 use std::collections::BTreeMap;
 use std::mem::{replace, take};
+use std::ops::{Bound, RangeBounds};
 
-use miden_processor::math::{Felt, StarkField};
+use miden_processor::math::{Felt, FieldElement, StarkField};
 use winter_prover::math::fft;
 
 use crate::config::CodegenConfig;
@@ -119,6 +120,83 @@ fn load_quadratic_element(
     }
 
     Ok(())
+}
+
+/// Precomputes the exemption points for a given power.
+///
+/// The current version of the generated code has hardcoded 2 exemption points, this means the
+/// points can be precomputed during compilation time to save a few cycles during runtime.
+fn points_for_power(power: u32) -> (u64, u64) {
+    let g = Felt::get_root_of_unity(power);
+    let trace_len = 2u64.pow(power);
+    let one = g.exp(trace_len - 1).as_int();
+    let two = g.exp(trace_len - 2).as_int();
+    (one, two)
+}
+
+/// Generate code to push the exemptions point to the top of the stack.
+///
+/// This procedure handles two powers using conditional drops, instead of control flow with if
+/// statements, since the former is slightly faster for the small number of instructions used. The
+/// emitted code assumes the trace_length is at the top of the stack, and afer executing it will
+/// leave the stack as follows:
+///
+/// Stack: [g^{trace_len-2}, g^{trace_len-1}, ...]
+fn exemption_points(writer: &mut Writer, small_power: u32) {
+    let (lone, ltwo) = points_for_power(small_power);
+    let (hone, htwo) = points_for_power(small_power + 1);
+
+    writer.push(2u64.pow(small_power));
+    writer.u32checked_and();
+    writer.neq(0);
+    writer.comment(format!(
+        "Test if trace length is a power of 2^{}",
+        small_power
+    ));
+
+    writer.push(hone);
+    writer.push(lone);
+    writer.dup(2);
+    writer.cdrop();
+
+    writer.push(htwo);
+    writer.push(ltwo);
+    writer.movup(3);
+    writer.cdrop();
+}
+
+/// Helper function to emit efficient code to bisect the trace length value.
+///
+/// The callbacks `yes` and `no` are used to emit the code for each branch.
+fn bisect_trace_len<L, R>(writer: &mut Writer, range: impl RangeBounds<u32>, yes: L, no: R)
+where
+    L: FnOnce(&mut Writer),
+    R: FnOnce(&mut Writer),
+{
+    let mask = match (range.end_bound(), range.start_bound()) {
+        (Bound::Included(&start), Bound::Included(&end)) => {
+            let high_mask = 2u64.pow(start + 1) - 1;
+            let low_mask = 2u64.pow(end) - 1;
+            high_mask ^ low_mask
+        }
+        _ => panic!("Only inclusive ranges are supported"),
+    };
+
+    writer.dup(0);
+    writer.push(mask);
+    writer.u32checked_and();
+    writer.neq(0);
+    writer.comment(format!(
+        "{:?}..{:?}",
+        range.start_bound(),
+        range.end_bound()
+    ));
+
+    writer.r#if();
+    yes(writer);
+    writer.r#else();
+    no(writer);
+    writer.r#end();
 }
 
 /// Assumes a quadratic element is at the top of the stack and square it `n` times.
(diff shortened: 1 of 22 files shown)


hackaugusto:
Relevant previous comment: https://github.com/0xPolygonMiden/air-script/pull/316#discussion_r1234879092

bobbinth:
> Relevant previous comment: [#316 (comment)](https://github.com/0xPolygonMiden/air-script/pull/316#discussion_r1234879092)

Based on this discussion, should we update the approach? If I'm understanding things correctly, we'd need to do the following:

1. Update code in `stdlib` of Miden VM to save $g$ to some well-known memory location.
2. Use this $g$ here to compute two exemption points using division as $g^{N-1} = \frac{1}{g}$ and and multiplication as $g^{N-2} = \frac{g^{N-1}}{g}$.

Al-Kindi-0:
We can use the binary-search based approach that @hackaugusto came up with in order to look-up the LDE domain generator $g_l$. Given that, the trace domain generator (should be double checked but I am quiet sure) is just $g := g_l^8$. Given $g$, and trace length $N$, we can get $g^{N-1}$ as $g^{-1}$ and $g^{N-2}$ as $g^{-1}*g^{-1}$

codegen/masm/src/codegen.rs
```diff
@@ -103,6 +104,83 @@ fn load_quadratic_element(
...
+    let two = g.exp(trace_len - 2).as_int();
+    (one, two)
+}
+
+/// Generate code to push the exemptions point to the top of the stack.
+///
+/// This procedure handles two power using conditional drops, instead of control flow with if
+/// statements, since the former is slightly faster for the small number of instructions used. The
+/// emitted code assumes the trace_length is at the top of the stack, and afer executing it will
+/// leave leave the stack as follows:
```
Al-Kindi-0, line 124: minor nit: leave (single)

codegen/masm/src/codegen.rs
```diff
@@ -103,6 +104,83 @@ fn load_quadratic_element(
...
+    let g = Felt::get_root_of_unity(power);
+    let trace_len = 2u64.pow(power);
+    let one = g.exp(trace_len - 1).as_int();
+    let two = g.exp(trace_len - 2).as_int();
+    (one, two)
+}
+
+/// Generate code to push the exemptions point to the top of the stack.
+///
+/// This procedure handles two power using conditional drops, instead of control flow with if
```
Al-Kindi-0, line 121: minor nit: "... handles two powers"

codegen/masm/src/codegen.rs
```diff
@@ -287,6 +365,140 @@ impl<'ast> CodeGenerator<'ast> {
         Ok(())
     }
 
+    /// Emits code for the procedure `get_exemptions_points`.
+    ///
+    /// The generated procedure contains the precomputed exeption points to be used when computing
```
Al-Kindi-0, line 370: minor nit: exception

codegen/masm/src/codegen.rs
```diff
@@ -287,6 +365,140 @@ impl<'ast> CodeGenerator<'ast> {
...
+    /// the divisors. The values are returned in the stack.
+    fn gen_get_exemptions_points(&mut self) -> Result<(), CodegenError> {
+        // Notes:
+        // - Computing the exemption points on the fly would require 1 exponentiation to find the
+        // root-of-unity from the two-adacity, followed by another exponetiation to compute the
```
Al-Kindi-0, line 375: minor nit: adicity

codegen/masm/src/codegen.rs
Al-Kindi-0: or exemption (first time I notice they are close in Hamming distance :smile:)

codegen/masm/src/codegen.rs
```diff
@@ -287,6 +365,140 @@ impl<'ast> CodeGenerator<'ast> {
...
+        // two-to-last exemption point, and a multiplication to compute the last exemption point.
+        // Each exponentiation is 41 cycles, giving around 83 cycles to compute the values.
+        // - For the range from powers 3 to 32 there are 30 unique values, which requires 8 words
+        // of data. Storing the data to memory requires pushing the 4 elements of a word to the
```
Al-Kindi-0, line 379: I think the ranges need to be update, either for the current approach or the other proposed one.

Al-Kindi-0 (commented):
Looks great @hackaugusto , I like the binary search based approach and I wonder if we can apply it to the LDE domain (Low-degree extension domain i.e. it has size equal to the trace domain times a blow-up factor of $2^3$ (with an offset which is not important here)) instead to gain even further efficiency.

codegen/masm/src/codegen.rs
hackaugusto: fixed

codegen/masm/src/codegen.rs
hackaugusto: fixed

codegen/masm/src/codegen.rs
hackaugusto: fixed

hackaugusto:
@bobbinth @Al-Kindi-0 to clarify, the idea is to:

1. rename `get_exemption_points` to something like `get_constants`
2. the procedure will be:

```
# Input: [...]
# Output: [g, g', e1, e0]
#  where:
#   g is the LDE generator
#   g' is the domain generator
#   e1 is g'^{trace_len-1}
#   e0 is g'^{trace_len-2}
prod.get_constants
  # binary search ...
end
```

hackaugusto:
@bobbinth @Al-Kindi-0 anything else needs to be done on this PR? This procedure is a dependency for the rest of the divisor code

Al-Kindi-0:
> @bobbinth @Al-Kindi-0 to clarify. Is the idea is to the following?
> 
>     1. rename `get_exemption_points` to something like `get_constants`
> 
>     2. the procedure will be:
> 
> 
> ```
> # Input: [trace_len, ...]
> # Output: [g, g', e1, e0]
> #  where:
> #   g is the LDE generator
> #   g' is the domain generator
> #   e1 is g'^{trace_len-1}
> #   e0 is g'^{trace_len-2}
> prod.get_constants
>   # binary search ...
> end
> ```

Yes, it makes sense to me. The main work of the procedure is getting the the LDE domain generator corresponding to the LDE domain using the binary search approach. Once that is found, we can get the other 3 by multiplications and an inversion. The trace domain generator is just the LDE domain generator to the power $2^3$ and the exemption points are found by a simple inversion and a multiplication.
If I compute correctly, the total cost is the cost of the binary search plus $(3 + 1)$ multiplications plus $1$ inversion in the field.

bobbinth:
> 1. rename `get_exemption_points` to something like `get_constants`
> 2. the procedure will be:
> 
> ```
> # Input: [trace_len, ...]
> # Output: [g, g', e1, e0]
> #  where:
> #   g is the LDE generator
> #   g' is the domain generator
> #   e1 is g'^{trace_len-1}
> #   e0 is g'^{trace_len-2}
> prod.get_constants
>   # binary search ...
> end
> ```

I guess one question for me is whether there is significant benefit for doing binary search. If we are saving something like a dozen of cycles or so, I'd still probably prefer a simple exponentiation because it make the output much easier to reason about.

hackaugusto:
changing to the exponentiation approach

PR #321, opened 2023-06-20
masm: moved map creation to constructor

hackaugusto:
the previous code had a required and implicit ordering, were the public input and constants had to be visited first to create the mappings, so that later on the walk over the constraints would have the data available to emit the required code. this implicit dependency makes the code harder to maintain, so this commit moves the construction of the mapping to the constructor instead.

### DIFF
--- a/codegen/masm/src/codegen.rs
+++ b/codegen/masm/src/codegen.rs
@@ -43,9 +43,6 @@ pub struct CodeGenerator<'ast> {
     /// Maps the public input to their start offset.
     public_input_to_offset: BTreeMap<String, usize>,
 
-    /// Holds the count of public inputs seen so far, this is used to compute the offset.
-    public_input_count: usize,
-
     /// Map of the constants found while visitint the [AirIR].
     ///
     /// These values are later used to emit immediate values.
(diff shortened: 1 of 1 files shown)


codegen/masm/src/codegen.rs
```diff
@@ -418,15 +440,9 @@ impl<'ast> AirVisitor<'ast> for CodeGenerator<'ast> {
...
+        _constant: &'ast ConstantBinding,
     ) -> Result<Self::Value, Self::Error> {
-        match self.constants.entry(constant.name()) {
-            Entry::Occupied(_) => Err(CodegenError::DuplicatedConstant),
-            Entry::Vacant(entry) => {
-                entry.insert(constant.value());
-                Ok(())
-            }
-        }
+        Ok(())
```
bobbinth, line 445: Since this and `visit_public_input` methods don't really do anything any more, should we just get rid of them?

bobbinth (approved):
Looks good! Thank you! I left one non-blocking comment inline.

codegen/masm/src/codegen.rs
```diff
@@ -114,15 +111,40 @@ fn quadratic_element_square(writer: &mut Writer, n: u32) {
 
 impl<'ast> CodeGenerator<'ast> {
     pub fn new(ir: &'ast AirIR, config: CodegenConfig) -> CodeGenerator<'ast> {
+        // remove duplicates and sort period lengths in descending order, since larger periods will
+        // give smaller number of cycles (which means a smaller number of exponentiations)
```
Al-Kindi-0, line 115: minor nit: "will have smaller" sounds better to me

codegen/masm/src/codegen.rs
```diff
@@ -114,15 +111,40 @@ fn quadratic_element_square(writer: &mut Writer, n: u32) {
...
+        let mut periods: Vec<usize> = ir.periodic_columns().iter().map(|e| e.len()).collect();
+        periods.sort();
+        periods.dedup();
+        periods.reverse();
+
+        // map from public input name to start offset, used to compute the memory location of the
```
Al-Kindi-0, line 121: minor nit: can we reformulate the sentence? I am not getting `map from public input name to start offset`

Al-Kindi-0 (commented):
Looks good to me, thank you @hackaugusto !

codegen/masm/src/codegen.rs
hackaugusto: I expect a correct visitor pattern implementation to walk the whole structure, that is why I left that in the trait and as a consequence left it here too.

I guess this is not being used for anything else, so it can be removed

bobbinth: Yeah - I think starting with a generic approach is a fine, but as we learn more about how we need process the structure (i.e., some elements are processed at construction time, other elements may be processed together), I think it is OK to make things more specific.

hackaugusto: removed unused methods.

codegen/masm/src/codegen.rs
hackaugusto: changed

codegen/masm/src/codegen.rs
hackaugusto: updated

hackaugusto: Note: I'm going to merge this so that I can open the follow ups. Let me know if the comment still needs to be improved and I will open a follow up.

PR #322, opened 2023-06-20
masm: sorting boundary constraints

hackaugusto:
The Winterfell framework groups assertions by trace (main or auxiliary), by stride, and step (the first row to which the assertion is enforced). The assertions in a group are further ordered by the column which they apply to.

In AirScript only single value assertions are supported, which means the stride is always `0` and grouping only happens by trace and step. And they are further ordered by column.

The ordering described above is important because it defines the position of the composition coefficient to be used.

This commits introduces the ordering by `step`. The ordering by column needs to be implemented later on.

### DIFF
--- a/codegen/masm/src/codegen.rs
+++ b/codegen/masm/src/codegen.rs
@@ -1,7 +1,8 @@
 use crate::constants::{AUX_TRACE, MAIN_TRACE};
 use crate::visitor::{
-    walk_boundary_constraints, walk_constant_bindings, walk_integrity_constraint_degrees,
-    walk_integrity_constraints, walk_periodic_columns, walk_public_inputs, AirVisitor,
+    walk_boundary_constraints_in_natural_order, walk_constant_bindings,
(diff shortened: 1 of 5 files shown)


bobbinth (approved):
Looks good! Thank you!

codegen/masm/src/codegen.rs
```diff
@@ -386,6 +385,8 @@ impl<'ast> AirVisitor<'ast> for CodeGenerator<'ast> {
         self.writer
             .header("Multiply by the composition coefficient");
 
+        // Note: The corretness of the load below relies on the integrity constraint being iterated
```
Al-Kindi-0, line 388: minor nit: correctness

Al-Kindi-0 (commented):
Looks good, thank you!

codegen/masm/src/codegen.rs
hackaugusto: thanks, fixed :)

PR #323, opened 2023-06-26
masm: added get_exemptions_points procedure

hackaugusto:
follow up to https://github.com/0xPolygonMiden/air-script/pull/320

This adds a `get_exemption_points` procedure which relies on the pre computed generator from the FRI stdlib.

### DIFF
--- a/codegen/masm/src/codegen.rs
+++ b/codegen/masm/src/codegen.rs
@@ -397,6 +397,30 @@ impl<'ast> Backend<'ast> {
         Ok(())
     }
 
+    /// Emits code for the procedure `get_exemptions_points`.
+    ///
+    /// Generate code to push the exemption points to the top of the stack.
+    /// Stack: [g^{trace_len-2}, g^{trace_len-1}, ...]
+    fn gen_get_exemptions_points(&mut self) -> Result<(), CodegenError> {
+        self.writer.proc("get_exemptions_points");
+        self.load_trace_domain_generator();
+        self.writer.header("=> [g, ...]");
+
+        self.writer.push(1);
+        self.writer.swap();
+        self.writer.div();
+        self.writer.header("=> [g^{-1}, ...]");
+
+        self.writer.dup(0);
+        self.writer.dup(0);
+        self.writer.mul();
+        self.writer.header("=> [g^{-2}, g^{-1}, ...]");
+
+        self.writer.end(); // end proc
+
+        Ok(())
+    }
+
     /// Emits code for the procedure `evaluate_integrity_constraints`.
     ///
     /// Evaluates the integrity constraints for both the main and auxiliary traces.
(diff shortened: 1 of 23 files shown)


codegen/masm/src/config.rs
```diff
@@ -51,6 +51,9 @@ pub struct CodegenConfig {
     /// Memory range used to store exponentiations of Z, each address contains one point to be used
     /// on the evaluation of each periodic polynimal.
     pub z_exp_address: u32,
+
+    /// Memory position of the domain generator
+    pub domain_generator_address: u32,
```
bobbinth, line 56: This is the LDE domain generator, right? I may be good to differentiate between LDE and trace domain generator (the relationship is $g_{lde}^3 = g_{trace}$, I believe).

bobbinth (approved):
Looks good! Thank you! I left one non-blocking comment inline.

codegen/masm/src/config.rs
Al-Kindi-0: It is $g_{lde}^{2^3} == g_{trace}$, basically  $g_{lde}^{blowup} == g_{trace}$

bobbinth: ah - yes! Sorry for the confusion!

Al-Kindi-0 (commented):
Looks great, thank you!
As @bobbinth suggested, it makes sense to have `TRACE_DOMAIN_GENERATOR` and `LDE_DOMAIN_GENERATOR`.

codegen/masm/src/config.rs
hackaugusto: > This is the LDE domain generator, right?

This is the trace domain generator. I will add a `trace_` prefix.

hackaugusto:
Requested another review to make sure the change cleared any confusions.

codegen/masm/src/codegen.rs
```diff
@@ -397,6 +397,30 @@ impl<'ast> Backend<'ast> {
         Ok(())
     }
 
+    /// Emits code for the procedure `get_exemptions_points`.
+    ///
+    /// Generate code to push the exemptions point to the top of the stack.
```
Al-Kindi-0, line 402: minor nit: "... push the exemption points..."

codegen/masm/src/codegen.rs
```diff
@@ -458,6 +482,12 @@ impl<'ast> Backend<'ast> {
         self.writer.drop();
         self.writer.drop();
     }
+
+    /// Emits code to load `g`, the trace domain generator.
```
Al-Kindi-0, line 486: should we also add a suffix to g here?

Al-Kindi-0 (approved):
Looks great, thank you!

codegen/masm/src/codegen.rs
hackaugusto: I can change from `g` to something else, I'm not sure what would be a good name though. Any suggestions?

bobbinth (approved):
All looks good! Thank you!

PR #326, opened 2023-06-29
masm: added boundary constraint divisor computation

hackaugusto:
- The procedure to evalute the boundary constraints has been split, since different boundary constraints groups have different divisors.
- The divisor is applied and the points are accumulated

### DIFF
--- a/air-script/tests/aux_trace/aux_trace.masm
+++ b/air-script/tests/aux_trace/aux_trace.masm
@@ -52,6 +52,7 @@ proc.compute_integrity_constraint_divisor
     # => [z_1, z_0, zt_1-1, zt_0-1, ...]
     exec.get_exemptions_points
     # => [g^{trace_len-2}, g^{trace_len-1}, z_1, z_0, zt_1-1, zt_0-1, ...]
+    dup.0 mem_store.500000101 # Save a copy of `g^{trace_len-2} to be used by the boundary divisor
     dup.3 dup.3 movup.3 push.0 ext2sub
     # => [e_1, e_0, g^{trace_len-1}, z_1, z_0, zt_1-1, zt_0-1, ...]
     movup.4 movup.4 movup.4 push.0 ext2sub
@@ -71,8 +72,8 @@ end # END PROC compute_integrity_constraint_divisor
 # Input: [...]
 # Output: [(r_1, r_0)*, ...]
 # where: (r_1, r_0) is the quadratic extension element resulting from the integrity constraint evaluation.
-#        This procedure pushes 5 quadratic elements to the stack
-proc.compute_evaluate_integrity_constraints
+#        This procedure pushes 5 quadratic extension field elements to the stack
+proc.compute_integrity_constraints
     # integrity constraint 0 for main
     padw mem_loadw.4294900000 drop drop padw mem_loadw.4294900001 movdn.3 movdn.3 drop drop padw mem_loadw.4294900000 movdn.3 movdn.3 drop drop padw mem_loadw.4294900001 movdn.3 movdn.3 drop drop ext2mul padw mem_loadw.4294900002 movdn.3 movdn.3 drop drop ext2mul ext2add ext2sub
     # Multiply by the composition coefficient
@@ -93,19 +94,14 @@ proc.compute_evaluate_integrity_constraints
     padw mem_loadw.4294900073 movdn.3 movdn.3 drop drop padw mem_loadw.4294900073 drop drop padw mem_loadw.4294900002 movdn.3 movdn.3 drop drop padw mem_loadw.4294900150 movdn.3 movdn.3 drop drop ext2add ext2mul ext2sub
     # Multiply by the composition coefficient
     padw mem_loadw.4294900202 movdn.3 movdn.3 drop drop ext2mul
-end # END PROC compute_evaluate_integrity_constraints
+end # END PROC compute_integrity_constraints
 
-# Procedure to evaluate numerators of all boundary constraints.
-#
-# All the 2 main and 4 auxiliary constraints are computed.
-# The result constriants are evaluated in natural order, and their values are pushed to the stack.
-# The natural order is defined by `(trace, row, column)`, and the final evaluation is in stack-order.
+# Procedure to evaluate the boundary constraint numerator for the first row of the main trace
 #
 # Input: [...]
 # Output: [(r_1, r_0)*, ...]
-# where: (r_1, r_0) is the quadratic extension element resulting from the boundary constraint evaluation.
-#        This procedure pushes 6 quadratic elements to the stack
-proc.compute_evaluate_boundary_constraints
+# Where: (r_1, r_0) is one quadratic extension field element for each constraint
+proc.compute_boundary_constraints_main_first
     # boundary constraint 0 for main
     padw mem_loadw.4294900000 movdn.3 movdn.3 drop drop push.1 push.0 ext2sub
     # Multiply by the composition coefficient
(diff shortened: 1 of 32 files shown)


air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -66,9 +67,14 @@ proc.compute_evaluate_integrity_constraints
...
-end # END PROC compute_evaluate_integrity_constraints
+end # END PROC compute_integrity_constraints
 
-proc.compute_evaluate_boundary_constraints
+# Procedure to compute the boundary divisor for the first row of the main trace
+# 
+# Input: [...]
+# Output: [(r_1, r_0)*, ...]
+# Where: (r_1, r_0) is one quadratic element for each constraint
+proc.compute_boundary_constraints_main_first
```
bobbinth, line 77: I think the comment is wrong here (and for other similar procedures): we are computing boundary constraint numerators, not divisors here.

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -105,8 +135,25 @@ proc.evaluate_integrity_constraints
...
+    padw mem_loadw.4294903304 drop drop push.1 push.0 ext2sub
+    ext2div # divide the numerator by the divisor
+    exec.compute_boundary_constraints_main_first
+    # Accumulate the numerator for segment 0 FirstRow
+    ext2add ext2add
+    # Compute the denominator for segment 0 FirstRow
+    padw mem_loadw.4294903304 drop drop push.1 push.0 ext2sub
+    ext2div # divide the numerator by the divisor
+    # Accumulate point of the constraint polynomial
+    ext2add ext2add
```
bobbinth, line 157: Denominators for both main and aux segments are the same in the same row (e.g., line 148 is the same as line 154). So, we need to compute denominators for the first and last rows only once.

We can probably precompute denominator, keep it on the stack, and then compute the numerator and perform division (may be faster than saving denominator to memory).

bobbinth (approved):
Looks good! Thank you! I left a couple of comments inline. Once these are addressed, we can merge.

Also, one optimization note for the future: sometimes, boundary constraints are the same for the first and last row in the same column. For example, we may assert that the first and last value in a given column is equal to `ONE`.

In such case, the numerators are identical, and the only thing that differs between these constraints is denominators. With the current structure, we'd be computing numerators twice. Ideally, we should avoid it - but we should probably do it as a part of a more wholistic optimization in the future.

air-script/tests/aux_trace/aux_trace.masm
hackaugusto: ah, thanks! fixed

air-script/tests/aux_trace/aux_trace.masm
hackaugusto: > may be faster than saving denominator to memory

The denominator is not saved to memory, the load is for the `z`.

> Denominators for both main and aux segments are the same in the same row ... keep it on the stack

The values are computed using natural order, so:

1. main trace, first row
2. main trace, last row
3. aux trace, first row
4. aux trace, last row

to keep the values is the stack from steps 1/2 to steps 3/4 would require too many operations for it to be worth it (this particular example has very few constraints, for a bigger system it is not a profitable transformation)

hackaugusto: I meant that the loads in this part of the code are used to load either the value of `z` or the value of `g^{trace_len-2}`, not that it loaded the division of one by the other. Sorry for the confusion.

hackaugusto: Okay, I went over the code once again, and I was wrong. Only the codegen has to be iterated in the boundary constraint natural order (necessary to compute the correct offsets into the frames when loading data). But once the procedures are emitted the code will be fine, so the call order here doesn't have to match the natural order.

So I can implement what you guys are suggesting. I'm changing it.

hackaugusto: updated

hackaugusto:
humm, it seems there is a test failing. I'm looking at it

bobbinth (changes requested):
Thank you! I left some comments inline. Basically, it seems like some procedures are not being called from the main constraint evaluation function.

Also, could we modify one of the test `.air` files to include all 4 types of constraints (main/aux for first/last rows)? I've tried to find an example which has all 4 such constraints but upon a quick scan didn't see anything.

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -93,19 +94,14 @@ proc.compute_evaluate_integrity_constraints
...
     padw mem_loadw.4294900202 movdn.3 movdn.3 drop drop ext2mul
-end # END PROC compute_evaluate_integrity_constraints
+end # END PROC compute_integrity_constraints
 
-# Procedure to evaluate numerators of all boundary constraints.
-#
-# All the 2 main and 4 auxiliary constraints are computed.
-# The result constriants are evaluated in natural order, and their values are pushed to the stack.
-# The natural order is defined by `(trace, row, column)`, and the final evaluation is in stack-order.
+# Procedure to evaluate the boundary numerator for the first row of the main trace
```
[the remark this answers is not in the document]
hackaugusto: changed

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -114,6 +110,14 @@ proc.compute_evaluate_boundary_constraints
...
     # Multiply by the composition coefficient
     padw mem_loadw.4294900203 movdn.3 movdn.3 drop drop ext2mul
+end # END PROC compute_boundary_constraints_main_first
+
+# Procedure to evaluate the boundary divisor for the first row of the auxiliary trace
+#
+# Input: [...]
+# Output: [(r_1, r_0)*, ...]
+# Where: (r_1, r_0) is one quadratic extension field element for each constraint
+proc.compute_boundary_constraints_aux_first
```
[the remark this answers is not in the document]
hackaugusto: yep, you're right, changed

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -152,8 +164,23 @@ end # END PROC evaluate_integrity_constraints
...
+    exec.compute_boundary_constraints_aux_first
+    # Accumulate the numerator for segment 1 FirstRow
+    ext2add ext2add
+    # => [(aux_first1, aux_first0), ...]
+    # Compute the denominator for domain FirstRow
+    padw mem_loadw.4294903304 drop drop push.1 push.0 ext2sub
+    # Compute numerator/denominator for first row
+    ext2div
+    # Add first and last row groups
+    ext2add
```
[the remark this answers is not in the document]
hackaugusto: There was indeed a bug, thank you! I have fixed it and added a test.

air-script/tests/aux_trace/aux_trace.masm
```diff
@@ -152,8 +172,25 @@ end # END PROC evaluate_integrity_constraints
...
+    # Accumulate the numerator for segment 1 FirstRow
+    ext2add ext2add
+    # Compute the denominator for segment 1 FirstRow
+    padw mem_loadw.4294903304 drop drop push.1 push.0 ext2sub
+    ext2div # divide the numerator by the divisor
+    exec.compute_boundary_constraints_main_first
+    # Accumulate the numerator for segment 0 FirstRow
+    ext2add ext2add
+    # Compute the denominator for segment 0 FirstRow
+    padw mem_loadw.4294903304 drop drop push.1 push.0 ext2sub
```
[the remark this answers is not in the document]
hackaugusto: changed

bobbinth (approved):
Looks good! Thank you!

PR #328, opened 2023-06-29
feat: reintroduce match constraint syntax

bitwalker:
This PR brings back the syntax for match constraints, modified to remove grammar ambiguities discovered with the previous syntax.

An example of the syntax is shown below:

```airscript
integrity_constraints:
    enf match:
        case s[0] & s[1]: clk' = clk
        case !s[0] & !s[1]: clk' = 1
```

The choice of `enf match` vs `match enf` is intentional, and is intended to make the declaration of a constraint uniform, i.e. a constraint statement always begins with `enf` whether it is a simple constraint, a constraint comprehension, or a match constraint.

The implementation of this syntax simply compiles to a flat set of constraints for now, as if every branch of the match was a separate constraint conditioned on the selector for the match. This is handled directly in the parser for now, since we don't actually do anything meaningful with match constraints yet. Once we do, we'll update the parser accordingly and introduce distinct AST nodes to represent the construct.

---

I think we should still plan on eventually switching to block syntax for all "blocks" in the language, which was also discussed alongside the match syntax changes, but we were able to implement the latter without the former, so it is strictly speaking not necessary, but would make the grammar more regular (beneficial for any future syntax changes) and would improve the clarity of the language (IMO). In any case, that's a topic for a different discussion, but I wanted to make a note about it here for future reference.

### DIFF
--- a/air-script/tests/codegen/winterfell.rs
+++ b/air-script/tests/codegen/winterfell.rs
@@ -153,7 +153,6 @@ fn list_folding() {
 }
 
 #[test]
-#[ignore]
 fn selectors() {
     let generated_air = Test::new("tests/selectors/selectors.air".to_string())
         .transpile(Target::Winterfell)
--- a/air-script/tests/selectors/selectors.air
+++ b/air-script/tests/selectors/selectors.air
@@ -11,6 +11,6 @@ boundary_constraints:
 
 integrity_constraints:
     enf clk' = 0 when s[0] & !s[1]
-    match enf:
-        clk' = clk when s[0] & s[1] & s[2]
-        clk' = 1 when !s[1] & !s[2]
+    enf match:
+        case s[0] & s[1] & s[2]: clk' = clk
+        case !s[1] & !s[2]: clk' = 1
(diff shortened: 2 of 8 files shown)


bitwalker:
Looks like there is an issue with `miden-vm` causing the build to fail

bobbinth:
Yep - will fix later today.

bobbinth (approved):
Looks great! Thank you!

> The choice of `enf match` vs `match enf` is intentional, and is intended to make the declaration of a constraint uniform, i.e. a constraint statement always begins with `enf` whether it is a simple constraint, a constraint comprehension, or a match constraint.

Agreed! I like this approach.

> The implementation of this syntax simply compiles to a flat set of constraints for now, as if every branch of the match was a separate constraint conditioned on the selector for the match. This is handled directly in the parser for now, since we don't actually do anything meaningful with match constraints yet. Once we do, we'll update the parser accordingly and introduce distinct AST nodes to represent the construct.

Yes, agreed. I think with this PR we can probably close #125 and create a separate issue to address improvements I mentioned in https://github.com/0xPolygonMiden/air-script/issues/125#issuecomment-1522759268.

> I think we should still plan on eventually switching to block syntax for all "blocks" in the language, which was also discussed alongside the match syntax changes, but we were able to implement the latter without the former, so it is strictly speaking not necessary, but would make the grammar more regular (beneficial for any future syntax changes) and would improve the clarity of the language (IMO). In any case, that's a topic for a different discussion, but I wanted to make a note about it here for future reference.

I've been thinking about this as well, but as you said, this would be a separate discussion. Could you create an issue for this?

PR #336, opened 2023-07-04
add constraint evaluation export

hackaugusto:
~Review/Merge after https://github.com/0xPolygonMiden/air-script/pull/326~

This adds an export that performs the evaluation of both transition and boundary constraints, to be called by the stark verifier in the Miden's stdlib.

### DIFF
--- a/codegen/masm/src/codegen.rs
+++ b/codegen/masm/src/codegen.rs
@@ -827,6 +827,29 @@ impl<'ast> Backend<'ast> {
         }
     }
 
+    /// Emits code for the procedure `evaluate_constraints`.
+    ///
+    /// This will compute and cache values, the transition and boundary constraints for both the main and auxiliary traces.
+    fn gen_evaluate_constraints(&mut self) {
+        self.writer
+            .header("Procedure to evaluate the integrity and boundary constraints.");
+        self.writer.header("");
+        self.writer.header("Input: [...]");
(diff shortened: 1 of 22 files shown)


codegen/masm/src/codegen.rs
```diff
@@ -827,6 +827,29 @@ impl<'ast> Backend<'ast> {
         }
     }
 
+    /// Emits code for the procedure `evaluate_constraints`.
+    ///
+    /// This will compute and cache values, the transition and boundary constraints for both the main and auxiliary traces.
+    fn gen_evaluate_constraints(&mut self) {
+        self.writer
+            .header("Procedure to evaluate the transition and boundary constraints.");
```
bobbinth, line 835: nit: "transition and boundary constraints" should be "integrity and boundary constraints".

bobbinth (approved):
Looks good! Thank you! I left one small comment inline. Once it is addressed, we can merge.

Also, just for reference, this procedure would need to be integrated into the verifier [here](https://github.com/0xPolygonMiden/miden-vm/blob/main/stdlib/asm/crypto/stark/verifier.masm#L136).

codegen/masm/src/codegen.rs
hackaugusto: hahah, seems like calling it transition is hard wired in my brain. Thanks for pointing it out. Fixed

PR #337, opened 2023-07-06
Tidy up masm codegen

hackaugusto:
Best reviewed/merged after https://github.com/0xPolygonMiden/air-script/pull/336

This removes a bit of code duplication, and split the code a bit into separate modules to make `codegen.rs` smaller.

### DIFF
--- a/air-script/tests/aux_trace/aux_trace.masm
+++ b/air-script/tests/aux_trace/aux_trace.masm
@@ -13,9 +13,9 @@ proc.cache_z_exp
     dup.0 neq.0
     while.true
         movdn.2 dup.1 dup.1 ext2mul
-        # => [(z_1, z_0)^n, i, ...]
+        # => [(e_1, e_0)^n, i, ...]
         movup.2 add.1 dup.0 neq.0
-        # => [b, i+1, (z_1, z_0)^n, ...]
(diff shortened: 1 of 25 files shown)


bobbinth (approved):
Thank you! This looks good - but I think it would probably be a good idea to refactor the code a bit more (in a separate PR) to reduce the size of `codegen.rs` file even more and to separate things a bit better.

Specifically, I see two high-level parts in this:

1. Code which traverses the algebraic graph and generates procedures like `compute_integrity_constraints`, `compute_boundary_constraints_main_first` etc. This is the code which uses the visitor pattern as is intended.
2. All other code which generates kind of "boilerplate" procedures. This code doesn't really need to be a part of the visitor pattern.

The latter part can probably broken up into several components - e.g., component for dealing with periodic columns, component dealing with public inputs, component dealing with computing divisors etc. They would still share the same `Writer` - but could probably be encapsulated better.

To given an example: I think the code which deals with periodic columns could probably be moved into a separate module. This module could have a public function which:

- Takes a in iterator over periodic column definitions from the IR and the writer (and maybe some other relevant parameters).
- Emits all procedures relevant to periodic column evaluation into the writer.
- Returns `periods` vector, which I believe is needed later on.

PR #338, opened 2023-07-09
Update documentation for v0.3 release

bobbinth:
This PR updates documentation (primarily mdbook docs) for v0.3 release. Main concepts to be described/updated:

- [x] Modules.
- [x] Evaluators.
- [x] Constraint comprehension.
- [x] Selectors.

Left for future PR: update docs for MASM code generation option.


tohrnii (approved):
Thank you. Looks good to me. I left a few minor nits.

docs/src/description/organization.md
```diff
@@ -0,0 +1,77 @@
+# Code organization
+
+AirScript project can consist of one or more modules, each module located in a separate file with a `.air` extension. For projects consisting of multiple modules, one module must be declared as the root module, and all other modules must be declared as supporting modules.
+
+Currently, all modules must be located in a single directory, but in the future this limitation will be removed.
+
+All modules must start with a module name declaration followed by a set of source sections. These sections describe both the metadata and constraint evaluation logic for the AIR. Depending on the module type, some source sections may be required, be optional, a may not be allowed. The table below summarizes this information.
```
bitwalker, line 7: The part about source sections only applies to root modules really, as library modules only require the module declaration to be valid, and can contain any number of constant, function, or evaluator declarations, as well as imports. Might be worth distinguishing that here.

docs/src/description/organization.md
```diff
@@ -0,0 +1,77 @@
...
+
+| Section                                                                               | Root module | Supporting module |
+| ------------------------------------------------------------------------------------- | :---------: | :---------------: |
+| [constants](./declarations.md#constants-const)                                        | optional    | optional          |
+| [tace columns](./declarations.md#execution-trace-trace_columns)                       | required    | not allowed       |
```
bitwalker, line 12: NIT: you meant `trace` here

docs/src/description/organization.md
```diff
@@ -0,0 +1,77 @@
...
+| [public inputs](./declarations.md#public-inputs-public_inputs)                        | required    | not allowed       |
+| [periodic columns](./declarations.md#periodic-columns-periodic_columns)               | optional    | optional          |
+| [random values](./declarations.md#random-values-random_values)                        | optional    | not allowed       |
+| [boundary constraints](./constraints.md#boundary-constraints-boundary_constraints)    | required    | not allowed       |
+| [integrity constraints](./constraints.md#integrity-constraints-integrity_constraints) | required    | not allowed       |
+| [evaluators](./evaluators.md)                                                         | optional    | required          |
```
bitwalker, line 18: Evaluators aren't required in library modules, just optional, for example, you could have a module that just defines constants used across all modules

docs/src/description/organization.md
```diff
@@ -0,0 +1,77 @@
...
+
+Note that constants and evaluators are not really distinct sections but rather a set of declarations which can be done in-between any other sections.
+
+### Root module
+A root module defines an entrypoint into an AirScript project. It must start with a name declaration which consists of a `def` keyword followed by the name of the AIR project. For example:
+```
+def ExampleAir
```
bitwalker, line 25: NOTE: Module names must match the file name they are defined in, though without fickle checking the code, I do think I made the search case-insensitive. Just something to point out anyway. I'd also recommend using the `snake_case` naming convention for that reason.

docs/src/description/organization.md
```diff
@@ -0,0 +1,77 @@
...
+```
+mod example_module
+```
+where the name of the module must:
+- Be the same as the name of the file in which the supporting module is defined (e.g., the above module must be located in `example_module.air` file).
+- Be a string consisting of alpha-numeric characters and underscores.
+- Start with a letter.
+- End with a newline.
+
+Besides the name declaration, supporting modules must include at least one *evaluator*. Evaluators defined in a supporting module may be imported by a root or other supporting modules.
```
bitwalker, line 54: This isn't true currently, and not necessary to implement anyway, so we probably want to remove this.

docs/src/description/organization.md
```diff
@@ -0,0 +1,77 @@
...
+
+Supporting modules inherit random value declarations of the root module. That is, evaluators defined in a supporting module can reference random values declared in the root module.
+
+Additionally, supporting module may contain local definitions of constants and periodic columns.
+
+## Importing evaluators
+A module can import evaluators from supporting modules via a `use` statement. For example:
```
bitwalker, line 61: Constants can be imported as well

bitwalker (commented):
LGTM, just a few notes, but nothing critical

docs/src/description/convenience-syntax.md
```diff
@@ -52,4 +52,133 @@ integrity_constraints:
...
+trace_columns:
+    main: [a, b, c, s]
+
+integrity_constraints:
+    enf s^2 = s
+    enf match:
+        case s: c' = a + b
+        case !s: c' = a * c
+```
+In the above, the syntax of each "option" is `case <selector expression>: <constraint>`, where selector expression consists of values composed using binary operands `!`, `&`, and `|`. AirScript reduces these operands to their equivalent algebraic expressions as follows:
```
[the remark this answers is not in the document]
bobbinth: Yes - I like it better this way. Fixed!

docs/src/description/convenience-syntax.md
```diff
@@ -52,4 +52,133 @@ integrity_constraints:
...
+```
+trace_columns:
+    main: [a[5], b[5]]
+
+integrity_constraints:
+    enf x' = i * y for (x, y, i) in (a, b, 0..5)
+```
+The above will enforce that $a_i' = i \cdot b_i$ for $i \in [0, 5)$. If the length of either `a` or `b` is not 5, this will throw an error.
+
+## Constraint selectors
```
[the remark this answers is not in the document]
bobbinth: I like "Conditional constraints" better. There is a bit of domain-specific naming here as "selectors" is the name usually used for columns which turn some constraints on/off. So, I think we'll still use "selectors" for the expressions which follow the `case` keyword - but we'll use "Conditional constraints" for the whole `enf match` statement.

docs/src/description/organization.md
bobbinth: This is true only for library modules, right? I was able to compile a root module named `ExampleAir` from `example.air` file without issues.

bitwalker: That's a good point, it applies to import resolution/loading, so is really only relevant for library modules. That said, having module names that don't match the file name (or some other convention) makes it difficult to find modules programmatically, which might be a nice property to preserve. Definitely not essential as long as the typical workflow is compiling by passing a file path to the root module (or a string containing a root module definition).

PR #341, opened 2023-07-21
fix: incorrect propagation of constraint flag during inlining

bitwalker:
This fixes an issue which caused expansion of a folded comprehension within the context of a constraint to be improperly treated as a constraint comprehension. This expansion occurs during the inlining phase.

This bug occurred because a flag determines whether a given comprehension is a constraint comprehension or a regular list
comprehension, and is set to true when expanding a constraint. However, while expanding a constraint, we may encounter a call to a function, such as one of the list folding builtins, in which case a comprehension passed as an argument to those builtins is not expressing a constraint, so the flag should be false while expanding it.

We already manage this flag correctly in other contexts, but this one was missed during development.

### DIFF
--- a/parser/src/parser/tests/inlining.rs
+++ b/parser/src/parser/tests/inlining.rs
@@ -1113,3 +1113,256 @@ fn test_inlining_constraints_with_folded_comprehensions_in_evaluator() {
 
     assert_eq!(program, expected);
 }
+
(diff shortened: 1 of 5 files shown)


bobbinth (approved):
Looks great! Thank you!

PR #344, opened 2023-07-25
[WIP] feat: add support for functions

tohrnii:
This PR adds support for pure functions to AirScript.

TODOs:
- [ ] Using function calls as iterables in list comprehension.
- [ ] Add inlining tests for functions
- [ ] Using non-symbol arguments in function calls.
    - [ ] Binary expressions (`func(a + b)`)
    - [ ] Non scalar expressions (`func([1, 2, 3, 4])`)
    - [ ] Function calls (`func_a(func_b(a))`)

### DIFF
--- a/parser/src/transforms/inlining.rs
+++ b/parser/src/transforms/inlining.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{BTreeMap, HashMap, HashSet, VecDeque},
     ops::ControlFlow,
+    vec,
 };
 
 use air_pass::Pass;
@@ -19,12 +20,12 @@ use super::constant_propagation;
 /// * Monomorphizing and inlining evaluators/functions at their call sites
 /// * Unrolling constraint comprehensions into a sequence of scalar constraints
 /// * Unrolling list comprehensions into a tree of `let` statements which end in
-/// a vector expression (the implicit result of the tree). Each iteration of the
-/// unrolled comprehension is reified as a value and bound to a variable so that
-/// other transformations may refer to it directly.
+///   a vector expression (the implicit result of the tree). Each iteration of the
+///   unrolled comprehension is reified as a value and bound to a variable so that
+///   other transformations may refer to it directly.
 /// * Rewriting aliases of top-level declarations to refer to those declarations directly
 /// * Removing let-bound variables which are unused, which is also used to clean up
-/// after the aliasing rewrite mentioned above.
+///   after the aliasing rewrite mentioned above.
 ///
 /// The trickiest transformation comes with inlining the body of evaluators at their
 /// call sites, as evaluator parameter lists can arbitrarily destructure/regroup columns
@@ -75,6 +76,8 @@ pub struct Inlining<'a> {
     imported: HashMap<QualifiedIdentifier, BindingType>,
     /// All evaluator functions in the program
     evaluators: HashMap<QualifiedIdentifier, EvaluatorFunction>,
+    /// All pure functions in the program
+    functions: HashMap<QualifiedIdentifier, Function>,
     /// A set of identifiers for which accesses should be rewritten.
     ///
     /// When an identifier is in this set, it means it is a local alias for a trace column,
@@ -82,7 +85,8 @@ pub struct Inlining<'a> {
     /// identifier in `bindings`.
     rewrites: HashSet<Identifier>,
     in_comprehension_constraint: bool,
-    next_ident: usize,
+    next_ident_lc: usize,
+    next_ident_call: usize,
 }
 impl<'p> Pass for Inlining<'p> {
     type Input<'a> = Program;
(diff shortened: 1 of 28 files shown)


parser/src/transforms/inlining.rs
```diff
@@ -293,9 +302,14 @@ impl<'a> Inlining<'a> {
             }
             // Expression statements are introduced during inlining, and are always already expanded,
             // but they are recursively visited to apply rewrites
-            Statement::Expr(mut expr) => {
-                self.rewrite_expr(&mut expr)?;
-                Ok(vec![Statement::Expr(expr)])
+            Statement::Expr(mut expr) => match expr {
+                Expr::Call(call) => {
```
bitwalker, line 306: I would revert this specific change here - when we expand a statement block, we will expand call sites, and in the process recursively expand the callee. We do want to rewrite call expressions which are introduced into the tree when inlining the callee (so that arguments are rewritten to use the expressions from the caller). I think expanding the call here would cause inlining of the callee to occur without rewriting the arguments, and result in incorrect rewriting of the callee body.

I'm working through this by running the code in my mental interpreter, so to speak, so I may be forgetting something that makes this necessary - but the comment I left just above this is what reminded me that this is already supposed to be an expression introduced as the result of inlining, and we only are interested in applying rewrites.

It may also be very possible that `Expr::Call` can never be observed here - a good way to check that would be to make the match arm panic and see if you can trigger it. Without working through some tests designed to see what exactly is going on when a call is seen here, it's hard for me to say for sure whether that would be unexpected with certainty, but my gut feeling is that we don't want to attempt expanding calls here.

tohrnii: Question: Would we not need it when the return value is a function call of another function?
```
fn foo(a: felt) -> felt:
    return bar(a)
```

Maybe there's another way to handle this?

parser/src/transforms/inlining.rs
```diff
@@ -431,7 +447,50 @@ impl<'a> Inlining<'a> {
...
+            ScalarExpr::Binary(BinaryExpr {
+                op,
+                ref mut lhs,
+                ref mut rhs,
+                span,
+            }) => {
+                let mut stmts = Vec::new();
+
+                // Recursively expand the lhs and rhs.
+                stmts.extend(self.expand_scalar_expr(lhs)?);
```
bitwalker, line 481: This is an example of a place where we need to be careful about how we handle the results of `expand_scalar_expr`. In particular, we need to handle this similarly to what I described in `expand_binary_expr`. Expand both operands, and then do the following:

Check the expansion of the left-hand operand to see if it terminates with a `let`. If it does, we're going to use `with_let_result` on it to perform a transformation like we did in `expand_binary_expr`. In the call to `with_let_result`, we will  check if the right-hand expansion also terminates with a let, and if so, we're going to immediately call `with_let_result` on it, and emit the binary operator in the body of a new `let`. The, now rewritten, expansion of the right-hand operand will be used as the body of a new `let` in the expansion of the left-hand statement block.

At the end, we simply return the expanded left-hand statement block (as we've moved the right-hand statement block into it).

NOTE: You won't need to call `rewrite_scalar_expr`, as by expanding both operands, that function gets called already.

NOTE: If either operand expands to a block that doesn't terminate in a `let`, which should only happen if there is just a single `Statement::Expr` in the block, you can emit the `with_let_result` for that operand, and just use that expression as the operand for the binary operator directly. Obviously, in the case where both operands are simple expressions, you can omit generating a `let` entirely, just as in `expand_binary_expr`

parser/src/transforms/inlining.rs
```diff
@@ -800,10 +862,13 @@ impl<'a> Inlining<'a> {
                     self.bindings.insert(binding, binding_ty);
                     Expr::SymbolAccess(current_access)
                 }
-                // TODO: Currently, calls cannot be used as iterables, because we don't have pure functions
-                // which can produce aggregates. However, when those are added, we may want to add support
-                // for that here. This branch is set up to raise an appropriate panic if we forget to do so.
-                Expr::Call(_) => unimplemented!("calls to functions as iterables"),
+                // Expr::Call(call) => {
```
bitwalker, line 865: We probably should perform a transformation earlier in the compiler pipeline that rewrites comprehension expressions with calls as iterable values, into a `let` tree, where the result of each call is bound to a new variable via `let`, and at the bottom of the let tree we place the comprehension expression, with the calls rewritten to their corresponding `let`-bound variable.

Alternatively, we could disallow calls in iterable position at the syntax level, but since we don't have any specific reason why we can't support them, that might not be desirable, aside from a little less complexity.

In either case, we would then handle `Expr::Call` here just like `Expr::Binary` and `Expr::ListComprehension` below.

parser/src/transforms/inlining.rs
bitwalker: Expansion is performed recursively, so when we expand the call to `foo`, we will also expand it's body, and thus the call to `bar` as well.

As an aside, we need to make sure we check for recursive functions calls and raise a diagnostic. We can do that during inlining by tracking what calls we're currently expanding. Each time we start expanding a call, we check if we're currently in an expansion of the callee, and if so, raise a diagnostic; if not, then we insert the callee into the visited set and proceed with inlining. When we finish expanding a call, we remove the callee from the set.

We should probably also do that for evaluators, since the same issue can occur there, I don't think I handled that anywhere yet.

We can handle that later though, no need to do that as part of this PR.

bitwalker:
Superceded by #351

PR #350, opened 2024-04-10
MassaLabs : Use a block syntax for declarations and statements blocks

AurelienFT:
Implements the RFC : https://github.com/0xPolygonMiden/air-script/issues/331


bitwalker (approved):
Looks good! Definitely a pain to have to go through and update all the existing AIR syntax and tests 😅, the effort is much appreciated!

@bobbinth I'll let you merge this at your discretion, but 👍 from me

bobbinth (approved):
Looks great! Thank you!

A question (for a potential different PR): since we are adopting a more C-like syntax, should we also use semicolons for statement terminators?

PR #351, opened 2024-06-21
feat: implement support for functions

bitwalker:
This PR introduces a new type of function to the AirScript language: "pure" functions. To summarize:

* Pure functions take zero or more parameters, and return a result of scalar or aggregate type.
* Pure functions must declare the type of all parameters and the result. These types are checked to ensure that uses of the parameters are consistent, and that all callers are passing values of the correct type.
* Pure functions **MUST** return a value, using the `return` statement
* Correspondingly, they **MAY NOT** have an empty body, and at minimum must contain a `return` statement
* `return` is the keyword for a new statement of the form `return EXPR`, where `EXPR` is the expression which evaluates to the value to be used as the output of the function. A bare `return` is not valid syntax.
* Pure functions **MAY NOT** contain constraints. This is reserved for evaluator functions, which have slightly different syntax and semantics.
* Like evaluator functions, pure functions may be placed in library modules and imported for use in any program linked to that library.

## Example

```
fn madd3(a: felt[3], b: felt) -> felt {
    let d = [c * b for c in a]
    return sum(d)
}
```

The example above demonstrates the key features of the new syntax. A few notes:

* Valid types are `felt`, `felt[N]`, or `felt[N][M]` where `N` and `M` are placeholders for the corresponding dimension of the given aggregate type (vectors and matrices, respectively)
* Function bodies can contain all the things you'd expect to find in `integrity_constraints` or `boundary_constraints`, except the constraints themselves as mentioned above (i.e. the `enf` keyword is not allowed in this context at all)

This PR supercedes #344 

---

Internally, the support for functions also comes with some refactoring of the AST, so as to allow us to represent arbitrarily complex expansions of syntax nodes in expression position. Previously, the AST was designed to forbid representing syntax that was not supported by the surface language. However, due to the requirements of the inlining phase, it has been necessary to relax this to some degree, and this PR takes that even further, now allowing the use of the `Let` syntax node in both `Expr` and `ScalarExpr` as an explicit variant.

Despite this additional flexibility though, it is still the case that the surface language does not have syntax for this. Furthermore, we explicitly do not support the use of `Let` in expression position until after semantic analysis; and while supported during constant propagation, we would not expect to see `Let` in this position until after the inlining pass has been applied.

The point being: do not look at the changes to the AST as representing changes to the language syntax. What is supported by the language is determined by the language grammar/parser, and the AST itself is considerably less restrictive so as to allow for arbitrary transformations during compilation.

@bobbinth I've added some tests, as well as built on top of those initially implemented by @tohrnii. I'm not certain how best to evaluate the codegen side of things, so I'll defer to you on that. However, the codegen tests written by @tohrnii do pass, so I'm assuming those are good.

### DIFF
--- a/parser/src/parser/grammar.lalrpop
+++ b/parser/src/parser/grammar.lalrpop
@@ -75,6 +75,7 @@ Declaration: Declaration = {
     PeriodicColumns => Declaration::PeriodicColumns(<>),
     RandomValues => Declaration::RandomValues(<>),
     EvaluatorFunction => Declaration::EvaluatorFunction(<>),
+    Function => Declaration::Function(<>),
     <l:@L> <trace:Trace> <r:@R> => Declaration::Trace(Span::new(span!(l, r), trace)),
     <PublicInputs> => Declaration::PublicInputs(<>),
     <BoundaryConstraints> => Declaration::BoundaryConstraints(<>),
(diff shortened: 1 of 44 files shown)


parser/src/parser/grammar.lalrpop
```diff
@@ -256,6 +257,42 @@ EvaluatorSegmentBindings: (SourceSpan, Vec<Span<(Identifier, usize)>>) = {
...
+
+FunctionBindingType: Type = {
+    <l:@L> "felt" <r:@R> => Type::Felt,
+    <l:@L> "felt" <size: Size> <r:@R> => Type::Vector(size as usize),
+    <l:@L> "felt" "[" <row_size: Num_u64> "," <col_size: Num_u64> "]" <r:@R> => Type::Matrix(row_size as usize, col_size as usize),
+}
+
+FunctionBody: Vec<Statement> = {
+    // TODO: validate
+    <l:@L> <stmts: StatementBlock> <r:@R> =>? {
```
bobbinth, line 284: Question: what does this `TODO` mean here?

bobbinth (approved):
Looks great! Thank you! I did a relatively shallow review of the parser, but I did verify that  Winterfell codegen works correctly (MASM codegen should be OK too as we have a test which compares identical constraints written with and without functions).

parser/src/parser/grammar.lalrpop
bitwalker: I think it's actually already done, I just forgot to remove the TODO, I'll do that and then merge this.

PR #353, opened 2024-08-07
MassaLabs: Add semicolon in AirScript

sydhds:


parser/src/parser/grammar.lalrpop
```diff
@@ -290,7 +290,7 @@ StatementBlock: Vec<Statement> = {
 }
 
 Let: Let = {
-    <l:@L> "let" <name: Identifier> "=" <value: Expr> <r:@R> <body: StatementBlock>
+    <l:@L> "let" <name: Identifier> "=" <value: Expr> ";" <r:@R> <body: StatementBlock>
```
bitwalker, line 293: Once we have a proper IR to lower the AST into, we should be able to simplify the grammar by flattening it, i.e. we won't need to nest `StatementBlock`s inside `Let`.

For now, it's necessary because we're using the AST both as a syntax tree and as an intermediate representation, so keeping the language structure expression-oriented simplifies aspects of that. One upside of having that limitation, if I had to pick one, is that it forces us to avoid syntax that would require execution semantics we can't or don't want to support, but that's something we can still accomplish by sticking to functional-immutable semantics.

Anyway, nothing actionable here, just wanted to provide some context on this, since I'm sure it seems weird.

parser/src/parser/tests/boundary_constraints.rs
```diff
@@ -11,11 +11,11 @@ const BASE_MODULE: &str = r#"
 def test
 
 trace_columns {
-    main: [clk]
+    main: [clk];
```
bitwalker, line 14: @bobbinth Perhaps it would make more sense to use `,` with these top-level definitions? Semicolons feels a bit bizarre in what amounts to a data structure definition.

In statement blocks and stuff, it feels pretty natural, but less so here.

bitwalker (changes requested):
Overall, looks great! If you could fix the clippy errors so the build passes, that's the only real blocker.

I would like to get @bobbinth's input on whether semicolons are the right choice for internal declarations of `trace_columns` and the like, but even if we're not convinced that it's the best choice of syntax, we can always address it later, so we don't need to hold up merging this on that aspect.

parser/src/parser/tests/boundary_constraints.rs
bobbinth: Agreed - in declarations like these, commas would be a more natural choice.

air-script/tests/bitwise/bitwise.air
```diff
@@ -1,58 +1,58 @@
...
-    main: [s, a, b, a0, a1, a2, a3, b0, b1, b2, b3, zp, z, dummy]
+    main: [s, a, b, a0, a1, a2, a3, b0, b1, b2, b3, zp, z, dummy];
 }
 
 periodic_columns {
-    k0: [1, 0, 0, 0, 0, 0, 0, 0]
-    k1: [1, 1, 1, 1, 1, 1, 1, 0]
+    k0: [1, 0, 0, 0, 0, 0, 0, 0];
+    k1: [1, 1, 1, 1, 1, 1, 1, 0];
 }
```
bobbinth, line 14: Following @bitwalker 's comment, I think in these sections, we should use commas - e.g., they would look like so:
```
public_inputs {
    stack_inputs: [16],
}

trace_columns {
    main: [s, a, b, a0, a1, a2, a3, b0, b1, b2, b3, zp, z, dummy],
}

periodic_columns {
    k0: [1, 0, 0, 0, 0, 0, 0, 0],
    k1: [1, 1, 1, 1, 1, 1, 1, 0],
}
```

bobbinth (commented):
Looks great! Thank you! I don't have much additional comments besides @bitwalker 's suggestion to use commas in the "declarative" sections. Once this is addressed, we can merge.

bitwalker:
@sydhds I've fixed the clippy warnings in `next`, so once you rebase your changes, the build should be green again. There may be some conflicts due to the function implementation, but shouldn't be anything major, mostly just the fact that we likely touched the same files and tests, but I think the actual changes are pretty disjoint.

air-script/tests/bitwise/bitwise.air
sydhds: Fixed :)

sydhds:
> @sydhds I've fixed the clippy warnings in `next`, so once you rebase your changes, the build should be green again. There may be some conflicts due to the function implementation, but shouldn't be anything major, mostly just the fact that we likely touched the same files and tests, but I think the actual changes are pretty disjoint.

Rebase done so I think it should be all good :)

bitwalker (approved):
LGTM! Thanks @sydhds!

air-script/tests/selectors/selectors_with_evaluators.air
```diff
@@ -1,33 +1,33 @@
...
 }
 
 integrity_constraints {
-    enf next_is_zero([clk]) when s[0] & !s[1]
+    enf next_is_zero([clk]) when s[0] & !s[1];
     enf match {
         case s[1] & s[2]: is_unchanged([clk, s[0]])
         case !s[1] & !s[2]: next_is_one([clk])
-    }
+    };
```
bobbinth, line 32: Should we add commas to the end of each `case` clause? For example, something like this:
```rust
enf match {
    case s[1] & s[2]: is_unchanged([clk, s[0]]),
    case !s[1] & !s[2]: next_is_one([clk]),
};
```

ir/src/tests/selectors.rs
```diff
@@ -155,35 +155,35 @@ fn selectors_inside_match() {
...
     }
 
     integrity_constraints {
-        enf next_is_zero([clk]) when s[0] & !s[1]
+        enf next_is_zero([clk]) when s[0] & !s[1];
         enf match {
             case s[1] & s[2]: is_unchanged([clk, s[0]])
             case !s[1] & !s[2]: next_is_one([clk])
-        }
+        };
```
bobbinth, line 186: Similar comment as the one above here.

bobbinth (approved):
Looks good! Thank you! I left a couple of comments inline about adding commas to separate clauses of the `match` statement. But if that's difficult to do - we can do that in a separate PR.

air-script/tests/selectors/selectors_with_evaluators.air
sydhds: I was planning to add that in another PR but let me know if you want in this one?

bobbinth: A follow-up PR is fine! I'll merge this one now then.

PR #356, opened 2024-08-19
MassaLabs: add comma in match/case statement

sydhds:
As discussed in #353, this PR modify the grammar to require a comma at the end of case statements inside a match statement.


bobbinth (approved):
Looks good! Thank you!

bitwalker (approved):

PR #359, opened 2024-10-17
MassaLabs: Implement a Intermediate Representation to improve the compilation process

Leo-Besancon:
The goal of this PR is to introduce a Middle Intermediate Representation, to avoid making optimization on the AST directly, while keeping enough information to handle type checking, optimization for each pass.

See the initial issue and design discussion for additional context:
- https://github.com/0xPolygonMiden/air-script/issues/357
- https://github.com/0xPolygonMiden/air-script/discussions/358

Putting it as draft now that we have:
- Basic MIR graph structure. We basically started from the AIR and added the operations we needed to represent the same expressions as the AST.
- Lowering from the AST. The tests that check whether various AirScript programs compile with the pipeline `Parse to AST > Constant propagation > Lowering to MIR` all pass, but the resulting MIR is not checked. We may need to improve testing on this side.
- Start of the inlining pass from MIR to MIR. For now, we focused on only inlining the function bodies at their callsite, but we do not unroll vectors for instance. We feel it may be better to make this in a separate unrolling pass if possible.

Additionally, we have a partial pretty printer for the MIR to help us debug in the future (to ensure the graph constructed is what we expect after each pass)

@bitwalker, don't hesitate to comment on things that should be handled differently.
For now, we haven't fully settled on the various nodes of the MIR graph, as we add / change things depending on the needs of our implementation. We will also add checks made to ensure the proper diagnostics are raised (potentially after each pass as discussed previously), but we will probably do this at a later stage.


ir/src/passes/constant_propagation.rs
```diff
@@ -0,0 +1,36 @@
+use std::ops::ControlFlow;
+
+use air_pass::Pass;
+use miden_diagnostics::DiagnosticsHandler;
+
+use crate::MirGraph;
+
+pub struct ConstantPropagation<'a> {
```
bitwalker, line 8: This pass becomes trivial with the design I outlined above, after one has implemented the generic pre- and post-order visitors for `Operation`, basically:

1. Perform a post-order visit of the `Function`, pushing all of the operations which have at least one use of a constant op result on a stack. This could also be an opportunity to unique constants, and separately enqueue constant ops that should be erased, while also recording a mapping from the values produced by those ops to the canonical value for that constant.
2. Process the next item on the work stack, and determine if that operation can be constant folded based on its constant operands. If so, fold the operation. If folded in place, add all users of the op's results to the work stack, and go to the next op on the stack. If the op can be folded away completely and replaced with a constant value, go to step 3. If the operation could not be folded, start processing the next item.
3. Materialize a new constant op representing the folded constant, and insert it before the op we want to replace.
4. Replace the results of the original op with those of the constant op (updating all the users at the same time). Add each unique using operation to the work stack, if not already present on the stack.
5. Erase the original op. Go to step 2.

Because the data flow graph is implicit in the edges between operations/operands/results, doing the above is quite efficient, since we only need to do a single traversal of the graph, and we only revisit operations when one of their uses has been constant folded (so as to see if the change makes it foldable).

ir/src/passes/inlining.rs
```diff
@@ -0,0 +1,211 @@
...
+//use miden_diagnostics::DiagnosticsHandler;
+
+use crate::{MirGraph, NodeIndex, Operation};
+
+//pub struct Inlining<'a> {
+//     #[allow(unused)]
+//     diagnostics: &'a DiagnosticsHandler,
+//}
+
+pub struct Inlining {}
```
bitwalker, line 13: For this pass, you can again use a post-order visit of the graph, inlining the regions of structured ops (e.g. comprehensions, functions) into the containing region, by replacing all uses of entry block arguments with the operands used at the call site.

The intrusive linked lists containing operations in a `Block` can be spliced/split and stolen relative to a cursor, so it is quite efficient to do the actual mechanical inlining.

It might not also be all that important to do inlining with the representation I outlined above, since I suspect it is relatively straightforward to lower to AIR without it, but some optimizations will be region or block-local, so that's the only tradeoff.

ir/src/passes/translate.rs
```diff
@@ -1,46 +1,80 @@
...
+        let mut builder = MirBuilder {
             diagnostics: self.diagnostics,
-            air: &mut air,
+            mir: &mut mir,
             random_values,
             trace_columns,
             bindings: Default::default(),
         };
 
+        // Insert placeholders nodes for future Operation::Definition (needed for function bodies to call other functions)
```
bitwalker, line 50: Could you elaborate a bit on what the placeholder op is and why it is needed? Again, i suspect the representation I outlined above makes this unnecessary, but would like to make sure I understand the problem you ran into.

bitwalker (commented):
Sorry it took me so long! I wanted to make sure I spent some time thinking about this and evaluating the design.

I've left some comments on some details that I would like clarification on, or that feel a bit awkward (or at least unusual), as well as one which goes into much more detail on an alternative representation for operations/values/etc. that I think might provide a much nicer substrate to build on. I would be interested to get your take on that, in terms of whether it would be easier for you to work with, or whether there are other constraints I am not recalling that make it less ideal.

In any case, I think what is implemented here can work, though there are a few things I'm uncertain about, as mentioned in some of my comments. I'm going through the inlining pass in more detail right now, but wanted to get my initial feedback to you first.

Feel free to ping me with any questions, or if you just want to bounce an idea of me, happy to discuss this further!

ir/src/passes/inlining.rs
```diff
@@ -0,0 +1,211 @@
...
+        if let Operation::Definition(_, _, body) = &def_node.op {
+            // Find all calls in the body
+            for (index_in_body, call_index) in body.iter().enumerate() {
+                inline_call(ir, call_index, &def_node_index, index_in_body);
+            }
+        }
+    }
+}
+
+fn inline_call(
```
bitwalker, line 63: It isn't clear to me whether this transformation is sound:

* `inline_all` is iterating over what amounts to a snapshot of the body of the outer definition (if I understand correctly).
* But here in `inline_call`, the actual body of the definition is being mutated/replaced (to insert new items at the index at which the call occurred in the outer def). But after splicing ops into the body of the outer def, all subsequent calls to `inline_call` are using stale indices.

Am I misunderstanding what's actually happening in `inline_all`/`inline_call`?

ir/src/passes/inlining.rs
```diff
@@ -0,0 +1,211 @@
...
+    // 1. Understand the basics of the previous inlining process
+    // 2. Remove what is done during lowering from AST to MIR (unroll, ...)
+    // 3. Check how it translates to the MIR structure
+    fn run_visitor(&mut self, ir: &mut MirGraph) -> ControlFlow<()> {
+        inline_all(ir);
+        ControlFlow::Continue(())
+    }
+}
+
+fn inline_all(ir: &mut MirGraph) {
```
bitwalker, line 49: One of the things I was hoping we could achieve with the new IR was avoiding unbounded recursion to visit the graph and perform analysis/transformation. Writing it in recursive style for the initial implementation is fine, but we should try and refactor it into iterative style so that the pass can handle arbitrarily deep graphs without panicking.

ir/src/tests/ir/inlining.rs
```diff
@@ -0,0 +1,125 @@
...
+    use crate::MirType;
+    use crate::MirValue;
+    use crate::Node;
+    use crate::NodeIndex;
+    use crate::Operation;
+    use crate::SpannedMirValue;
+    use air_pass::Pass;
+
+    #[test]
+    fn test_inlining() {
```
bitwalker, line 16: We should test a slightly more complex module, containing multiple calls in the same block, and calls to functions which themselves contain calls, preferably with at least one of the non-leaf functions containing a comprehension that calls a function in its body. This will catch most of the edge cases that come up when inlining non-trivial programs. It is particularly important to test the interaction of comprehensions containing function calls.

In particular, my question about soundness of how calls are inlined about probably be answered by whether or not such a test produces the expected expansion of the input program.

bitwalker:
> The tests that check whether various AirScript programs compile with the pipeline Parse to AST > Constant propagation > Lowering to MIR all pass, but the resulting MIR is not checked. We may need to improve testing on this side.

All of the transformation tests are operating on the AST using the old passes AFAICT, so those tests will need to be ported to the IR to validate that the IR upholds the expected behavior. Once the lowering to AIR is implemented, we can also evaluate how the resulting AIR looks. I'd say porting over the tests for inlining and some of the other transformations (once implemented in the IR), are the most essential, but will be a bit tedious, since those tests are against the AST, and the IR differs in various ways. Those tests largely aimed to ensure that:

* Everything was inlined (i.e. there were no un-inlined blocks remaining in the output)
* Anything that was constant-foldable, was folded (to the degree to which we implemented folding)
* All references to trace columns, random values, let-bound variables, function parameters, etc., were properly expanded. For example, passing a trace access to a function, which contained a comprehension where that function parameter was used as an iterable, would expand to the equivalent of hand-unrolling the comprehension at the top-level. There are a lot of fiddly details there, much of which is due to using the AST as an IR, but we'll still want to verify that the outcomes are the same.

> Start of the inlining pass from MIR to MIR. For now, we focused on only inlining the function bodies at their callsite, but we do not unroll vectors for instance. We feel it may be better to make this in a separate unrolling pass if possible.

That should be fine to split up into call inlining and loop unrolling, but you'll want to still probably want to visit the comprehension bodies during the function call inlining pass, to ensure that there are no calls remaining in the program to inline after it completes. Then, when unrolling comprehensions, no further work needs to be done. The tradeoff there is that you won't have the opportunity to do constant propagation across comprehension boundaries until the comprehensions are unrolled, which means you'll have more work to do when unrolling comprehensions. Probably the result would be that you need to run both inlining passes _first_, and only then do constant propagation and dead code elimination. 

With compilers, transformation ordering is a whole science/art, luckily we're working with a pretty restrictive ISA, but we'll need to be careful to ensure that transformations either don't make too many assumptions about when they are run, or that we run them in the exact order needed.

ir/src/ir/value.rs
```diff
@@ -1,15 +1,28 @@
...
 
-/// Represents a scalar value in the [AlgebraicGraph]
+/// Represents a scalar value in the [MIR]
 ///
 /// Values are either constant, or evaluated at runtime using the context
 /// provided to an AirScript program (i.e. random values, public inputs, etc.).
-#[derive(Debug, Eq, PartialEq, Copy, Clone)]
-pub enum Value {
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub enum MirValue {
```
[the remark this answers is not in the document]
Soulthym: Our current design for the IR is closer to a sea-of-nodes than the more common CFG representation. The main difference with a classic sea-of-nodes implementation is that we do preserve ordering in the function/block bodies. This also means that our SSA values can be extrapolated deterministically when pretty printing rather than being stored in the nodes directly. We currently do this by traversing block bodies in order, but could also be done with post-order traversal of the dependency graph.
We have a few questions/remarks about your proposed structure:
1) What would be the advantage of using an intrusive linked list here?
2) Switch from sea-of-nodes to CFG: might delay work on passes and need a rewrite of the lowering AST->MIR. The 2 models are valid for an IR, with varying advantages and drawbacks.
3) We agree that the node types should be more properly separated, we'll tend towards your proposed design on this aspect over time, incorporating elements as we need them for this implementation.
4) Keeping the regionalised owner attributes would help simplify passes greatly. We'll incorporate that into our design.
Overall we think we can benefit from the proposed changes without necessarily changing the core design of the sea-of-nodes system we already have. This also minimizes time spent on redoing existing work.

ir/src/ir/operation.rs
```diff
@@ -1,22 +1,73 @@
...
+    For(Vec<NodeIndex>, NodeIndex, Option<NodeIndex>),
+    /// If (condition, then, else)
+    If(NodeIndex, NodeIndex, NodeIndex),
+
+    /// A reference to a specific variable in a function
+    /// Variable(MirType, argument position)
+    Variable(SpannedVariable),
+    /// A function definition (Vec_params, optional return_variable, body)
+    /// Definition(Vec<Variable>, Variable, body)
+    Definition(Vec<NodeIndex>, Option<NodeIndex>, Vec<NodeIndex>),
```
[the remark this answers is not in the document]
Soulthym: That's a good point, will do.

ir/src/passes/inlining.rs
Soulthym: We agree with your analysis and are aware of the current implementation's drawbacks. The next version will use a post-order visit coupled with a work stack where we add newly generated Call nodes while inlining.

ir/src/ir/operation.rs
```diff
@@ -1,22 +1,73 @@
...
+#[derive(Debug, PartialEq, Eq, Clone)]
 pub enum Operation {
-    /// Evaluates to a [Value]
-    ///
-    /// This is always a leaf node in the graph.
-    Value(Value),
+    /// Begin primitive operations
+
+    /// Evaluates to a [TypedValue]
+    Value(SpannedMirValue),
```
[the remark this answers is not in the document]
Soulthym: Agreed on this, we go into more details on this reply https://github.com/0xPolygonMiden/air-script/pull/359#discussion_r1816898610

ir/src/ir/operation.rs
```diff
@@ -1,22 +1,73 @@
...
     /// Evaluates by multiplication over two operands (given as nodes in the graph)
     Mul(NodeIndex, NodeIndex),
+    /// Enforces a constraint (a given value equals Zero)
+    Enf(NodeIndex),
+
+    /// Begin structured operations
+    /// Call (func, arguments)
+    Call(NodeIndex, Vec<NodeIndex>),
+    /// Fold an Iterator according to a given FoldOperator and a given initial value
+    Fold(NodeIndex, FoldOperator, NodeIndex),
```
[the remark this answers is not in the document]
Soulthym: That's a good point and shouldn't be too hard to implement. We'll incorporate it into our design.

ir/src/passes/translate.rs
Soulthym: The PlaceHolder Operation variant is currently used to obtain the NodeIndex before the node is fully defined to help with incremental graph edits. Main caveat is that we might have to rework the uniquing strategy.
Should be possible to do without this node type with a slight rework of how we insert nodes.

ir/src/tests/ir/inlining.rs
Soulthym: Will add more complex tests to assert soundness of this pass.

ir/src/passes/inlining.rs
Soulthym: The naming might be causing confusion here, there is no unbounded recursion, just a while loop calling inline_all -> inline_call -> inline_body -> inline_op. I'll try and simplify this once the proposed Node design is implemented with back-references via Op.owner, and will rename the functions to differentiate more easily between inline_all and inline_call.

ir/src/graph/mod.rs
```diff
@@ -50,94 +54,114 @@ impl Node {
...
+            Operation::For(_iterator, _body, _selector) => todo!(),
+            Operation::If(_condition, _then, _else) => todo!(),
+            Operation::Variable(_var) => todo!(),
+            Operation::Definition(_params, _return, _body) => todo!(),
+            Operation::Vector(_vec) => todo!(),
+            Operation::Matrix(_vec) => todo!(),
+        }
+    }*/
+
     /// Insert the operation and return its node index. If an identical node already exists, return
```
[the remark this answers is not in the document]
Soulthym: Seems like a good idea to scope uniqueness per region, for now the unique property was not upheld during inlining but expected afterwards due to the way we use PlaceHolder nodes, which means we have to do a pass of CSE. We think we can do without those node types after scope regionalisation is implemented, it also seems to be the best option in our opinion.

ir/src/ir/operation.rs
```diff
@@ -1,22 +1,73 @@
...
+    /// For (iterators, body, selector)
+    For(Vec<NodeIndex>, NodeIndex, Option<NodeIndex>),
+    /// If (condition, then, else)
+    If(NodeIndex, NodeIndex, NodeIndex),
+
+    /// A reference to a specific variable in a function
+    /// Variable(MirType, argument position)
+    Variable(SpannedVariable),
```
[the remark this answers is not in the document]
Soulthym: Variable is improperly named here, it is meant to represent function parameters and not variables, will rename to Parameter to avoid further confusion

ir/src/passes/constant_propagation.rs
Soulthym: We agree.

ir/src/passes/inlining.rs
Soulthym: > The intrusive linked lists containing operations in a Block can be spliced/split and stolen relative to a cursor, so it is quite efficient to do the actual mechanical inlining.

Wouldn't reusing existing nodes by splicing them during inlining cause conflicts in case a function is called twice (or more) with different arguments?

The current implementation duplicates those nodes for this reason. The main problem in doing that is maintaining the unique property globally, which should be fixed once a regionalized uniquing strategy is implemented.

Soulthym:
> That should be fine to split up into call inlining and loop unrolling, but you'll want to still probably want to visit the comprehension bodies during the function call inlining pass, to ensure that there are no calls remaining in the program to inline after it completes.

> With compilers, transformation ordering is a whole science/art, luckily we're working with a pretty restrictive ISA, but we'll need to be careful to ensure that transformations either don't make too many assumptions about when they are run, or that we run them in the exact order needed.


Agreed with your analysis. Pass ordering can probably be reworked after the fact, we'll keep that in mind while implementing the various passes.

ir/src/graph/mod.rs
```diff
@@ -148,40 +172,290 @@ impl AlgebraicGraph {
...
+            .field("indent", &self.indent)
+            .field("nl", &self.nl)
+            .field("in_block", &self.in_block)
+            .field("var_count", &self.shared.borrow().var_count)
+            .field("fn_count", &self.shared.borrow().fn_count)
+            .finish()
+    }
+}
+
+pub fn pretty(graph: &MirGraph, roots: &[NodeIndex]) -> String {
```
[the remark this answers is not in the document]
Soulthym: Noted, we'll convert our implementation to use miden-formatting instead of our custom implementation once other points discussed here are resolved/implemented.

Soulthym:
@Bitwalker, first of all, thank you for this detailed answer, it will sure help a lot!
We have left several comments trying to address all the points you have raised. We hope that helps clarifying the design we proposed and how we plan to incorporate your feedback.

ir/src/ir/operation.rs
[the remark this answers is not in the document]
bitwalker: So I think my actual question is: does `Variable` (or `Parameter` if you rename it) represent a named binding, or an SSA value?

My concern is that named bindings remain in the IR in any form. In other words, we shouldn't need to concern ourselves with names and their associated lexical scoping rules - particulary in the presence of uniquing. I would expect the IR to be composed of, essentially, blocks (with optional parameter lists) and operations (with operands and results), where block parameters, operands, and results are all SSA values. The dataflow of the program is then implicit in the edges of the value dependency graph.

As an aside, it might be a bit of a misnomer to call them SSA values, as AirScript doesn't _really_ need to enforce the dominance property of SSA-form programs (i.e. all uses of a value are dominated by the definition of that value). The only property we would need to enforce is that the value dependency graph doesn't have cycles. In effect, by construction, the program will be in SSA form, even though we don't really have a control flow graph per se, more of a loose scheduling order.

In any case, renaming from `Variable` to something else is probably still a good idea, since it implies the notion of a mutable binding, but I wanted to try and understand the distinction between these and `Value`. If I understand correctly, you are using `Variable` and `Value` to distinguish between what I call `BlockArgument` and `OpResult` in my other comment, is that correct?

ir/src/ir/value.rs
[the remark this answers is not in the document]
bitwalker: > Our current design for the IR is closer to a sea-of-nodes than the more common CFG representation.

I should note that this is true of the representation described above as well - more precisely, it is a simplified representation of a Regionalized Value-State Dependence Graph (RVSDG), where the connections between nodes are entirely dictated by the value dependency graph. The exception to this, if you can call it that, are "SSA regions", i.e. regions with multiple blocks, where values are expected to adhere to the SSA dominance property - but you can also have regions where all operations go in a single block with no implicit scheduling order, beyond that dictated by the value dependency graph. It is basically up to the semantics of each operation to dictate what kind of regions it has. MLIR, which uses a similar representation, names these two kinds of regions "SSACFG" regions and "graph" regions.

> This also means that our SSA values can be extrapolated deterministically when pretty printing rather than being stored in the nodes directly.

The use of `ValueId` above is really one of debugging convenience, you can absolutely skip it, and assign integral identifiers during pretty printing. 

The tradeoff of that, is that it means that if you are debugging an issue with the compiler, dumping a vector of values doesn't give you a way to correlate those values to what you see in the pretty printed output (and from that, to the source code), all you get are opaque values (and some associated information like types, but that is commonly not enough to know precisely which value you are looking at, or how it was derived). It is convenient to be able to correlate the output from something like `dbg!(value)` with the pretty printed output and source code of the program/function/region, so that you can better understand what is happening. In any case, you could certainly omit it, I just find in practice that it is worth assigning a debugging-friendly canonical identifier for uses like that (one can also just perform this only when under `#[cfg(debug_assertions)]` or something, plenty of different ways to approach things of this nature).

> What would be the advantage of using an intrusive linked list here?

The biggest benefits are:

* You don't need to allocate any memory for containers that a node might be a member of, e.g., the list of operations in a block, a list of users of some symbol, all can be maintained by links between nodes directly. 
* It lets you manipulate nodes in the graph independently of their parents. 
* It permits a node to be present in multiple lists at once. For example, if you wanted to track symbol usage (where operations can be symbols), then an operation might have an intrusive link both for its parent block _and_ the use list of the symbol it uses (if it uses one).
* Many common transformations/rewrites rely on: inserting nodes at the front or back of some list of nodes, inserting a node relative to some other node; moving nodes around relative to their current position, or to the position of another node;  appending a list of nodes to another list, splicing a list of nodes into the middle of another list; splitting a list of nodes at arbitrary positions; and "stealing" a list of nodes from its current owner, leaving an empty list in its place. The intrusive linked-list is very efficient at all of these operations, in particular it is a constant-time operation to construct a cursor to a node in its containing list, so that you can then perform some operation on the list relative to that cursor (e.g. removing the node, splitting the list at that node, splicing another list at that node, inserting a new node). Vectors are great for cache locality, iteration, indexed access, and pushing to the end (assuming it doesn't need to reallocate the backing storage in the process) - but they are much less efficient at all of the other operations I mentioned above. 

Basically, it is an excellent data structure for a compiler IR which relies on an implicit graph, rather than an explicit one. The primary downside of linked lists is the potential for poor cache locality - but that is mitigated significantly by allocating nodes from an arena. I didn't go that far with my example code, but in practice that is usually what I do. Most traversals of an IR are jumping between different node types (regions -> blocks -> operations -> values -> operations -> blocks -> values -> ...), so cache locality is inherently an issue even when using vectors for node storage. Vectors are of course the primary alternative, using them as an arena, and using handles (indexes in some vector of a specific node type) for node-to-node references - implementing things like lists of regions, blocks, operations, etc., as vectors of handles. It's not a _bad_ alternative by any means, perhaps even better in some regards, but it does make some things more awkward in Rust, particularly mutation.


> Switch from sea-of-nodes to CFG: might delay work on passes and need a rewrite of the lowering AST->MIR. The 2 models are valid for an IR, with varying advantages and drawbacks.

As an aside: I want to push back on characterizing this as a CFG representation - it is an RVSDG, which is a form of dataflow graph. To the extent that there is any control flow represented, it is by convention only (i.e. the fact that operations are ordered in a block is incidental). What actually dictates whether a given region must _also_ represent a CFG, is whether or not you implement analyses/transformations over it that depend upon the SSA dominance property. Since AirScript doesn't have any real notion of control flow, only data flow, I think it is fair to say that we don't have any particular reason to require this of our IR, regardless of representation.

In any case, I certainly agree that switching representations is probably not worth it at this point, since I don't see any reason to assume that the current approach is not going to work out. My primary purpose in bringing it up and elaborating on it is two-fold:

1. The current sea-of-nodes representation is fine, aside from nits like splitting up the representation of values and operations, but I have two concerns (not unresolvable ones, but want to note them), and I wanted to frame it in comparison to the alternative approach that I probably would have taken if I was the one doing the work:
  * Uniquing nodes, and the interaction with program transformations. It may be that this isn't actually an issue in practice, but we'll want to keep an eye on this, and test thoroughly all of the potential edge cases.
  * The ability for less senior engineers to reason about this representation and work on the AirScript compiler. More thorough documentation once things are firmed up will be essential, particularly any assumptions made that have implications for correctness of the translation from the AST. For example, at what points are bindings resolved to SSA values, and does that occur before/after uniquing (presumably all bindings are resolved before uniquing, as let-shadowing is permitted in AirScript, but all of those kind of details should be outlined).
2. I had outlined the RVSDG design in our previous discussions elsewhere, (in #358 IIRC), but not elaborated on what that would actually look like in practice. I wanted to do so here in order to compare/contrast, and provide you with a concrete idea of what I'm talking about if I refer to things not present in this specific IR.

To be 100% clear, it was more to provide a background for discussion about this IR, rather than to dictate a new design.

> We agree that the node types should be more properly separated, we'll tend towards your proposed design on this aspect over time, incorporating elements as we need them for this implementation.

👍

>  Keeping the regionalised owner attributes would help simplify passes greatly. We'll incorporate that into our design.

I'm assuming it will be necessary in order to do per-region uniquing. However, after mulling it over further, perhaps it is not necessary to do per-region uniquing? I would like to better understand the implications of global uniquing, and I think that requires a broader set of tests and implemented functionality to see how it all interacts. It may very well be that there are no adverse effects because of how simple AirScript is in terms of primitive operations.

> Overall we think we can benefit from the proposed changes without necessarily changing the core design of the sea-of-nodes system we already have. This also minimizes time spent on redoing existing work.

Sure, I think that's fine. At the very least, you have an example of an alternative approach if you find you need to elaborate on the current representation to address any issues you run into. Again, the idea was more to compare/contrast and provide an explanation of what I'm talking about, rather than dictate how you implement the IR.

ir/src/passes/inlining.rs
bitwalker: > Wouldn't reusing existing nodes by splicing them during inlining cause conflicts in case a function is called twice (or more) with different arguments?

Sorry, I didn't mean stealing then splicing in this context, but rather that splicing the _cloned_ list of ops into the middle of the op list you want to inline into, is very efficient due to the intrusive lists, only requiring writes to two pointers.

The exception would be things like a comprehension that turns out to only have a single iteration, you wouldn't need to clone in that case, but could just steal the list from the comprehension body.

ir/src/passes/translate.rs
bitwalker: Ah, makes sense. The approach I've taken with this in the past, and which I've seen in other similar data structures, e.g. `cranelift_entity::PrimaryMap<K, V>`, have an underlying vector of `Option<V>`, or require `V` be default-constructible, so that:

* You can allocate indices before storing anything at that slot while ensuring that all indices are valid for the underlying vector, i.e. given an index, they can simply do `items.resize_with(index + 1, || None)` or `items.resize_with(index + 1, Default::default)`, to ensure that the index is valid, and returns something sensible. I think the `Option<V>` approach works best, since you can actually determine if a slot is occupied or not, but both options let you allocate indices before you construct the actual `V`.
* You can reuse slots, assuming you use the `Option<V>` representation, by maintaining the next available index (or a free list of them, depending on how you "release" nodes). You then check if you have a slot you can reuse when a new index is requested.
* The primary API can be basically `fn push(&mut self, node: V) -> K`, i.e. you store the node and get back its handle, but if you need to store the key in the node, you need to use two operations: `fn alloc_key(&mut self) -> K` and `unsafe fn insert(&mut self, index: K, node: V);`. The latter is marked unsafe since the index must be valid, and allocated by `alloc_key`, though it isn't strictly speaking an unsafe operation in terms of memory safety.

In any case, using that approach is definitely preferable to needing a special placeholder variant in your node type that you then need to handle everywhere even though you know it will never appear in practice.

ir/src/passes/inlining.rs
bitwalker: Ok, I was under the impression that `inline_call` transitively visits all inlineable items reachable from the root node being visited, but if I understand your description correctly, it is only visiting the immediate child of the `Call` node, i.e. it isn't also inlining calls in the callee, just the body of the callee without further inlining. Is that correct?

If so, then yeah we're good - I think I assumed the inliner was recursing into children, which is what prompted my comment.

bitwalker:
> @bitwalker, first of all, thank you for this detailed answer, it will sure help a lot!

You bet! Sorry if presenting a sketch of the RVSDG design gave you a scare - didn't want to stress you out worrying about needing to refactor things, hopefully it just provided a useful reference for past and future conversations.

> We have left several comments trying to address all the points you have raised. We hope that helps clarifying the design we proposed and how we plan to incorporate your feedback.

Yes, I believe all of my questions are addressed, except maybe a few small fresh ones I've left as comments. Ping me once a few of the more complex tests are implemented, and I'll re-review ASAP.

Thanks for all the hard work!

ir/src/ir/value.rs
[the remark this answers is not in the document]
Soulthym: > I should note that this is true of the representation described above as well - more precisely, it is a simplified representation of a Regionalized Value-State Dependence Graph (RVSDG), where the connections between nodes are entirely dictated by the value dependency graph

I see, I think we misinterpreted your proposed implementation here, you can safely discard any mention of a CFG in our previous comment.
After thinking more about it, your design does seem to present another advantage in the case of dead code elimination. Since nodes are referenced by pointer (or `Rc`) and not their NodeIndex in the Vec of all nodes, this would enable automatic cleaning of unreferenced nodes without worrying about having to shift subsequent NodeIndices in order to compactify the graph. Using reference counting makes this a lot easier to manage than doing it manually

> The use of ValueId above is really one of debugging convenience, you can absolutely skip it, and assign integral identifiers during pretty printing.

Noted, this is a very good point I hadn't thought about, we'll add tracking of ssa values to the nodes

> The biggest benefits are: ... 

I see what you mean, thanks for the detailed explanation. We'll probably not implement this immediately, it would probably be worth measuring real performance before optimizing this part. We'll keep that structure in mind as a potential future optimization.

> As an aside: I want to push back on characterizing this as a CFG representation - it is an RVSDG, which is a form of dataflow graph

Noted, that is mostly due to our misunderstanding of your proposed design. After thinking about it longer, I am under the impression that the 2 designs are functionally equivalent.

> To be 100% clear, it was more to provide a background for discussion about this IR, rather than to dictate a new design.

No problem! And thank you for this detailed feedback, it clears a few points I hadn't understood properly.

> I'm assuming it will be necessary in order to do per-region uniquing. However, after mulling it over further, perhaps it is not necessary to do per-region uniquing? I would like to better understand the implications of global uniquing, and I think that requires a broader set of tests and implemented functionality to see how it all interacts. It may very well be that there are no adverse effects because of how simple AirScript is in terms of primitive operations.

That was our impression aswell, it seemed a bit irrelevant after the inlining/unrolling passes since they remove all function "scopes" and only keeps the root block. Note that we havent fully thought through the implications for conditional's and loop's blocks.
Our current understanding is that keeping it regionalized makes sense if we want to have pre-inlining/unrolling optimizations, since after inlining/unrolling, those would all be sharing the same scope anyway.

We'll implement regionalisation for now, but might try to remove that requirement in the future if at all possible.

> At the very least, you have an example of an alternative approach if you find you need to elaborate on the current representation to address any issues you run into. Again, the idea was more to compare/contrast and provide an explanation of what I'm talking about, rather than dictate how you implement the IR.

Thanks again for this!

Soulthym:
> Sorry if presenting a sketch of the RVSDG design gave you a scare - didn't want to stress you out worrying about needing to refactor things, hopefully it just provided a useful reference for past and future conversations.

No problem! It will serve as a reference for further discussions, and as a target to tend towards as we refactor our implementation based on what we need. 

> Yes, I believe all of my questions are addressed, except maybe a few small fresh ones I've left as comments. Ping me once a few of the more complex tests are implemented, and I'll re-review ASAP.

I have left comments addressing these. For now I don't think the remaining points require immediate attention, we already have plenty to do based on your feedback. I'll be sure to ping you when a review is needed.

Thanks again for all your valuable feedback!

Leo-Besancon:
Hello @bitwalker !

We wanted to give you an update on what we've been working on, and to get your input on a few things.

Our main rework working branch is the following: https://github.com/massalabs/air-script/tree/thy-rework-ir

We've mainly worked on:
- A rework of the base structure for the Mir based on your proposal (without NodeIndex, Operation directly references their children / parents). @Soulthym will give details on this aspect below.
- Implemented a generic visitor pattern for the Mir -> Mir passes
- Improved Inlining based on previous comments
- Implemented Unrolling (on the previous structure) 
- Implemented Lowering from Mir to Air (on the previous structure)

What we have to do:
- Complete rework of the structure
- Now that we have the full pipeline AST -> Mir -> AIR:
    - Adapt a few pass behaviors to the new structure. For instance, unrolling List Comprehensions still relies on incomplete logic as it needs the new structure to be implemented cleanly
    - Make current tests on the Air pass (with the whole pipeline, we can better debug edge cases).
        - We have unreachable()! triggered for a lot of tests, debug and resolve these issues
        - We have infinite loops for some tests to investigate
- Fully test everything

As a result, it would be great if you could look at the structure and answer our questions below so we can finalize the design!

Thank you again.

Soulthym:
Hey! 

Below is a summary of the new structure for the IR.
We wonder which of 1.a) or 1.b) would be the best approach to take, what's your thought on this?

# Our rework of the MirGraph structure:
## Our current structure:
Implemented traits are marked [Trait].
Rc/RefCell/Weak are ignored for clarity.
Fields not referenced in the Problems & Solution sections are also omitted.

```python
dyn Op: gives access to the Operation 
dyn Value: * shared behaviour by all nodes

Operand: [Value]
  owner: Operation
  value: dyn Value
OpResult: [Value]
  owner: dyn Op
Operation: [Value]
  owner: dyn Op
  Vec<Operand>
  result: OpResult

Add: [Op, Value]
  Operation
...
Matrix: [Op, Value]
  Operation

Parameter: [Value]
  owner: FunctionOrEvaluator

Block:
  owner: FunctionOrEvaluator
  Vec<dyn Op>

Fonction/Evaluator:
  params: Parameter
  blocks: Vec<Block>
```

# Problems & Potential Solutions:
## P: Dynamic dispatch makes things tricky due to casting to non-primitive types.
S: Replace with enums.

## P: Rc<dyn Op> only works on operations.
S: Make Function, Block and Evaluator [Op] that wrap Operation.

## P: Rc<T> is immutable.
S1: Replaced by RefCell in most cases, but we loose ref-counting.
S2: We can use Rc<RefCell<T>.

## P: Ref-counting cycles leak memory.
S: use RefCel<Weak<T>>.

## P: You can't always know the owner ahead of time.
S: Some need to be Option<T>, mainly when inserting existing nodes into a block.

# Proposed solutions:
## 1) To help match the semantics to the actual structure and avoid dynamic dispatch.
Either:
### a) Rename traits and make them static:
- Replace dyn Op with impl Parent:
  exposes:
  - insert_child -> impl Child
  - get_children -> Vec<impl Child>
  - as_operation -> Operation
- Implement Parent for Function, Evaluator and Block
- Replace dyn Value with impl Child:
  exposes:
  - get_parent -> impl Parent
- Implement Child for all except {Function, Evaluator}
- Add empty Leaf trait (not sure if necessary):
  - allows expressing not(Parent) in generic trait bounds.
- Add enums for all node kinds, to restore the original api semantics:
  - Node: * (all kinds) (old: Value)
  - Ops: Add, Mul, Sub, Felt, Vector, Matrix, Enf, For, If, Call, Fold
  - HasBlock: Function, Evaluator, If, For

### b) Use traits on top of enums:
- Add enums for all node kinds, to restore the original api semantics:
  - Node (old: dyn Value): * (all kinds)
  - Ops (old: dyn Op): Add, Mul, Sub, Felt, Vector, Matrix, Enf, For, If, Call, Fold
  - HasBlock: Function, Evaluator, If, For
  - Parents: Ops + HasBlock - (Felt, Vector, Matrix)
  - Children: Ops + Block
- Implement Op, Value, Child, Parent on top of the enums
- Reference the enums in nodes instead of the nodes diectly

## 2) To avoid memory leaks and make owners mutable
Instead of Rc<dyn Op>, make owner fields for structs implementing [Child]:
- if 1.a)
  - Optional<RefCell<Weak<impl Parent>>>
- elif 1.b)
  - Optional<RefCell<Weak<Ops>>>

Soulthym:
Hello @bitwalker,
Since our last discussion, we added:

- new graph structure for the ir based on enum dispatching
- split NodeType into sub-categories (LeafNode, MiddleNode, RootNode)
- single depth-level uniquing within scopes
- Derive macro for IsNode, used for operations that wrap a Node (operations containing other nodes like Function, Eval, Add, Mul...)

Next step is to replace the old structure by this new one, which will be done through this PR: https://github.com/massalabs/air-script/pull/4
(The various sub-tasks are listed in the PR comment)

Soulthym:
*Conversation moved from https://github.com/massalabs/air-script/pull/4*

*The [original message](https://github.com/massalabs/air-script/pull/4#issuecomment-2633598154) was in response to:
https://github.com/massalabs/air-script/pull/4#issuecomment-2612830365*

Hey @bitwalker,
I think I understand the design from a high-level perspective, and overall think it would be a very welcome upgrade to the current implementation.
I am however somewhat still confused about the following points:

## Phi Nodes in Block
> 
> ```rust
> /// The value representation for block parameters. These represent distinct
> /// SSA values, so that transformations within a block can be done without
> /// having to know how the block is reached. These are equivalent
> /// to Phi nodes in SSA literature, but in a more intuitive form.
> pub struct BlockArgument {
>     owner: WeakEntity<Block>,
>     ty: MirType,
>     users: UseList,
> }
> ```

I'm not sure I understand the relationship with Phi nodes. I thought those reconciled `Block` results in SSA style IRs, as in they represent a ternary operation of both `Blocks`, as-in an if else statement?

Regardless, I understand their usage as replacement targets with bound values during inlining/unrolling, so it shouldn't be much of an issue in practice, I am mostly asking in case I missed other cases that would behave differently.

## parent, owner, and use_list
In the current design, we use the field `owner` to store the operations that use an `Op` accessed with the `Parent` trait, and do not have the corresponding `block` stored.

Given that, am I understanding the following correctly? 
Starting from the current design and ending at yours, semantically:
- an `owner` becomes a `user` in the operation's `UseList`
- a `parent` becomes an `owner` as in the parent in the Mir graph, minus the owning block (cf next)
- `parent` now stores the operation's owning `Block`(Function, Evaluator, For, If,...) as well in all cases.

## Future progressive migration

### Checklist before merging

Our design is passing all existing tests correctly, but still needs some improvements.

Our plan is to do the following, roughly in this order:
- [x] add span fields
- [x] propagate spans through transformations (translation AST -> Mir and passes)
- [ ] document public APIs

### Checklist after merging (to do according to the priority of the other milestone)

Once merged upstream, we plan to:
- [ ] simplify the boilerplate
- [ ] refactor/simplify complex parts:
  - [ ] really use the visitor in passes. Currently passes re-implement dispatching instead of using the internal dispatcher.
  - [ ] replace the lazy part of singletons by creating them during struct instantation, simplifying their update mechanism (less cases to cover)
- [ ] add a pass of constant propagation/folding in the IR (these causes the only differences in codegen for Winterfell):
  - [ ] remove useless nodes, such as multiplying by 1, adding 0, etc.
  - [ ] simplify exponentiation where possible. For example: `2.pow(3)` turns into `2*2*2` but what we really want is `8`.

#### Migrate in parallel

At this point we plan on working on the next steps, while migrating to your design progressively in the background.

Migration checklist:
- [ ] add parents automatically within the builder pattern, currently not present in most places.
- [ ] implement an alternative update mechanism, that mutates the contents of parent structs instead of primitive enums (Op and Root), to match your design's behaviour.
- [ ] migrate shared mutability usage to the new update mechsnism
- [ ] rename functions/methods/fields in our API to match yours where update behaviour is the same.
- [ ] implement the Block abstraction, like it is in your design.
- [ ] add owner references, like in your design.
- Then, simulatenously:
  - [ ] replace our uses of primitive enums (Op and Root, those that own the inner struct) with their respective object safe dyn Traits
  - [ ] remove our uses of wrapper enums (mainly Owner and Node, those that contain a BackLink to the primitive enum) as they can now be used by dyn Traits
  - [ ] swap current remaining structs' implementation with your design.

I have updated previous checklists, marking items with ~~strikethrough~~ that became irrelevant with our latest modifications.

bitwalker:
> Hey @bitwalker, I think I understand the design from a high-level perspective, and overall think it would be a very welcome upgrade to the current implementation. I am however somewhat still confused about the following points:
> 
> ## Phi Nodes in Block
> > ```rust
>>  ..snip..
> > ```
> 
> I'm not sure I understand the relationship with Phi nodes. I thought those reconciled `Block` results in SSA style IRs, as in they represent a ternary operation of both `Blocks`, as-in an if else statement?

Phi nodes are needed anywhere that there is a join point in dataflow. Typically this reflects returning a result from a conditional statement, or any type of loop with loop-carried arguments/results. Without them, it's not possible to keep a program with those constructs in SSA form, and SSA form is essential for a lot of analyses/transformations (note: of course, there are other SSA-equivalent representations, such as continuation-passing style, but I'm ignoring those for our purposes here, because those representations come with other challenges). Block arguments are just a more intuitive representation of the phi concept, which also enable some other conveniences.

One such convenience with block arguments, is that they can also be used even when there isn't a join point that needs to be handled. They in effect create aliases of the block inputs, which are then scoped to blocks dominated by the block in which those aliases were introduced. For example, inlining a function body can rely on this so that a `call` can be rewritten as a `br` instruction, with the callee body inlined into the current function. Later, the redundant aliasing of the block arguments can be eliminated, by simply replacing all of the uses of the block arguments with the original inputs - but you may not want to do that if there are multiple calls in the caller to the same function that was inlined; preserving the block arguments lets you just inline once.

> Regardless, I understand their usage as replacement targets with bound values during inlining/unrolling, so it shouldn't be much of an issue in practice, I am mostly asking in case I missed other cases that would behave differently.

Yeah I suspect in practice, block arguments due to join points in dataflow will be less common, and using them for aliasing during inlining will be the most common use case.

One place where we might find ourselves wanting something that would normally call for dataflow joins, are operations like `min` or `max`, where the result of the operation is conditional on the inputs. Unless that is a primitive operation, you may need to represent the conditional nature of it explicitly in the IR (i.e. by actually encoding the computation of minimum or maximum operand), at least temporarily. For the most part, such operations can often be collapsed into a `select`-like operation without conditional branches, so long as your target ISA has appropriate primitives.

> ## parent, owner, and use_list
> In the current design, we use the field `owner` to store the operations that use an `Op` accessed with the `Parent` trait, and do not have the corresponding `block` stored.
> 
> Given that, am I understanding the following correctly? Starting from the current design and ending at yours, semantically:
> 
> * an `owner` becomes a `user` in the operation's `UseList`
> * a `parent` becomes an `owner` as in the parent in the Mir graph, minus the owning block (cf next)
> * `parent` now stores the operation's owning `Block`(Function, Evaluator, For, If,...) as well in all cases.

I think I would characterize it as:

* The immediate parent of an operation (if it has a parent) is always a `Block`
* The immediate parent of a `Block` is always an operation
* You can reach an ancestor op of an operation by traversing the op -> block -> op chain
* Operations don't have users, values do
* An operation is used if any of its results are used
* The values used by an operation correspond to its operands
* A block argument with no uses is dead and can be removed

So there are really two distinct concepts: 

1. Parents/children of nodes in the IR graph, of which there are three main kinds: Operation, Block, and Value. This is more about the basic structure of the IR, and how it is traversed generically; although it also reflects the control flow graph (note however that the presence of a control flow graph does not imply a specific scheduling of operations at runtime, that's up to the code generator). To be more precise, the control flow graph is given by the relationship between operations and blocks.
2. Value definitions and uses, i.e. the use/def graph, which reflects the dataflow of the program. This is really where I think the most significant change lies, as the current IR you've assembled does not represent the value use/def graph explicitly, and this complicates certain analyses/transformations, and to some extent constrains operations (i.e. operations may not have multiple results, and the semantics of "using" an operation with no results is unclear; although I don't believe we have any such cases now, it is technically permitted by the structure).

Let me know if there are other details you'd like clarity/elaboration on, I'm happy to do so!

Leo-Besancon:
Hello @bitwalker I've updated the toolchain config file to latest stable and re-ran clippy, it should fix the failing CI if you want to re-trigger it. Or tell me if you had something else in mind!

bitwalker (approved):
LGTM, the only feedback I really have we've already discussed, so I think we can merge this for now.

PR #363, opened 2024-12-22
docs: typo fix Update keywords.md

defitricks:
There is a typo in the section describing the `const` keyword. The link points to `./declarations.md#constant-constant`, but it should be `./declarations.md#constants`.

Corrected.


bitwalker:
We no longer accept PRs of this type, without some other meaningful change being contributed. Unfortunately, people have started trying to farm for crypto rewards by contributing these sort of frivolous changes to various projects in the ecosystem.

PR #365, opened 2025-01-08
Update book.yml

crStiv:
### Key changes:

- Added a permissions section to explicitly specify required permissions
- Updated actions/checkout from v2 to v4
- Added a name for the checkout step
- Removed quotes around "latest" as they are not necessary in this context
- Improved formatting to make it more consistent

These changes will make your workflow more modern, secure, and easier to read.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -5,11 +5,16 @@ on:
     branches:
       - next
 
+permissions:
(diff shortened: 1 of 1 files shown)


bobbinth:
Thank you - but I think we can probably keep the `book.yaml` as is.

PR #367, opened 2025-01-30
Update LICENSE

maximevtush:
Updated the copyright year from 2022 to 2025 in the LICENSE file


bitwalker:
It is not necessary to update copyright years, it is the year it was established, not the current year.

PR #368, opened 2025-01-30
fix: Remove unnecessary pub before main Update main.rs

mdqst:
`main` was mistakenly declared as `pub fn main()`, which isn't needed


bitwalker:
I'd accept this change as part of a larger set of fixes, but we avoid merging nits like this on their own, as it takes up too much of our time. Thanks for the submission anyway, and feel free to reopen with other changes if you wish!

PR #370, opened 2025-02-20
ACE backend: a code generator from AirScript to the ACE chiplet of Miden's recursive verifier

paracetamolo:
This backend generates a table in the format required by the ACE chiplet as described in https://github.com/0xPolygonMiden/air-script/issues/366.

Currently the table is in the Graphviz dot format, the textual form is a superset of what is needed for the ACE chiplet and can be used to verify that rows are in the right order. The image generated can be used to debug the connections in the graph.

```
cd codegen/ace
cargo test -- test_ace && dot -Tsvg ace.dot > ace.svg
```


Al-Kindi-0:
Do we need to change the base branch from `main`?

paracetamolo:
> Do we need to change the base branch from `main`?

It's based on a branch on the Massa Labs repo `massa/feat-implement-ir`, github doesn't show it for this repo.

codegen/ace/src/codegen.rs
```diff
@@ -62,6 +62,50 @@ fn table_to_string(table: Vec<Operation>) -> Result<String, std::fmt::Error> {
...
+// - 20 bits for left
+// - 20 bits for right
+pub fn operation_to_felt(out: usize, op: Operation) -> Felt {
+    let out = (out as u64) << 40;
+    let op_and_ins: u64 = match op {
+        Operation::Value(Value::PublicInput(_))
+        | Operation::Value(Value::RandomValue(_))
+        | Operation::Value(Value::TraceAccess(_)) => 0u64 << 60,
+        Operation::Value(Value::Constant(_)) => 1u64 << 60,
+        Operation::Value(Value::PeriodicColumn(_)) => panic!("periodic"),
```
Al-Kindi-0, line 81: I added a comment in the issue on a proposal to tackle `Value::PeriodicColumn`

codegen/ace/src/codegen.rs
```diff
@@ -62,6 +62,50 @@ fn table_to_string(table: Vec<Operation>) -> Result<String, std::fmt::Error> {
...
+// - 20 bits for out
+// - 20 bits for left
+// - 20 bits for right
+pub fn operation_to_felt(out: usize, op: Operation) -> Felt {
+    let out = (out as u64) << 40;
+    let op_and_ins: u64 = match op {
+        Operation::Value(Value::PublicInput(_))
+        | Operation::Value(Value::RandomValue(_))
+        | Operation::Value(Value::TraceAccess(_)) => 0u64 << 60,
+        Operation::Value(Value::Constant(_)) => 1u64 << 60,
```
Al-Kindi-0, line 80: We need to also serialize the constant value

Al-Kindi-0 (commented):
As a draft, looks great, thank you!
Could we add some tests to test for the correct evaluation on some small examples?
For example, and for small circuits, we can compare the output of the circuit to the value we get from the algebraic expression.

codegen/ace/src/codegen.rs
paracetamolo: done

paracetamolo:
Added support for periodic columns as described [here](https://github.com/0xPolygonMiden/air-script/issues/366#issuecomment-2697324777), this required a bit of a refactor that overall improved the quality of the code.

bobbinth:
@paracetamolo - seems like this PR includes changes from other PRs. It would be good to rebase it from the latest `next` branch. This way, it'll contain only the relevant changes.

paracetamolo:
Done. I had rebased the branch on `next` a while ago but forgot to change the base of the PR.

paracetamolo:
Things left to do:
- test periodic columns. Most of the AirScript examples used in the MASM backend tests are also tested in the ACE backend, except for the ones using periodic columns.
- incorporate the recent changes described in https://github.com/0xPolygonMiden/air-script/issues/366#issuecomment-2751848648

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,257 @@
...
+/// - `zn_1 = z^n -1`
+///
+/// Which can be rewritten as
+/// `int * z_g^2 * z_1 + bf * zn_1 * z_g + bl * zn_1 * z_1 / (zn_1 * z_1 * z_g)`
+/// This function returns a circuit computing the numerator above, given the following inputs:
+/// - alpha
+/// - z
+/// - z^n
+/// - g^(-1)
+pub fn build_ace_circuit(ir: &Air) -> anyhow::Result<(NodeIndex, Circuit)> {
```
bobbinth, line 161: nit: this seems to be the main function in this module - I'd move it to the top of the file. The other functions/structs could go into helper sections after it.

codegen/ace/src/serialization.rs
```diff
@@ -0,0 +1,138 @@
...
+// from most significant to least:
+// - 3 bits for the OPCODE, with READ 0, CONST 1, ADD 2, MUL 3, SUB 4
+// - 20 bits for out
+// - 20 bits for left
+// - 20 bits for right
+// Left and right are only used for ADD MUL and SUB and are expected to be zero
+// in the other cases.
+// When a felt encodes an opcode CONST and its out position, the following felt
+// must encode the 64bit constant.
+pub fn operation_to_felts(out: usize, op: Operation) -> Vec<Felt> {
```
bobbinth, line 76: This needs to be updated to be consistent with the lates specs.

Also, shouldn't this return a single field element? I think the only reason we have a vector as the return value is to handle the `Constant` operation, right? If so, is the current treatment of constants correct? I though they would be provided in a way similar to inputs - i.e., there would not be a lot of difference between inputs and constants.

codegen/ace/src/serialization.rs
```diff
@@ -0,0 +1,138 @@
...
+        Value::PeriodicColumn(_) => "Per".to_string(),
+        Value::PublicInput(acc) => format!("PI {} {}", acc.name, acc.index),
+        Value::RandomValue(v) => format!("Random {}", v),
+        Value::TraceAccess(acc) => {
+            format!("Trace {} {} {}", acc.segment, acc.column, acc.row_offset)
+        }
+    }
+}
+
+fn operation_to_dot(out: NodeIndex, op: &Operation) -> Result<String, std::fmt::Error> {
```
bobbinth, line 20: Should this be a method of the `Operation` struct? - e.g.,

```rust
impl Operation {

    pub fn to_dot(&self, out: NodeIndex) -> Result<String, std::fmt::Error> {
        ...
    }

}
```
Let's also add doc comments to this function.

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,436 @@
...
+            let op = remap_op(*op, &mapping);
+            res.push(op);
+            mapping.insert(*out, (res.len() - 1).into());
+        }
+        (mapping[&root], Circuit(res))
+    }
+}
+
+#[test]
+fn test_sum() {
```
bobbinth, line 380: nit: we usually put tests into `tests` module.

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,436 @@
+use crate::comparison::compare_operation;
+use crate::serialization::{operation_to_felts, operations_to_dot};
+use air_ir::{NodeIndex, Operation, TraceAccess, Value};
+use miden_core::Felt;
+use std::cmp::Ordering;
+use std::collections::{btree_map::BTreeMap, btree_set::BTreeSet};
+
+pub struct Circuit(Vec<Operation>);
```
bobbinth, line 8: It may make sense to make the circuit a bit more structured:

- Have a field that describes the shape of expected inputs and constants.
- Maybe have something like the `Operation` struct but specifically tailored to the needs of this crate which more closely maps to the gate specifications.
- Arrange the operations in the order in which they will appear in the serialized form.

bobbinth (commented):
Thank you! Looks good! I left some comments inline. Many of them are about improving code organization to be more inline with our usual style, but I also think that this needs to be updated to be inline with the latest ACE chiplet design.

codegen/ace/0.dot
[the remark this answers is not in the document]
paracetamolo: They are regression files to detect when we change the compilation output. I was looking for a "Rust way" to place these files inside the test directory, I ended up using the environment variable `CARGO_MANIFEST_DIR`, not sure if there is a better way. Now they are placed in `codegen/ace/tests/regressions/`.

codegen/ace/src/serialization.rs
paracetamolo: It is just an private internal function for `operations_to_dot`, not really worth exposing it.
I documented the functions.

codegen/ace/src/circuit.rs
paracetamolo: done

codegen/ace/src/circuit.rs
paracetamolo: In this PR I was trying to reuse as much as possible `air_ir` where the `Operation` type comes from. It is the type that the ACE backend receives and, after a lot of processing, also the one that it outputs. For the output part we don't need all that structure so we can define a simpler type closer to the serialization format but I would leave this as a future PR so we can agree on the right design with @adr1anh.

codegen/ace/tests/utils.rs
```diff
@@ -0,0 +1,26 @@
+use air_codegen_ace::Circuit;
+use air_ir::NodeIndex;
+use miden_diagnostics::{
+    term::termcolor::ColorChoice, CodeMap, DefaultEmitter, DiagnosticsHandler,
+};
+use std::sync::Arc;
+
+pub fn codegen(source: &str) -> (NodeIndex, Circuit) {
```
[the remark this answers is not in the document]
paracetamolo: done

paracetamolo:
Update:
- I addressed most comments.
- Converted `eval` and tests to `QuadExtension`.
- `z^min_num_cycles` now is always an input, even if the circuit does not have periodic columns.
- I fixed a bug where unused inputs would be optimized away. Now inputs are always preserved, even if unused.

Todo:
- update formula as in point 7 of https://github.com/0xPolygonMiden/air-script/issues/366#issuecomment-2751848648
- fix serialization format

codegen/ace/src/serialization.rs
paracetamolo: The serialization you reviewed was the old one. Now I updated it and, as you mentioned, constants correspond to a single field element.

codegen/ace/src/serialization.rs
```diff
@@ -0,0 +1,181 @@
+use air_ir::{NodeIndex, Operation, Value};
+use miden_core::{Felt, FieldElement, QuadExtension};
+use std::fmt::Write;
+
+pub type Quad = QuadExtension<Felt>;
+pub fn to_quad(e: u64) -> Quad {
+    QuadExtension::new(Felt::new(e), Felt::ZERO)
+}
```
adr1anh, line 8: Nit: The type `QuadExtension` implements `From<u64>` which could be invoked instead of this methd

codegen/ace/src/comparison.rs
```diff
@@ -0,0 +1,64 @@
...
+    if segment == Ordering::Equal {
+        let row = a.row_offset.cmp(&b.row_offset);
+        if row == Ordering::Equal {
+            a.column.cmp(&b.column)
+        } else {
+            row
+        }
+    } else {
+        segment
+    }
```
adr1anh, line 16: Nit: Tuples implement `Ord` if all its components implement `Ord`. If you need a different ordering of the structs, this should be simpler
```suggestion
    (a.segment, a.column, a.row_offset).cmp(&(b.segment, b.column, b.row_offset))
```

codegen/ace/src/comparison.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+fn value_order(a: &Value) -> usize {
```
adr1anh, line 19: This and the methods below look like an implementation of [`PartialOrd`](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html) which is like `Ord` but instead of `cmp(..) -> Ordering` we have `partial_cmp(..) -> Option<Ordering>`. Returning an `Option<usize>` here instead of `unreachable` would shift the responsibility to the caller to `unwrap` if they can ensure the ordering is always valid.

codegen/ace/src/comparison.rs
```diff
@@ -0,0 +1,64 @@
...
+    match a {
+        Value::PublicInput(_) => 0,
+        Value::TraceAccess(_) => 1,
+        Value::RandomValue(_) => 2,
+        Value::Constant(_) => 3,
+        Value::PeriodicColumn(_) => unreachable!("periodic"),
+    }
+}
+
+fn compare_value(a: &Value, b: &Value) -> Ordering {
```
adr1anh, line 29: Same here, returning `Option<Ordering>`

codegen/ace/src/comparison.rs
```diff
@@ -0,0 +1,64 @@
...
+fn value_operation(a: &Operation) -> usize {
+    match a {
+        Operation::Value(_) => 0,
+        Operation::Add(_, _) | Operation::Sub(_, _) | Operation::Mul(_, _) => 1,
+    }
+}
+
+/// Compare functions used to sort the operations as needed by the ace chiplet.
+pub fn compare_operation(a: &Operation, b: &Operation) -> Ordering {
+    let cmp = value_operation(a).cmp(&value_operation(b));
```
adr1anh, line 55: Since `Operation` already derives `Ord`, you could just set `cmp = a.cmp(b)`, since the derived implementation guarantees lexicographical comparison of struct fields.

air/src/graph/mod.rs
```diff
@@ -23,6 +23,18 @@ impl core::ops::Add<usize> for &NodeIndex {
     }
 }
 
+impl From<usize> for NodeIndex {
```
adr1anh, line 26: There is a blanket implementation `Into<U> for T` when we implement `From<T> for U` [source](https://doc.rust-lang.org/std/convert/trait.From.html#generic-implementations). This mean you can write
```rust
let n = NodeIndex::default();
let u: usize = n.into().
```

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,507 @@
...
+            }
+        }
+        None
+    }
+}
+
+/// Remap the children of an operation. Typically when migrating an operation
+/// between data structures.
+pub fn remap_op(op: Operation, mapping: &BTreeMap<NodeIndex, NodeIndex>) -> Operation {
+    use Operation::*;
```
adr1anh, line 166: Importing all enum variants is generally discourages since it may lead to incorrect matches when the enum variants change. In this case it would be okay, but if the variants did not have fields inside them, a removed field would be matched as the "catch-all" case.

paracetamolo:
Done:
- updated formula as in point 7 of
https://github.com/0xPolygonMiden/air-script/issues/366#issuecomment-2751848648
- fixed serialization format
- tested circuits with periodic columns

Todo:
- documenting some functions
- there is something fishy with public inputs, need to test it better

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,507 @@
...
+            .collect();
+        Self {
+            cache,
+            circuit,
+            n_inputs: 0,
+            zero: constants[0],
+            one: constants[1],
+            alpha: None,
+            next_alpha: constants[1],
+        }
```
adr1anh, line 226: Arrays have a `map` method that returns an array of the same size. It can be destructured in-place. This avoids an extra `Vec` allocation
```suggestion
        let [zero, one] = [0, 1].map(|c| {
            let op = Operation::Value(Value::Constant(c as u64));
            let out = circuit.append(op);
            cache.insert(op, out);
            out
        });
        Self {
            cache,
            circuit,
            n_inputs: 0,
            zero,
            one,
            alpha: None,
            next_alpha: one,
        }
```

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,507 @@
...
+                let out = self.circuit.append(op);
+                self.cache.insert(op, out);
+                out
+            }
+        }
+    }
+
+    /// Append an operation to the circuit after applying some optimizations
+    /// from the algebraic laws of addition, multiplication and subtraction.
+    pub fn append(&mut self, op: Operation) -> NodeIndex {
```
adr1anh, line 245: Again, `push` would be match `Vec`

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    pub fn sub(&mut self, l: NodeIndex, r: NodeIndex) -> NodeIndex {
+        self.append(Operation::Sub(l, r))
+    }
+
+    /// `\sum_i vec[i]`
+    pub fn sum(&mut self, vec: &[NodeIndex]) -> NodeIndex {
+        assert!(!vec.is_empty());
+        vec.iter().fold(self.zero, |acc, r| self.add(acc, *r))
```
adr1anh, line 323: I think adding zero is optimized, but `reduce` ensures that we don't perform a trivial addition.

```suggestion
        vec.iter().reduce(|acc, r| self.add(acc, *r)).unwrap_op(self.zero)
```

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    /// `\prod_i vec[i]`
+    pub fn prod(&mut self, vec: &[NodeIndex]) -> NodeIndex {
+        assert!(!vec.is_empty());
+        vec.iter().fold(self.one, |acc, r| self.mul(acc, *r))
```
adr1anh, line 329: Same as above
```suggestion
        vec.iter().reduce(|acc, r| self.mul(acc, *r)).unwrap_op(self.one)
```

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,507 @@
...
+            }
+        }
+        res
+    }
+
+    /// Evaluates a polynomial with coefficients `coeffs` on a point `point`,
+    /// using Horner's method.
+    /// `\sum_i coeffs[i] * x^i`
+    pub fn horners_method(&mut self, point: NodeIndex, coeffs: &[NodeIndex]) -> NodeIndex {
+        coeffs.iter().rev().fold(self.zero, |acc, coeff| {
```
adr1anh, line 353: Same as above for `reduce`

adr1anh (commented):
Not a full review of the functionality, but left some code-related comments as I tried getting familiar with the circuit.

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,509 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    fn append(&mut self, op: Operation) -> NodeIndex {
+        self.0.push(op);
+        (self.0.len() - 1).into()
+    }
+
+    fn iter(&self) -> core::slice::Iter<Operation> {
```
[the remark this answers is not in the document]
paracetamolo: Right. I removed this code now, but good note for the future.

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,509 @@
...
+
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    fn append(&mut self, op: Operation) -> NodeIndex {
```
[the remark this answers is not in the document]
paracetamolo: done

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,509 @@
...
+        self.0.push(op);
+        (self.0.len() - 1).into()
+    }
+
+    fn iter(&self) -> core::slice::Iter<Operation> {
+        self.0.iter()
+    }
+
+    fn get_inputs(&self) -> Vec<NodeIndex> {
```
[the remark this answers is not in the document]
paracetamolo: Right. I removed this code now, but good note for the future.

codegen/ace/src/circuit.rs
paracetamolo: Very nice. Done.

codegen/ace/src/circuit.rs
paracetamolo: done

codegen/ace/src/circuit.rs
paracetamolo: Nice to learn about reduce. I tried but it gets messy with the references. I would leave it a fold for now.

codegen/ace/src/circuit.rs
paracetamolo: Same as above

paracetamolo:
Update:
- I did a large refactor of input handling. Before it was possible to define inputs anywhere during the circuit definition and they would be ordered later. Right now all inputs must be defined at circuit creation, which simplifies some things.
- Defined the arrangement of all inputs of the Airscript and ACE circuits.
- Improved testing a lot, typically to correctly test public inputs, random values, and next-row values.

paracetamolo:
Note on future work.
This PR uses the `Operation` type because it was rich in information and we were not entirely sure of the design.
Now that the design is clearer, a new much simpler type can be defined with simply `Add, Mul, Sub` so that most of the unreachable patter matchings can be removed and also `NodeIndex` which causes a bit of overhead.
This is left as future work to be done at the same time as VM trace generation, which will share code.

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,342 @@
...
+            let res = cb.prod(&[lc, z_n_minus_one, z_minus_g]);
+            els.push(res)
+        };
+        // bl * zn_1 * z_1
+        if !boundary_last_roots.is_empty() {
+            let boundary_last_roots = boundary_last_roots.iter().map(|r| mapping[r]).collect();
+            let lc = cb.linear_combination_alpha(boundary_last_roots);
+            let res = cb.prod(&[lc, z_n_minus_one, z_minus_one]);
+            els.push(res)
+        };
```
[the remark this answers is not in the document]
Al-Kindi-0: But isn't `CircuitBuilder::next_alpha()` there for exactly this purpose? or maybe there is something else?

paracetamolo:
Rebased branch on `next`, after busses. Just need to squash the history.

codegen/ace/src/codegen.rs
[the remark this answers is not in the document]
paracetamolo: Unless I made I mistake somewhere, `linear_combination_alpha` keeps the successive `alpha` stored in the `CircuitBuilder` state. So the three calls use always different alphas. I was undecided if to move the alpha as an argument and to pass them explicitly to make the code clearer but at the cost of making a mistake.
I added a test to check that two calls to `linear_combinarion_alpha` using the same coeffs and input lead to different results.

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,487 @@
...
+                one,
+                alpha: None,
+                next_alpha: one,
+            },
+            (0..n_inputs).map(NodeIndex::from).collect(),
+        )
+    }
+
+    // Append an operation to the circuit. A new operation is added only if
+    // syntactically different from any operations already added to the circuit
```
Al-Kindi-0, line 218: nit: "... if it is syntactically different from any operation already added ..."

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+/// This is equivalent to the check $$\frac{\mathsf{num}_0 }{\frac{z^n - 1}{z - g^{-1}}} + \frac{\mathsf{num}_1}{z-1} + \frac{\mathsf{num}_2}{z - g^{-1}} = Q(z)$$.
+///
+/// The circuit expects as inputs:
+/// - auxiliary randomness of the AirScript e.g. `random_values { rand: [2] }`
+/// - the public inputs of the AirScript e.g. `public_inputs { stack_inputs[16] }`
+/// - the main segment of trace inputs of the AirScript e.g. `trace_columns { main: [a b] }`
+/// - the aux segment of trace inputs of the AirScript e.g. `trace_columns { aux: [f] }`
+/// - the next row main segment of trace inputs of the AirScript e.g. `a' b'`
+/// - the next row aux segment of trace inputs of the AirScript e.g. `f'`
+/// - qi(z) with i in [0,7)
```
Al-Kindi-0, line 34: the range should be [0, 8)

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+/// - alpha
+/// - z
+/// - z^n
+/// - g^(-1)
+/// - z^min_num_cycles
```
Al-Kindi-0, line 39: nit: maybe the formula computing min_num_cycles should be added

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+/// - auxiliary randomness of the AirScript e.g. `random_values { rand: [2] }`
+/// - the public inputs of the AirScript e.g. `public_inputs { stack_inputs[16] }`
+/// - the main segment of trace inputs of the AirScript e.g. `trace_columns { main: [a b] }`
+/// - the aux segment of trace inputs of the AirScript e.g. `trace_columns { aux: [f] }`
+/// - the next row main segment of trace inputs of the AirScript e.g. `a' b'`
+/// - the next row aux segment of trace inputs of the AirScript e.g. `f'`
+/// - qi(z) with i in [0,7)
+/// - alpha
+/// - z
+/// - z^n
```
Al-Kindi-0, line 37: nit: maybe mention that n is the execution trace length

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+        .clone()
+        .into_iter()
+        .chain(boundary_first_roots.clone())
+        .chain(boundary_last_roots.clone())
+        .collect();
+    let (periodic_ops, non_periodic_ops): (Vec<(NodeIndex, Operation)>, _) =
+        UniqueOperationIterator::new(ir.constraint_graph(), all_roots)
+            .partition(|(_, op)| matches!(op, Operation::Value(Value::PeriodicColumn(_))));
+
+    // A circuit builder is intantiated with the inputs of the circuits plus the 13 needed by the ACE chiplet
```
Al-Kindi-0, line 76: typo: instantiated

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+    let n_inputs = n_random(ir) + n_public_inputs(ir) + n_trace_accesses(ir);
+    let (mut cb, inputs) = CircuitBuilder::new(n_inputs + 13);
```
Al-Kindi-0, line 78: Maybe we should use a constant and define how it is computed?

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+    let qs: Vec<NodeIndex> = inputs[n_inputs..n_inputs + 8].to_vec();
+    let alpha = inputs[n_inputs + 8];
+    cb.alpha = Some(alpha);
+    let z_point = inputs[n_inputs + 9];
+    let z_n = inputs[n_inputs + 10];
+    let inverse_generator = inputs[n_inputs + 11];
+    let z_min_num_cycles = inputs[n_inputs + 12];
```
Al-Kindi-0, line 85: Same comment regarding constants here

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    // A mapping is necessary to convert the indexes of the Air to the new indexes of the circuit
+    let mut mapping: BTreeMap<NodeIndex, NodeIndex> = BTreeMap::new();
+
+    periodic_columns(ir, &mut cb, &mut mapping, periodic_ops, z_min_num_cycles);
```
Al-Kindi-0, line 90: nit: could we add a comment here?

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    // Inputs are filtered and their indexes are converted. They are not added
+    // to the circuit because, unlike in the algebraicgraph where inputs are
+    // nodes in the graph, in the ACE circuit inputs are inplicit. They are
```
Al-Kindi-0, line 94: implicit?

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+    // simply the first indexes.
+    // Their converted indexes however need to be added to the mapping so that
+    // any nodes using an index can be remapped correctly.
```
Al-Kindi-0, line 97: nit: " ... any nodes using such an index ..."

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+    // remapped and added to the circuit.
+    for (out, op) in non_input_ops {
+        let op = remap_op(0, op, &mapping); // In the algebraicgraph indexes start at 0
+        let new_out = cb.push(op);
+        mapping.insert(out, new_out);
+    }
+
+    // The original Airscript is not embedded inside the ACE circuit.
+    // We start adding new nodes to join the Airscript roots in the formula
+    // described above.
```
Al-Kindi-0, line 122: Could you reformulate/explain more here?

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+            .map(|(identifier, col)| {
+                let mut column: Vec<Felt> = col.values.iter().map(|el| Felt::new(*el)).collect();
+                let inv_twiddles = winter_math::fft::get_inv_twiddles::<Felt>(column.len());
+                winter_math::fft::interpolate_poly(&mut column, &inv_twiddles);
+                let column: Vec<u64> = column.into_iter().map(From::from).collect();
+                (*identifier, column)
+            })
+            .collect();
+
+        // The polynomials are stored in the circuit as constants
```
Al-Kindi-0, line 295: nit: The coefficients of the interpolated polynomials

codegen/ace/src/serialization.rs
```diff
@@ -0,0 +1,178 @@
...
+/// recovered from its felt serialization.
+/// The format starts with a header of 3 field elements containing:
+/// - the number of inputs,
+/// - the number of constants and
+/// - the number of evaluation gates.
+///
+/// In the case of a gate, the 63 bits available in a field element are used in
+/// the following order, from least significant:
+/// - 20 bits for right index
+/// - 20 bits for left index
```
Al-Kindi-0, line 81: I think we now use 30 bits each, @adr1anh ?

codegen/ace/src/serialization.rs
```diff
@@ -0,0 +1,178 @@
...
+/// ```text
+/// [
+/// // begin header
+/// 1, // one input
+/// 1, // one constant
+/// 2, // two gates
+/// // end header
+/// 14, // value of the constant, which has index 1, given that there is one input at index 0
+/// (2 << 40) + (0 << 20) + (1 << 0), // addition gate opcode with left intput 0 and right input 1. Output index 2.
+/// (1 << 40) + (2 << 20) + (0 << 0), // multiplication gate opcode with left intput 2 and right input 0. Output index 3.
```
Al-Kindi-0, line 98: same here

Al-Kindi-0 (changes requested):
Looks great!

There are a few optimizations improvement that we can do, but I thin we should just try to get this merged asap. This way we can progress the work on the VM side while at the same time open PRs here and elsewhere in order to clean up and improve the codegen pipeline.

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,487 @@
...
+            None
+        } else {
+            let i: usize = NodeIndex::into(i);
+            Some(self.ops[i - self.n_inputs])
+        }
+    }
+
+    pub fn to_dot(&self) -> Result<String, std::fmt::Error> {
+        operations_to_dot(&self.ops)
+    }
```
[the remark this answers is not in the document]
paracetamolo: done

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,487 @@
...
+
+    /// Serialize the circuit to a vector of field elements, as expected by the
```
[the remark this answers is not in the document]
paracetamolo: done

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,487 @@
...
+/// - inputs are preserved even if unused
+pub struct UniqueOperationIterator<'a> {
+    // circuit being iterated on
+    circuit: &'a Circuit,
+    // nodes visited but not necessarily processed yet
+    visited: BTreeSet<NodeIndex>,
+    // nodes already processed, meaning their children are visited
+    processed: BTreeSet<NodeIndex>,
+    // nodes to be processed
+    process: Vec<NodeIndex>,
```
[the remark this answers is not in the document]
paracetamolo: done

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,487 @@
...
+                    }
+                }
+            }
+        }
+        None
+    }
+}
+
+/// Remap the children of an operation. Typically when migrating an operation
+/// between data structures. Any index withtin the inputs is left untouched.
```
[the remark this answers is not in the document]
paracetamolo: done

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,487 @@
...
+}
+
+#[derive(Default)]
+pub struct CircuitBuilder {
+    // The ACE circuit being built
+    pub circuit: Circuit,
+    // A cache of nodes already inserted in the circuit, used to avoid duplicates
+    cache: BTreeMap<Operation, NodeIndex>,
+    // Constants zero and one are inserted first so that they can be used as
+    // identity and absorbing element in addition and multiplication.
```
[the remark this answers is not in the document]
paracetamolo: done

codegen/ace/src/codegen.rs
paracetamolo: done

codegen/ace/src/codegen.rs
paracetamolo: It's a very confusing typo, it was supposed to be "is noW embedded". I rephrased it.

codegen/ace/src/codegen.rs
paracetamolo: done

codegen/ace/src/serialization.rs
paracetamolo: It's not the comment, it's the code that was outdated, fixed it now.

codegen/ace/src/serialization.rs
paracetamolo: done

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    // The original Airscript is not embedded inside the ACE circuit.
+    // We start adding new nodes to join the Airscript roots in the formula
+    // described above.
+
+    let z_minus_g = cb.sub(z_point, inverse_generator);
+    let z_minus_one = cb.sub(z_point, cb.one);
+    let z_n_minus_one = cb.sub(z_n, cb.one);
+
+    let lhs = {
```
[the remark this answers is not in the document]
paracetamolo: I created a separate struct `linear_combination_alpha`

paracetamolo:
Update:
- addressed all the comments
- moved public inputs before random
- renamed all regression files with the more informative script name rather than a number e.g. 1.dot -> ComplexBoundary.dot

adr1anh:
Looks good to me!

codegen/ace/src/circuit.rs
```diff
@@ -0,0 +1,457 @@
+use crate::comparison::compare_operation;
+use crate::serialization::{operations_to_dot, operations_to_felts, to_quad, Quad};
+use air_ir::{NodeIndex, Operation, Value};
+use std::cmp::Ordering;
+use std::collections::{btree_map::BTreeMap, btree_set::BTreeSet};
+
+/// A circuit that can be consumed by the ACE chiplet.
+/// The only way to build a circuit is through the CircuitBuilder, it can then obtained
```
Al-Kindi-0, line 8: "can then be obtained" ?

codegen/ace/src/codegen.rs
Al-Kindi-0: @paracetamolo , could we fix this?

codegen/ace/src/codegen.rs
Al-Kindi-0: @paracetamolo , could we fix this?

codegen/ace/src/codegen.rs
Al-Kindi-0: @paracetamolo , could we fix this?

codegen/ace/src/codegen.rs
Al-Kindi-0: @paracetamolo , could we fix this?

codegen/ace/src/codegen.rs
Al-Kindi-0: @paracetamolo , could we fix this?

codegen/ace/src/codegen.rs
Al-Kindi-0: @paracetamolo , could we fix this?

codegen/ace/src/codegen.rs
Al-Kindi-0: @paracetamolo , could we fix this?

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,406 @@
...
+                        };
+                    }
+                };
+            }
+        }
+        None
+    }
+}
+
+// Helper functions to convert the Air inputs to ACE inputs.
```
Al-Kindi-0, line 249: Should we a section separator as we done in other repositories?

Al-Kindi-0 (changes requested):
Left some new and old comments, once these are addressed we can merge in say a `ace-codegen` branch.

codegen/ace/src/codegen.rs
Al-Kindi-0: nit: could we add a comment here ?

codegen/ace/src/serialization.rs
paracetamolo: done

codegen/ace/src/codegen.rs
```diff
@@ -0,0 +1,406 @@
...
+/// polynomial P. The polynomial P is then evaluated at `z l` where
+/// `l := trace_len / cycle_len`.
+/// However to avoid having one input per periodic column, a single power of z is
+/// received as input and the others are computed.
+/// Let `max_cycle_len` be the biggest cycle length appearing in the periodic columns
+/// of the circuit and let `min_num_cycles = trace_len / max_cycle_len`.
+/// The circuit gets as input `z^min_num_cycles`.
+/// For a given cycle_len let `k = max_cycle_len / cycle_len`, then `z^l`, where
+/// `l := trace_len / cycle_len`, is nothing but `(z^min_num_cycles)^k`.
+fn periodic_columns(
```
adr1anh, line 305: It might be clearer to factor out a method for computing the node of a single periodic column, wrapping it in a loop that computes all cycles. This can be done in a future refactor PR.

```rust
        // Maximum length of all periodic columns in the air. 
        let max_cycle_len = self
            .air
            .periodic_columns
            .values()
            .map(|pi| pi.values.len())
            .max()
            .unwrap();

        // The power of `z` for the longest column. Let `l` such that `z_min_cycle = z^l`, 
        // such that `l = trace_len / max_cycle_len`
        let z_min_cycle = Node::Input(self.layout.air_input_index(Z_MIN_CYCLE));
        
        let cycle_len = periodic_column.values.len();
        assert!(max_cycle_len.is_power_of_two());
        assert!(cycle_len.is_power_of_two());
        let log_k = (max_cycle_len / cycle_len).ilog2();
        let z_pow_l = (0..log_k).fold(z_min_cycle, |acc, _| self.square(acc));

        let mut column: Vec<_> = periodic_column
            .values
            .iter()
            .copied()
            .map(Felt::new)
            .collect();
        let inv_twiddles = winter_math::fft::get_inv_twiddles::<Felt>(column.len());
        winter_math::fft::interpolate_poly(&mut column, &inv_twiddles);

        let poly: Vec<_> = column
            .into_iter()
            .map(|c| self.get_constant_node(u64::from(c)))
            .collect();

        let result = self.horner(z_pow_l, poly.into_iter().rev());
```

codegen/ace/src/codegen.rs
paracetamolo: done

codegen/ace/src/codegen.rs
paracetamolo: done

codegen/ace/src/codegen.rs
paracetamolo: done

codegen/ace/src/comparison.rs
paracetamolo: I don't want lexicographic ordering, I want values to be first and the rest untouched. The sorting algorithm is stable so this should work.

paracetamolo (commented):
Fixed some old comments that I had missed.

codegen/ace/src/circuit.rs
paracetamolo: The only way to obtain an object of type Circuit is from the normalize function of CircuitBuilder. I'm trying to explain the flow of usage. I can rephrase if it is confusing.

Al-Kindi-0: the comment was only on a typo in "obtained" which is now fixed, thank you

codegen/ace/src/codegen.rs
Al-Kindi-0: Maybe we should create an issue on this?

PR #371, opened 2025-03-07
Feat: implement buses

Leo-Besancon:
PR implementing https://github.com/0xPolygonMiden/air-script/issues/183, targetting upstream next (instead of https://github.com/massalabs/air-script/pull/8 on massalabs's fork).

The goal of this PR is to make an initial implementation of buses, that can be declared:
```
buses {
    multiset p,
    logup q,
}
```

Currently, only empty bus boundary constraints are handled (that are reduced to a `0` or `1` value depending on the bus type)
```
boundary_constraints {
    enf p.first = null;
    enf p.last = null;
}
```
Boundary constraints with public inputs depend on https://github.com/0xPolygonMiden/air-script/issues/143 and will be handled in a future PR.

```
integrity_constraints {
    p.insert(a) when s1;
    p.remove(a, b) when 1 - s2;
}
```

What was done in this PR:
- [x] Parse bus related expressions (AST)
- [x] Handle semantic analysis and constant propagation for buses (AST)
- [x] In a new MIR > MIR pass, before lowering to the AIR, expand bus integrity constraints and handle empty bus boundary constraints
- [x] Add tests (in parser, Mir, Air, and Codegen) for new bus expressions
- [x] Add documentation on buses

In future PR, we will introduce, in order of priority:
- [ ] Update the bus constraint handling (MIR/AIR > CodeGen)
  - [ ] In the AIR, add bus structs with an exhaustive description
  - [ ] Prepare expanding the constraints in CodeGen
- [ ] Handle public inputs as Bus constraints
  - [ ] Needs variable-length public inputs support, see https://github.com/0xPolygonMiden/air-script/issues/143
  - [ ] Needs to store the Bus constraint as a separate struct in the AIR (see above)
  - [ ] Needs Codegen update to generate the resulting constraint
- [ ] Remove aux and random value access in Front-End (see branch https://github.com/massalabs/air-script/tree/refactor-remove-aux-rand)


air-script/tests/buses/buses_complex.rs
```diff
@@ -0,0 +1,96 @@
...
+        let aux_current = aux_frame.current();
+        let aux_next = aux_frame.next();
+        result[0] = (E::ONE + (aux_rand_elements.get_segment_elements(0)[0] + E::ONE * aux_rand_elements.get_segment_elements(0)[1] + E::from(main_current[0]) * aux_rand_elements.get_segment_elements(0)[2]) * E::from(main_current[2]) + E::ONE - E::from(main_current[2]) + (aux_rand_elements.get_segment_elements(0)[0] + E::from(2_u64) * aux_rand_elements.get_segment_elements(0)[1] + E::from(main_current[1]) * aux_rand_elements.get_segment_elements(0)[2]) * (E::ONE - E::from(main_current[2])) + E::ONE - (E::ONE - E::from(main_current[2]))) * aux_current[0] - (E::ONE + (aux_rand_elements.get_segment_elements(0)[0] + E::ONE * aux_rand_elements.get_segment_elements(0)[1] + E::from(main_current[1]) * aux_rand_elements.get_segment_elements(0)[2]) * E::from(main_current[3]) + E::ONE - E::from(main_current[3]) + (aux_rand_elements.get_segment_elements(0)[0] + E::from(2_u64) * aux_rand_elements.get_segment_elements(0)[1] + E::from(main_current[0]) * aux_rand_elements.get_segment_elements(0)[2]) * (E::ONE - E::from(main_current[3])) + E::ONE - (E::ONE - E::from(main_current[3]))) * aux_next[0];
```
bobbinth, line 93: Assuming I've decoded this correctly, I think this is not exactly correct. Denoting:

$$
r_0 = \alpha_0 + \alpha_1 + \alpha_2 \cdot a \
$$

$$
r_1 = \alpha_0 + \alpha_1 + \alpha_2 \cdot b \
$$

$$
r_2 = \alpha_0 + \alpha_1 + 2 \cdot \alpha_2 \cdot b \
$$

$$
r_3 = \alpha_0 + \alpha_1 + 2 \cdot \alpha_2 \cdot a \
$$

What we have here is:

$$
p \cdot (1 + r_0 \cdot s_1 + 1 - s_1 + r_2 \cdot (1 - s_1) + 1 - (1 - s_1)) - p' \cdot (1 + r_1 \cdot s_2 + r_3 \cdot (1 - s_2) + 1 - (1 - s_2))
$$

But what we want is:

$$
p \cdot ((1 + r_0 \cdot s_1 - s_1) \cdot (1 + r_2 \cdot (1 - s_1) - (1 - s_1)) - p' \cdot ((1 + r_1 \cdot s_2 - s_2) \cdot (1 + r_3 \cdot (1 - s_2) - (1 - s_2))
$$

Eventually (not in this PR), this will get simplified to something like:

$$
p \cdot ((r_0 \cdot s_1 - s_1 + 1) \cdot (r_2 \cdot (1 - s_1) + s_1)) - p' \cdot ((r_1 \cdot s_2 - s_2 + 1) \cdot (r_3 \cdot (1 - s_2) + s_2))
$$

@Al-Kindi-0, @paracetamolo, @adr1anh could you double-check my logic? This output corresponds to [these lines](https://github.com/0xPolygonMiden/air-script/blob/4e89655202782615c697ea46139643d34e0e3e5e/air-script/tests/buses/buses_complex.air#L30-L34).

air-script/tests/buses/buses_complex.rs
```diff
@@ -0,0 +1,96 @@
...
....get_segment_elements(0)[0] + E::from(3_u64) * aux_rand_elements.get_segment_elements(0)[1] + E::from(main_current[0]) * aux_rand_elements.get_segment_elements(0)[2]) * E::from(main_current[2]) - (E::ONE * (aux_rand_elements.get_segment_elements(0)[0] + E::from(3_u64) * aux_rand_elements.get_segment_elements(0)[1] + E::from(main_current[0]) * aux_rand_elements.get_segment_elements(0)[2]) * (aux_rand_elements.get_segment_elements(0)[0] + E::from(3_u64) * aux_rand_elements.get_segment_elements(0)[1] + E::from(main_current[0]) * aux_rand_elements.get_segment_elements(0)[2]) * (aux_rand_elements.get_segment_elements(0)[0] + E::from(3_u64) * aux_rand_elements.get_segment_elements(0)[1] + E::from(main_current[0]) * aux_rand_elements.get_segment_elements(0)[2]) * aux_next[1] + E::ZERO + E::ONE * (aux_rand_elements.get_segment_elements(0)[0] + E::from(3_u64) * aux_rand_elements.get_segment_elements(0)[1] + E::from(main_current[0]) * aux_rand_elements.get_segment_elements(0)[2]) * (aux_rand_elements.get_segment_elements(0)[0] + E::from(3_u64) * aux_rand_elements.get_segment_elements(0)[1] + E::from(main_current[0]) * aux_rand_elements.get_segment_elements(0)[2]) * E::from(main_current[4]));
```
bobbinth, line 94: I haven't checked this output, but it should correspond to [these lines](https://github.com/0xPolygonMiden/air-script/blob/4e89655202782615c697ea46139643d34e0e3e5e/air-script/tests/buses/buses_complex.air#L36-L38). @Al-Kindi-0, @paracetamolo, @adr1anh - could you double-check it?

air-script/tests/buses/buses_simple.air
```diff
@@ -0,0 +1,32 @@
...
+    #enf p.last = inputs;
+    #enf q.last = inputs;
+}
+
+integrity_constraints {
+    p.add(1) when 1;
+    p.rem(1) when 1;
+    q.add(1, 2) when 1;
+    q.add(1, 2) when 1;
+    q.rem(1, 2) for 2;
```
bobbinth, line 31: I'm actually not sure to what extent this is a helpful test case. Using constants in `when` expressions is probably not that indicative of potential issues. I'd probably use something like:
```
p.add(1) when a;
p.rem(1) when (1 - a);
q.add(1, 2) when a;
q.rem(1, 2) for 2;
```

bobbinth (commented):
Looks great! Thank you! I mostly reviewed the `air`, `docs`, and tests and left some comments inline.

docs/src/description/buses.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Bus types
+
+## Multiset `unit`
+
+Multiset-based buses can represent constraints specifying given values have been added or removed from a column, in no specific order.
+
+## LogUp `mult`
+
+LogUp-based buses are more complex than multiset buses, and can encode the multiplicity of an element: an element can be added or removed multiple times.
```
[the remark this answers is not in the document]
adr1anh: I recall some folks referring to [BG15](https://eprint.iacr.org/2015/195.pdf) as the paper introducing multi-set via grand-product arguments, though it's very disconnected from it's usage in arithmetizations and doesn't really explain the concept. Personally, I found the best resource was the Miden docs. 

Regarding naming, I had thought of `chan` (for channel) for `unit` because it reminds me of the same concept in Go, where values are "sent" and "received" exactly once. In fact, it matches with the non-deterministic nature of STARKs, since a channel in Go can start receiving and block, waiting for the message to be sent later on. If the messages don't match up, then the program halts.

For the logUp bus, I’m not sure what to call the bus itself, but I'd think of the following actions as representing the access with multiplicity +1/-m
- read/write: modeling write-once memory
- get/insert: modeling append-only map
- receive/broadcast: modeling a multi-sender single receiver channel
- require/provide: modeling statements which can be required for the current row to be correct, but can be proved elsewhere.

air-script/tests/buses/buses_complex.rs
adr1anh: Yes, it does look like the current output is incorrect. It feels like this may be an optimization relating to the fact that $$s_i, 1-s_i$$ are mutually exclusive so it could be written as 
$$p \cdot (1 + (r_0 - 1)\cdot s_1 + (r_2 - 1) \cdot (1-s_1)) - p' \cdot (1 + (r_1 - 1)\cdot s_2 + (r_3 - 1) \cdot (1-s_1))$$

air-script/tests/buses/buses_complex.rs
adr1anh: Seems to be the same issue as the above comment.

adr1anh:
Perhaps this was discussed earlier, but I found the terms `add` and `rem` a bit confusing in the context of a logUp `mult` column. I think these should be reserved for operations which respectively add or subtract a tuple from the bus, but only when the `when` condition is binary (not sure though if this is enforceable or if we would just make it a convention). 

I think having an alternative `clear` operator would make it clearer that we are removing tuples as a batch, using a given multiplicity. It could then also be combined with a `when` clause indicating that the flushing only occurs under some conditions. 

I'm thinking it would look something like this. If we take `t` as a tuple of elements and a binary flag `f` then
- `p.add(t) when f` should enforce $$p' = p + \frac{f}{t}$$
- `p.rem(t) when f` should enforce $$p' = p - \frac{f}{t}$$
- `p.clear(t, m) when f` enforces $$p' = p - \frac{f \cdot m}{t}$$, where the `when` is optional, and `m` should represent the total number of times the element is removed from the bus. We could also do $$p' = p - \frac{m}{t}$$ and add a constraint that `f => m==0`.

Perhaps another useful side effect would be that we can avoid specifying what kind of bus we want to use. When a bus uses the `clear` operator, we have to use logUp. If only `add` and `rem` are used, then the optimizer could choose either kind, potentially reusing a column for two different bus types.

air-script/tests/buses/buses_complex.rs
Al-Kindi-0: @bobbinth , the expression above corresponds to what you wrote but with a missing $1 - s_2$ term i.e., it should be
$$
p \cdot (1 + r_0 \cdot s_1 + 1 - s_1 + r_2 \cdot (1 - s_1) + 1 - (1 - s_1)) - p' \cdot (1 + r_1 \cdot s_2 + 1 - s_2 + r_3 \cdot (1 - s_2) + 1 - (1 - s_2))
$$

but what we want to compute is indeed what you wrote. Thus it seems that the strategy for combining multiple (grand-product-based) bus requests/responses is currently addition but it should instead be multiplication.

air-script/tests/buses/buses_complex.rs
Al-Kindi-0: Same issue is here in addition to the selectors not being used correctly i.e., the pattern is $1 + r \cdot s - s$. Also the gating with the column $d$ is not translated correctly and the whole bus constraint is not following the pattern $q \cdot (\cdots) = q^{'}  \cdot (\cdots)$

docs/src/description/buses.md
```diff
@@ -0,0 +1,61 @@
...
+The above example states that the bus `p` should be empty at the beginning and end of the trace.
+
+## Bus integrity constraints
+
+In the integrity constraints section, we can add and remove elements (as tuples of felts) to and from a bus. In the following examples, `p` and `q` are respectivelly multiset and LogUp based buses.
+
+```
+integrity_constraints {
+    p.add(a) when s1;
+    p.rem(a, b) when 1 - s2;
```
bitwalker, line 44: I feel like `add` and (particularly) `rem` visually just look too much like their usual mathematical operations, rather than something more set-oriented. Maybe that's just my misconception in terms of understanding what is implied by these operations, but Is there any reason not to use `insert`/`append` and `remove`/`delete` for these?

Generally speaking, I really prefer _not_ abbreviating things unless it is absolutely unambiguous what the name refers to. In this case, I don't think that's true, and more explicit language is likely to make the resulting AirScript more readable. I feel the same way about `enf` versus `enforce`, but never got around to changing that one myself.

Anyway, thoughts? /cc @bobbinth

parser/src/ast/declarations.rs
```diff
@@ -76,6 +78,56 @@ pub enum Declaration {
...
+        }
+    }
+}
+#[derive(Default, Hash, Debug, Clone, PartialEq, Eq)]
+pub enum BusType {
+    /// A multiset bus
+    #[default]
+    Unit,
+    /// A logup bus
+    Mult,
```
bitwalker, line 105: I find it a tad bit confusing that the "multiset" bus type is called `Unit`, while `Mult` refers to the LogUp-style bus. IIUC, `Mult` here refers to "multiplicative", in contrast to a non-multiplicative bus, hence "unit".

I think this is a case again where abbreviating things just obscures the meaning too much.

parser/src/ast/display.rs
```diff
@@ -111,6 +111,7 @@ impl fmt::Display for DisplayStatement<'_> {
                 write!(f, "enf {}", expr)
             }
             Statement::Expr(ref expr) => write!(f, "return {}", expr),
+            Statement::BusEnforce(ref expr) => write!(f, "bus_enf {}", expr),
```
bitwalker, line 114: For display, this should probably be `enf` as well, just like `EnforceAll`, no? There is no `bus_enf` keyword in any case. For `Debug` purposes though, obviously you'd want to distinguish the two, but `Display` is usually used for pretty printing, so it should probably follow the actual AirScript syntax as much as possible.

parser/src/ast/expression.rs
```diff
@@ -335,6 +340,7 @@ impl Expr {
             Self::Call(ref call) => call.ty,
             Self::ListComprehension(ref lc) => lc.ty,
             Self::Let(ref let_expr) => let_expr.ty(),
+            Self::BusOperation(_) | Self::Null(_) => todo!(),
```
bitwalker, line 343: *Question:* Is this planned for a future PR?

parser/src/lexer/mod.rs
```diff
@@ -578,6 +605,7 @@ where
         let next = self.read();
         match Token::from_keyword_or_ident(self.slice()) {
             Token::Ident(id) if next == '(' => Token::FunctionIdent(id),
+            //Token::Ident(id) if next == '.' => Token::BusIdent(id),
```
bitwalker, line 608: *Nit:* Looks like this got left in by mistake

parser/src/parser/grammar.lalrpop
```diff
@@ -180,6 +181,26 @@ PeriodicColumn: PeriodicColumn = {
         => PeriodicColumn::new(span!(l, r), name, values),
 }
 
+
+// BUSES
+// ================================================================================================
+
+// Buses are not required, and there is no limit to the number that can be provided.
+Buses: Span<Vec<Bus>> = {
+    <l:@L> "buses" "{" <bus:Bus*> "}" <r:@R>
```
bitwalker, line 190: *Question:* Would an empty `buses` block ever make sense? I'd expect either the block to be non-empty, or to be removed altogether.

Probably doesn't matter one way or the other in this case, but quite often with parser generators it can make the grammar ambiguous, so modifying the syntax in this block in the future could get hairy if doing so introduces an ambiguity unintentionally.

parser/src/sema/semantic_analysis.rs
bitwalker, line 1: Looks like there are a couple places where the `ControlFlow` result of a function is being ignored, which will result in those errors being swallowed rather than propagated.

bitwalker (changes requested):
Overall, looks great! I have a few nits/questions I'd like to address before signing off, but it's small stuff. Nice work!

docs/src/description/buses.md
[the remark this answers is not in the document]
bobbinth: > Personally, I found the best resource was the Miden docs.

Maybe then we can link [here](https://0xpolygonmiden.github.io/miden-vm/design/lookups/multiset.html) for the multiset busses and to [here](https://0xpolygonmiden.github.io/miden-vm/design/lookups/logup.html) for LogUp busses?

parser/src/parser/grammar.lalrpop
bobbinth: > Would an empty `buses` block ever make sense? I'd expect either the block to be non-empty, or to be removed altogether.

Yeah, an empty `busses` block is not really useful. So, if we can force that if a `busses` block is defined it always defines at least one bus, that would be best.

bobbinth:
> I find it a tad bit confusing that the "multiset" bus type is called `Unit`, while `Mult` refers to the LogUp-style bus. IIUC, `Mult` here refers to "multiplicative", in contrast to a non-multiplicative bus, hence "unit".

> Regarding naming, I had thought of `chan` (for channel) for `unit` because it reminds me of the same concept in Go, where values are "sent" and "received" exactly once. In fact, it matches with the non-deterministic nature of STARKs, since a channel in Go can start receiving and block, waiting for the message to be sent later on. If the messages don't match up, then the program halts.

Agree with the comments above - maybe we don't need to abbreviate too much. Maybe we could use:

- `multiset` for simple multiset busses.
- `logup` for LogUp-based busses.

This does expose the underlying methodology, which may be a bad or a good thing.

If we did want to hide the methodology, we could use `channel` for simple multiset busses as @adr1anh suggested, but then I'm also not sure what to use for LogUp-based busses.

> I feel like `add` and (particularly) `rem` visually just look too much like their usual mathematical operations, rather than something more set-oriented. Maybe that's just my misconception in terms of understanding what is implied by these operations, but Is there any reason not to use `insert`/`append` and `remove`/`delete` for these?

I agree that `add` and `rem` are probably not as clear. I'm fine with `insert`/`remove` or maybe `send`/`receive`.

> Perhaps this was discussed earlier, but I found the terms `add` and `rem` a bit confusing in the context of a logUp `mult` column. I think these should be reserved for operations which respectively add or subtract a tuple from the bus, but only when the `when` condition is binary (not sure though if this is enforceable or if we would just make it a convention).
> 
> I think having an alternative `clear` operator would make it clearer that we are removing tuples as a batch, using a given multiplicity. It could then also be combined with a `when` clause indicating that the flushing only occurs under some conditions.

I am thinking about LogUp busses a bit differently. This is described in more detail in https://github.com/0xPolygonMiden/air-script/issues/183#issuecomment-2543024385, but to summarize, here's how it would work:

```
q.add(a, b, c) for d
q.rem(e, f, g) when s
```
These result in the following algebraic expression:

$$
q' + \frac{s}{\alpha_0 + \alpha_1 \cdot e + \alpha_2 \cdot f + \alpha_3 \cdot g} = q + \frac{d}{\alpha_0 + \alpha_1 \cdot a + \alpha_2 \cdot b + \alpha_3 \cdot c}
$$

From the algebraic standpoint, there is really no difference if we use `when` or `for` (the expression after this keyword ends up being in a numerator), but this does allow us to specify multiplicity for both "addition" and "removal" operations.

Al-Kindi-0:
(My two cents)

I think that we want to support two primitives:
1. Multi-set checks
2. Look-ups

Lookups are a stronger primitive than multi-set checks (i.e., lookups can be used to implement multi-set checks) while lookups can use multi-set checks underneath (e.g., Lasso) or they might not (e.g., LogUp).

If we forget for a second about LogUp-GKR, as this was the setting where we had to use LogUp for both lookups and multi-set checks, then

1. Multi-set checks should probably always be implemented using grand-products instead of grand-sums[^1], i.e., $b^{'} \cdot (1 + f_0 \cdot s_0 + (1 - s_0)) = b \cdot (1 + f_1 \cdot s_1 + (1 - s_1))$ instead of $b^{'} + \frac{s_0}{f_0}  = b + \frac{s_1}{f_1}$, as they are more degree-friendly with respect to selectors.
2. For lookups, we can either expose a primitive called lookups but this will probably be more complicated (e.g., the multiplicity column `m` should be opaque to the user) or we have to enshrine LogUp in the language and hence also the concept of multiplicity. The implies that for lookups we will be using grand-sums (of fractions) and that lookups will be distinguished from multi-set checks with the existence of a multiplicty column.

The above suggests the following syntax

For multi-set checks
```
p.add(a, b) when s
p.rem(c, d) when (1 - s)
```

For lookups (using LogUp)

```
p.add(a, b) when s
p.rem(c, d) with m
```

So lookups are distinguished with the `with` word and the multiplicity column `m` which can be made of arbitrary field elements (e.g., not necessarily a binary flag). On the other hand, `when` is always used with binary selectors. Note that selectors are always binary flags, and even in the case of LogUp-GKR numerators they are always binary flags except for numerators specifying multiplicities.

On another note, I think we can also use `push/pull` to signify the same as `add/rem`. The advantage of the former is that it is short and hence doesn't need to be shortened and it also has the same number of letters for both options.


[^1]: Not sure if this latter terminology is used elsewhere but it seems a reasonable one.

Al-Kindi-0:
As for a reference with the idea of multi-set checks, I think it will be difficult (and potentially unfair) to pin-point a single source on this idea. But if I had to choose one, I would go with [this](https://people.csail.mit.edu/devadas/pubs/mhashes.pdf).

The reason is that this reference really distills the core idea of all multi-set checks we are aware of, namely to check that two multi-sets are equal (e.g, the multi-set of `SEND`/`PUSH`-s is equal to the multi-set of `RECEIVE`/`PULL`-s), we use a primitive called a (incremental) multi-set hash to hash both sets and compare the digests for equality at the end. 

The multi-set hash could be the grand-product one, the grand-sum one, the elliptic-curve one or even a lattice-base one, each with its own set of tradeoffs.

air-script/tests/buses/buses_complex.rs
Leo-Besancon: Hello @bobbinth, @adr1anh @Al-Kindi-0 , thank you!
For multiset check, I indeed seem to have used Add instead of Mul to combine the different factors, it should be fixed in https://github.com/0xPolygonMiden/air-script/pull/371/commits/aa1e68d57a6e73c8c2602a3d66bca67525410dfd
I updated the related generated files.

air-script/tests/buses/buses_complex.rs
Leo-Besancon: For LogUp buses, once we removed all divisions and transformed them into multiplications, I think we have the correct formula.
I updated the test to better understand what is removed and added:
```
q.add(3, a) when s1;
q.add(3, a) when s1;
q.rem(4, b) for d;
```

Then, the equivalent `result[1]` constraint is the following (equation based on https://github.com/0xPolygonMiden/air-script/pull/371#issuecomment-2752875785):
```
// This corresponds to the denominator for adding (3, a)
let denom_3_a = (aux_rand_elements.get_segment_elements(0)[0]
+ E::from(3_u64) * aux_rand_elements.get_segment_elements(0)[1]
+ E::from(main_current[0]) * aux_rand_elements.get_segment_elements(0)[2]);

// This corresponds to the denominator for removing (4, b)
let denom_4_b = (aux_rand_elements.get_segment_elements(0)[0]
+ E::from(4_u64) * aux_rand_elements.get_segment_elements(0)[1]
+ E::from(main_current[1]) * aux_rand_elements.get_segment_elements(0)[2]);

// We multiply the whole equation corresponding to this comment: https://github.com/0xPolygonMiden/air-script/pull/371#issuecomment-2752875785 by the product of all denominators to get rid of divisions
let prod = denom_3_a * denom_3_a * denom_4_b;

let s1 = E::from(main_current[2]);
let d = E::from(main_current[4]);

result[1] = prod * aux_current[1]
    + denom_3_a * denom_4_b * s1
    + denom_3_a * denom_4_b * s1
    - (
        prod * aux_next[1]
        + denom_3_a * denom_3_a * d
    );
```

air-script/tests/buses/buses_simple.air
Leo-Besancon: Thank you, I updated the buses_simple test in https://github.com/0xPolygonMiden/air-script/pull/371/commits/230e1b1e4e04afa716f11ef1ae752360223b25c8

air-script/tests/buses/buses_complex.rs
Al-Kindi-0: Apologies Leo, what you wrote makes sense now as I didn't take into account the fact that this is a LogUp column.
A couple of points though:

1. It seems that the example in the code has `q.rem(3, b) for d` instead of `q.rem(4, b) for d`. I think `q.rem(4, b) for d` is a better test, as you used in your comment above.
2. There are a couple of  additions with `E::ZERO`, could we get rid of these as they bloat the code? Same with `E::ONE`, if it is not too much hassle.

Leo-Besancon: No worries! Sure, I've updated the test in https://github.com/0xPolygonMiden/air-script/pull/371/commits/a0d5ff7ba8c792bad292af0ca984cd7b7fe0a118.

For the `+ E::ZERO` and `* E::ONE`, indeed it would be best to track why they are generated and if we can remove them easily, we'll dig into it.

Al-Kindi-0 (commented):
Thank you Leo!
Not a deep review from my side but left a few comments.

air-script/tests/buses/buses_complex.rs
Leo-Besancon: I think I removed all unnecessary `+ E::ZERO` and `* E::ONE`. Some `* E::ONE` are left because we add/rem the `1` constant in the buses. Don't hesitate to ping me I you still see some that should not be there!

Al-Kindi-0: > I think I removed all unnecessary `+ E::ZERO` and `* E::ONE`. Some `* E::ONE` are left because we add/rem the `1` constant in the buses. Don't hesitate to ping me I you still see some that should not be there!

Just quickly double checked and it looks good now, thank you!

docs/src/description/buses.md
```diff
@@ -0,0 +1,63 @@
...
+```
+integrity_constraints {
+    q.rem(e, f, g) when s
+    q.add(a, b, c) for d
+}
+```
+
+Similarly to the previous example elements can be added or removed from `q` with binary selectors. However, as it is a LogUp-based bus, it is also possible to add and remove elements with a given scalar multiplicity with the `for` keyword (here, `d` does not have to be binary).
+
+The global resulting constraint on the column of the bus is the following: `q ′ + s ( α 0 + α 1 ⋅ e + α 2 ⋅ f + α 3 ⋅ g ) = q + d ( α 0 + α 1 ⋅ a + α 2 ⋅ b + α 3 ⋅ c )`, where `α i` corresponds to the i-th random value provided by the verifier.
```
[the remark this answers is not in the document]
Leo-Besancon: Indeed I multiplied instead of dividing, I updated the doc to use math and provide both expressions.

air-script/tests/buses/buses_complex.masm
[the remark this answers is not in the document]
Leo-Besancon: https://github.com/0xPolygonMiden/air-script/issues/372

parser/src/parser/grammar.lalrpop
Leo-Besancon: Updated the parser to avoid empty buses section and added a test in fec28df3ded49d4fbeca96a63696b45e0b578467

parser/src/ast/expression.rs
Leo-Besancon: Good point! Even if it's suboptimal, I've decided to assign the Felt type to BusOperation and Null, to stay consistent with Binary operation are handled (with the "Eq" for instance). Moreover, we never directly need the type of these exprs / scalar_exprs
Reworked this bit in 1cc8402143943541744cec54daca51750d49b3e7

Leo-Besancon:
@bobbinth, @bitwalker, @Al-Kindi-0 and @adr1anh, thank you for your comments!

In the latest commits we should have answered your comments, except for the naming conventions.

It seems that we should:
- replace `unit` with `multiset` or `channel` and `mult` with `logup` in the buses declaration section
- replace `.add(...)` and `.rem(...)` with other keywords that seem more consistent with what buses are trying to do. It mainly could be `insert/remove`, `send/receive`, `push/pull`, or others

Personnally, I like `multiset`, `logup`, `push` and `pull`, but we'll implement what reaches consensus! What should we go with?

parser/src/sema/semantic_analysis.rs
bitwalker: I was looking at the clippy warnings generated in CI. It can be pretty hard to spot those without clippy, and it might even be the case that it only appears with our particular clippy config being used in CI. If we're intentionally letting the `ControlFlow` be ignored, we usually do something like `let _ = function_that_returns_control_flow_or_result();` to indicate that it is being done on purpose. Clippy only complains when it thinks it is unintentional.

Leo-Besancon: Oh indeed using clippy on the nightly toolchain does show these warnings!
I've added the `let _ =` along with a comment when necessary in commit 0e4deb063226a0d30fa90210b7e5f20f056f6a53

mir/src/ir/graph.rs
```diff
@@ -1,45 +1,44 @@
...
-    functions: BTreeMap<QualifiedIdentifier, Link<Root>>,
-    evaluators: BTreeMap<QualifiedIdentifier, Link<Root>>,
-    pub boundary_constraints_roots: Link<Vec<Link<Op>>>,
-    pub integrity_constraints_roots: Link<Vec<Link<Op>>>,
+    functions: BTreeMap<QualifiedIdentifier, ir::Link<ir::Root>>,
+    evaluators: BTreeMap<QualifiedIdentifier, ir::Link<ir::Root>>,
+    pub boundary_constraints_roots: ir::Link<Vec<ir::Link<ir::Op>>>,
+    pub integrity_constraints_roots: ir::Link<Vec<ir::Link<ir::Op>>>,
+    pub buses: BTreeMap<QualifiedIdentifier, ir::Link<ir::Bus>>,
+    bus_count: usize,
```
paracetamolo, line 20: Where is this used?

mir/src/ir/link.rs
```diff
@@ -197,3 +197,55 @@ where
         }
     }
 }
+
+/// A wrapper around a [Link<T>] to block recursive implementations of [PartialEq] and [Hash].
```
paracetamolo, line 201: This seems a subtle concept, maybe add a few more lines of explanation with an example.

mir/src/ir/nodes/ops/bus_op.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+#[derive(Default, Clone, Eq, Debug, Spanned, Builder)]
+#[enum_wrapper(Op)]
+pub struct BusOp {
+    pub parents: Vec<BackLink<Owner>>,
+    pub bus: BackLink<Bus>,
+    pub kind: BusOpKind,
+    pub args: Vec<Link<Op>>,
+    pub _latch: Link<Op>,
```
paracetamolo, line 19: why the underscore for latch?

mir/src/ir/nodes/ops/value.rs
```diff
@@ -53,12 +65,39 @@ pub enum MirValue {
     PeriodicColumn(PeriodicColumnAccess),
     /// A reference to a specific element of a given public input
     PublicInput(PublicInputAccess),
+    // TODO: Will be used when handling variable-length public inputs
```
paracetamolo, line 68: Can we keep the variable-length notes out of this PR?

mir/src/ir/nodes/ops/value.rs
```diff
@@ -53,12 +65,39 @@ pub enum MirValue {
...
+pub struct BusAccess {
+    /// The trace segment being accessed
+    pub bus: Link<Bus>,
+    /// The offset from the current row.
+    ///
+    /// Defaults to 0, which indicates no offset/the current row.
+    ///
+    /// For example, if accessing a trace column with `a'`, where `a` is bound to a single column,
+    /// the row offset would be `1`, as the `'` modifier indicates the "next" row.
+    pub row_offset: usize,
```
paracetamolo, line 93: can this be more than 1?

mir/src/ir/nodes/ops/value.rs
```diff
@@ -164,6 +203,20 @@ impl PublicInputAccess {
     }
 }
 
+// TODO: Will be used when handling variable-length public inputs
```
paracetamolo, line 206: Same as above.

mir/src/passes/unrolling.rs
```diff
@@ -214,6 +214,8 @@ impl UnrollingFirstPass<'_> {
 
                     updated_value = Some(Vector::create(vec, value_ref.span()));
                 }
+                MirValue::BusAccess(_) => {}
+                MirValue::Null => {} //MirValue::PublicInputBinding(_public_input_binding) => {}
```
paracetamolo, line 218: Comment?

mir/src/tests/buses.rs
```diff
@@ -0,0 +1,207 @@
...
+    let _p_rem = bus.rem(&[x.clone()], not_sel.clone(), SourceSpan::default());
+    let bus_ident = result_mir.constraint_graph().buses.keys().next().unwrap();
+    let mut expected_mir = Mir::new(result_mir.name);
+    let _ = expected_mir
+        .constraint_graph_mut()
+        .insert_bus(*bus_ident, bus.clone().clone());
+    assert_bus_eq(&mut expected_mir, &mut result_mir);
+}
+
+// Tests that should return errors
```
paracetamolo, line 149: These negative tests are nice. Are there more cases?

mir/src/tests/buses.rs
```diff
@@ -0,0 +1,207 @@
...
+        unit p,
+        mult q,
+    }
+
+    public_inputs {
+        inputs: [2],
+    }
+
+    boundary_constraints {
+        enf p.first = 0;
```
paracetamolo, line 169: Can I have an expression that evaluates to a constant?

mir/src/tests/buses.rs
paracetamolo: Can I have a bus refer to another? `enf q.first = p.last`?

parser/src/sema/binding_type.rs
```diff
@@ -218,6 +223,8 @@ impl BindingType {
                 _ => Err(InvalidAccessError::IndexIntoScalar),
             },
             Self::Function(_) => Err(InvalidAccessError::InvalidBinding),
+            // TODO: FIXME
```
paracetamolo, line 226: leftover TODO

parser/src/sema/semantic_analysis.rs
```diff
@@ -1378,37 +1422,116 @@ impl SemanticAnalysis<'_> {
...
-                                )?;
+                                    "public input bus constraints are not yet supported",
+                                )
+                                .with_secondary_label(
+                                    access.name.span(),
+                                    "this a reference to a public input",
+                                )
+                                .emit();
+
+                                // TODO: Update when table public inputs are supported
```
paracetamolo, line 1440: Big commented section

mir/src/passes/bus_op_expand.rs
```diff
@@ -0,0 +1,451 @@
...
+}
+
+impl Pass for BusOpExpand<'_> {
+    type Input<'a> = Mir;
+    type Output<'a> = Mir;
+    type Error = CompileError;
+
+    fn run<'a>(&mut self, mut ir: Self::Input<'a>) -> Result<Self::Output<'a>, Self::Error> {
+        let mut max_num_random_values = ir.num_random_values as usize;
+        // TODO: When removing aux and rand values, use the following instead
```
paracetamolo, line 32: Maybe keep it in a branch?

mir/src/passes/bus_op_expand.rs
```diff
@@ -0,0 +1,451 @@
...
+                AccessType::Default,
+                1,
+                bus.borrow().span(),
+            );
+
+            // Expand bus boundary constraints first
+            self.handle_boundary_constraint(bus_type.clone(), first/*, air_parser::ast::Boundary::First, bus_access.clone(), bus.borrow().span()*/);
+            self.handle_boundary_constraint(
+                bus_type.clone(),
+                last, /*, air_parser::ast::Boundary::Last, bus_access.clone(), bus.borrow().span()*/
```
paracetamolo, line 72: commented line here and above

mir/src/passes/bus_op_expand.rs
```diff
@@ -0,0 +1,451 @@
...
+                            BusType::Mult => 0, // Sum of inverses, unit for sum is 0
+                        };
+                        let unit_val = Value::create(SpannedMirValue {
+                            span: SourceSpan::default(),
+                            value: MirValue::Constant(ConstantValue::Felt(unit_constant)),
+                        });
+
+                        to_update = Some(unit_val);
+
+                        /*let bus_boundary = Boundary::create(
```
paracetamolo, line 375: commented code

mir/src/passes/bus_op_expand.rs
```diff
@@ -0,0 +1,451 @@
...
+            _ => unreachable!(),
+        }
+
+        if let Some(to_update) = to_update {
+            link.set(&to_update);
+        }
+    }
+}
+
+/*impl Visitor for BusOpExpand<'_> {
```
paracetamolo, line 401: commented code

paracetamolo (commented):
Left some minor nits. For the future try to separate refactors from new features. For example the `Singleton` refactor could have been a separate PR. Also a cleaner git history helps breaking a huge diff in more digestible chunks.

docs/src/description/buses.md
```diff
@@ -0,0 +1,63 @@
...
+```
+integrity_constraints {
+    p.add(a) when s1;
+    p.rem(a, b) when 1 - s2;
+}
+```
+
+Here, `s1` and `1 - s2` are binary selectors: the element is added or removed when the corresponding selector's value is 1.
+
+The global resulting constraint on the column of the bus is the following, where \\(\alpha_i\\) corresponds to the i-th random value provided by the verifier: \\[p ′ ⋅ ( ( \alpha_0 + \alpha_1 ⋅ a + \alpha_2 ⋅ b ) ⋅ ( 1 − s2 ) + s2 ) = p ⋅ ( ( \alpha_0 + \alpha_1 ⋅ a ) ⋅ s1 + 1 − s1 ))\\]
```
bobbinth, line 50: Could we use the `$` and `$$`-based blocks? This way, math notation would be rended correctly both on Github and in mdBook.

In this specific case, I'd put the main formula into a separate block (i.e., using `$$`). Also, instead of `⋅` I'd use `\cdot`.

This is also relevant for other formulas in this file.

docs/src/description/buses.md
```diff
@@ -0,0 +1,63 @@
+# Buses
+
+A bus is a construct which aims to simplify description of non-local constraints defined via multiset or LogUp checks.
+
+## Bus types
+
+- Multiset (`unit`): Multiset-based buses can represent constraints which specify values that must have been added or removed from a column, in no particular order.
+[Incremental Multiset Hash Functions and Their Application to Memory Integrity Checking - Clarke et al. MIT CSAIL (2018)](https://people.csail.mit.edu/devadas/pubs/mhashes.pdf)
```
bobbinth, line 8: I would also add a link to [our docs](https://0xpolygonmiden.github.io/miden-vm/design/lookups/multiset.html) somewhere here.

docs/src/description/buses.md
```diff
@@ -0,0 +1,63 @@
...
+
+- LogUp (`mult`): LogUp-based buses are more complex than multiset buses, and can encode the multiplicity of an element: an element can be added or removed multiple times.
+[Multivariate lookups based on logarithmic derivatives - Ulrich Haböck, Orbis Labs, Polygon Labs](https://eprint.iacr.org/2022/1530)
```
bobbinth, line 11: I would also add a link to [our docs](https://0xpolygonmiden.github.io/miden-vm/design/lookups/logup.html) somewhere here.

air-script/tests/buses/buses_simple.rs
```diff
@@ -0,0 +1,94 @@
...
+        result[1] = (aux_rand_elements.get_segment_elements(0)[0] + E::ONE * aux_rand_elements.get_segment_elements(0)[1] + E::from(2_u64) * aux_rand_elements.get_segment_elements(0)[2]) * (aux_rand_elements.get_segment_elements(0)[0] + E::ONE * aux_rand_elements.get_segment_elements(0)[1] + E::from(2_u64) * aux_rand_elements.get_segment_elements(0)[2]) * aux_current[1] + (aux_rand_elements.get_segment_elements(0)[0] + E::ONE * aux_rand_elements.get_segment_elements(0)[1] + E::from(2_u64) * aux_rand_elements.get_segment_elements(0)[2]) * E::from(main_current[0]) - ((aux_rand_elements.get_segment_elements(0)[0] + E::ONE * aux_rand_elements.get_segment_elements(0)[1] + E::from(2_u64) * aux_rand_elements.get_segment_elements(0)[2]) * (aux_rand_elements.get_segment_elements(0)[0] + E::ONE * aux_rand_elements.get_segment_elements(0)[1] + E::from(2_u64) * aux_rand_elements.get_segment_elements(0)[2]) * aux_next[1] + (aux_rand_elements.get_segment_elements(0)[0] + E::ONE * aux_rand_elements.get_segment_elements(0)[1] + E::from(2_u64) * aux_rand_elements.get_segment_elements(0)[2]) * E::from(2_u64));
```
bobbinth, line 92: @Al-Kindi-0, @paracetamolo, @adr1anh - could you double-check this output?

bobbinth (commented):
Looks good! Thank you! I left a few small comments inline.

Regarding naming, unless someone feels strongly against it, let's go with:

- `multiset` and `logup` for the bus names.
- `insert` and `remove` for the action names.

air-script/tests/buses/buses_simple.rs
Al-Kindi-0: Double checked and it looks good to me!

Al-Kindi-0:
> Regarding naming, unless someone feels strongly against it, let's go with:
> 
>     * `multiset` and `logup` for the bus names.
> 
>     * `insert` and `remove` for the action names.

Sounds good to me! Only thing that still confuses me is the `for` keyword. What do you, and others, think about changing it to `with` so as to align with LogUp i.e., ".. with multiplicity"?

mir/src/ir/nodes/ops/value.rs
Leo-Besancon: Currently, in AirScript the frontend only allows for 0 or 1 (with direct column access `a` or a with a prime `a'`), but I think the original row_offset (for TraceAccess: https://github.com/massalabs/air-script/blob/feat-implement-bus/mir/src/ir/nodes/ops/value.rs#L120) was kept generic in case we wanted to connect with other frontends (IIRC Starknet can have bigger offsets?)

However, in the context of buses I'm not sure if this would ever be useful.

mir/src/tests/buses.rs
Leo-Besancon: Currently, based on https://github.com/massalabs/air-script/pull/8#issuecomment-2664755665 and the fact we don't handle variable-length public inputs yet, we only allow:
- Bus access on the left, and `null` on the right. Other buses (`enf q.first = p.last`) or constants (`enf p.first = A` with `A` a constant) will error out, but only because they are not "null".
- Trace columns on the left, to be constrained by anything (that was already valid before this PR) but `null` and bus access on the right.

So in bd032c058ce0593efe2676efc4cbe2db4ec5deb9 I've have added two negative test cases, I think it covers everything for now:
- `enf q.first = p.last`
- `enf a.first = p.last` (same but `a` is a normal trace column, not a bus)

mir/src/ir/nodes/ops/bus_op.rs
Soulthym: 9b513c3 attempts to fix this API mismatch.

This was due to the `mir::passes::translate` not knowing the latch at the point of creation. This would have forced to return an invalid (incomplete) `Op`, as it would still be a builder type, but the function's return signature requires it being a complete `Op`.
The old solution was to bypass the Builder macro by prepending an _ to the field (like we do with singletons).

I have instead set a placeholder `Link<Op::Value>` when translating. That is then replaced once the selector is known.

mir/src/tests/buses.rs
Leo-Besancon: (answered above, no it is not possible)

mir/src/ir/link.rs
Soulthym: I've added a rationale around its usage in optimizing/simplifying shared mutability, and an example with a few notes.

Please let me know if that's clear enough, like you said, it is fairly subtle.

docs/src/description/buses.md
```diff
@@ -0,0 +1,65 @@
...
+```
+integrity_constraints {
+    q.remove(e, f, g) when s
+    q.insert(a, b, c) for d
+}
+```
+
+Similarly to the previous example elements can be inserted or removed from `q` with binary selectors. However, as it is a LogUp-based bus, it is also possible to add and remove elements with a given scalar multiplicity with the `for` keyword (here, `d` does not have to be binary).
+
+The global resulting constraint on the column of the bus is the following, where $\alpha_i$ corresponds to the i-th random value provided by the verifier: $$q ′ + \frac{s}{ \alpha_0 + \alpha_1 \cdot e + \alpha_2 \cdot f + \alpha_3 \cdot g } = q + \frac{d}{ \alpha_0 + \alpha_1 \cdot a + \alpha_2 \cdot b + \alpha_3 \cdot c}$$
```
bobbinth, line 63: I would write the main equation in a separate block - e.g.,:
```
$$
q' + \frac{s}{ \alpha_0 + \alpha_1 \cdot e + \alpha_2 \cdot f + \alpha_3 \cdot g } = q + \frac{d}{ \alpha_0 + \alpha_1 \cdot a + \alpha_2 \cdot b + \alpha_3 \cdot c}
$$
```
And then it will show up as:

$$
q' + \frac{s}{ \alpha_0 + \alpha_1 \cdot e + \alpha_2 \cdot f + \alpha_3 \cdot g } = q + \frac{d}{ \alpha_0 + \alpha_1 \cdot a + \alpha_2 \cdot b + \alpha_3 \cdot c}
$$

docs/src/description/buses.md
```diff
@@ -0,0 +1,65 @@
...
+
+If we respectively note $ v_+ = \alpha_0 + \alpha_1 \cdot a + \alpha_2 \cdot b + \alpha_3 \cdot c $ and $ v_- = \alpha_0 + \alpha_1 \cdot e + \alpha_2 \cdot f + \alpha_3 \cdot g$ the tuples inserted into and removed from the bus, the equation can be rewritten: $$( q ′ - q ) \cdot v_+ \cdot v_- = s \cdot v_+  + d \cdot v_- $$
```
bobbinth, line 65: The first two equations don't render on Github. To fix this, we need to remove leading/trailing whitespace before/after `$`. For example:
```
$v_+ = \alpha_0 + \alpha_1 \cdot a + \alpha_2 \cdot b + \alpha_3 \cdot c$ and $v_- = \alpha_0 + \alpha_1 \cdot e + \alpha_2 \cdot f + \alpha_3 \cdot g$
```
Also, the last equation can go into a separate block (similar to what I described in the previous comment) to render like so:

$$
( q ′ - q ) \cdot v_+ \cdot v_- = s \cdot v_+ + d \cdot v_- 
$$

bobbinth (commented):
Looks good! Thank you! I left a couple of very minor comments inline. Also, @bitwalker, could you take another look?

> Only thing that still confuses me is the `for` keyword. What do you, and others, think about changing it to `with` so as to align with LogUp i.e., ".. with multiplicity"?

I actually like `with` better than `for` - and I don't think we are using `with` for anything. So, let's switch to it.

docs/src/description/buses.md
Leo-Besancon: Oh my bad, I didn't check the rendering on github on that last pass. It should be good now! I also made the switch to "with"

bitwalker (approved):
LGTM!

bobbinth (approved):
All looks good! Thank you!

@Leo-Besancon - could you guys create issues for the follow-up tasks (if there are some that are missing)? This would include the issue for computed boundary constraints.

PR #375, opened 2025-04-08
Refactor: remove `aux` trace and `rand` value in frontend

Leo-Besancon:
This PR aims to close https://github.com/0xPolygonMiden/air-script/issues/373

Following https://github.com/0xPolygonMiden/air-script/pull/371, it is no longer necessary to expose Random Values and the Auxiliary trace in AirScript. Users should use buses instead.

Task list:
- [x] Remove random values and aux trace parsing
- [x] Update documentation
- [x] Update tests
- [x] As this refactor contains changes over most of the codebase, do an additional pass to see what could be simplified/improved

In this PR, we will not refactor `Vec<TraceSegment>` into `TraceSegment`. This part will need to be refactored / adapted depending on a design for multi-main trace support in AirScript (see PR comments for details). Commit https://github.com/0xPolygonMiden/air-script/commit/fed3d095f254e811f5d743d731ee1d501b32065d, not included in this branches history, could be used as a starting point if needed


adr1anh:
Still need to take a deeper look at this, but there are some cases where we may want multiple segments and auxiliary values.

While Miden does not use them now, it's common to have preprocessed traces that contain "trusted" data, such as a table of program instructions, or pre-computed bitwise operations. It could also be used to implement some form of continuations between proofs by sharing tables between two different executions.

Random values can also be used outside of lookups. For example, [this talk](https://youtu.be/R07ina4k7hg?si=O-VfUmlMw4Sdkq7m) outlines some ideas for accelerating non-native arithmetic or SIMD-like behavior inside a STARK.

bobbinth:
> While Miden does not use them now, it's common to have preprocessed traces that contain "trusted" data, such as a table of program instructions, or pre-computed bitwise operations

Could this not be done using the bus construct we have? This is kind of how I was thinking we'd handle the current "Kernel ROM" chiplet (which relies on a trusted "lookup table"). But maybe there is something I'm missing?

> It could also be used to implement some form of continuations between proofs by sharing tables between two different executions.

> Random values can also be used outside of lookups. For example, [this talk](https://youtu.be/R07ina4k7hg?si=O-VfUmlMw4Sdkq7m) outlines some ideas for accelerating non-native arithmetic or SIMD-like behavior inside a STARK.


I haven't thought about these too much, but maybe here too we can use buses? Basically, while we call them "busses" they can also be used as lookup tables (I think) and so should be able to handle most use cases excluding the cases where we need to enforce relations directly between columns of the auxiliary trace. But I think such situations are pretty rare - though, maybe that's exactly what we need for non-native arithmetic or SIMD-like behavior?

adr1anh:
> Could this not be done using the bus construct we have? This is kind of how I was thinking we'd handle the current "Kernel ROM" chiplet (which relies on a trusted "lookup table"). But maybe there is something I'm missing?

Absolutely, and the Kernel ROM is also a good candidate for preprocessed tables. My comment was referring to the fact that with this PR we are only considering a single `main` trace. Removing the challenges makes sense though.

> I haven't thought about these too much, but maybe here too we can use buses? Basically, while we call them "busses" they can also be used as lookup tables (I think) and so should be able to handle most use cases excluding the cases where we need to enforce relations directly between columns of the auxiliary trace. But I think such situations are pretty rare - though, maybe that's exactly what we need for non-native arithmetic or SIMD-like behavior?

There's multiple ways of accelerating non-native arithmetic though it's been a while since I’ve thought of it. One easy solution though would be to provide access to additional randomness for polynomial identity over values in the main trace. I think this is what Falcon already does albeit by deriving randomness within the trace itself.

adr1anh:
> I guess one thing that isn't clear to me yet is what kind of constraints would need a direct access to the auxiliary trace (and/or challenges) rather than being expressible via the bus construct? For our Kernel ROM needs, I think the bus construct is sufficient (cc @Al-Kindi-0 in case I'm off here).

I don't think access to the auxiliary trace is necessary, but this PR seems to bake in the assumption that there will only be one main trace. For the Kernel ROM, it would likely be a pre-processed trace. In the case of a "bytes chiplet", the constraints would apply over the pre-processed and main trace as the latter would include logUp multiplicities. 

> How would the constraints look like in this case? Basically, I'm wondering if this is something that needs direct access to the auxiliary trace or if (1) it can be emulated using busses or (2) it may make sense to introduce some higher-level construct for this (e.g., at the same level as bus construct).

The kind of constraints I’m thinking of would only need access to a random challenge, but this might indeed be worth including as an additional higher-level abstraction. 

We can treat arrays of field elements in the main trace as polynomials, e.g. `p(X), q(X), r(X)`, we could check that `p(X) * q(X) = r(X)` by evaluating each polynomial in some challenge `alpha`, rather than computing the coefficients of the polynomial multiplication.


It may be a bit far out for now, but might help later on when implementing non-native field arithmetic for e.g. ECDSA signature verification.

bobbinth:
> I don't think access to the auxiliary trace is necessary, but this PR seems to bake in the assumption that there will only be one main trace. For the Kernel ROM, it would likely be a pre-processed trace. In the case of a "bytes chiplet", the constraints would apply over the pre-processed and main trace as the latter would include logUp multiplicities.

Ah - I see what you mean, and to confirm, you mean we may want to have multiple main traces - right?. If so, this is not currently supported by AirScript and we probably need to create an issue for this. I actually haven't thought through what the syntax for "multi-table" main trace for AirScript could look like.

Leo-Besancon:
Thank you all for the feedback. So from what I gather from the discussion, I could:
- Revert the commit replacing Vec<TraceSegment> with TraceSegment, to avoid losing the future possibility to handle multiple main traces.
- Create a followup issue to handle multiple main traces

Then we can merge this PR removing the `aux`, and later work on the design for multiple main traces. What do you think?

Al-Kindi-0:
> Thank you all for the feedback. So from what I gather from the discussion, I could:
> 
>     * Revert the commit replacing `Vec<TraceSegment>` with `TraceSegment`, to avoid losing the future possibility to handle multiple main traces.
> 
>     * Create a followup issue to handle multiple main traces
> 
> 
> Then we can merge this PR removing the `aux`, and later work on the design for multiple main traces. What do you think?

Thank you, Leo! Sounds good to me, but let's wait for the other opinions.

adr1anh:
That sounds good to me too! Multiple traces have a clear use case in the near term future. As for accessing the challenges I think it's a lot more niche and we may want to wait until we have an idea of how to use them before we start supporting it in AirScript (accelerating non-native arithmetic for signature verification).

bobbinth:
> * Revert the commit replacing `Vec<TraceSegment>` with `TraceSegment`, to avoid losing the future possibility to handle multiple main traces.

I don't have a strong preference here: keeping `TraceSegment` or reverting to `Vec<TraceSegment>` are both fine as future solutions to multiple main trace segments may or may not use the same structure. So, I'd go with w/e is easier.

docs/src/description/constraints.md
```diff
@@ -2,19 +2,17 @@
 
 ## Boundary constraints (`boundary_constraints`)
 
-The `boundary_constraints` section consists of expressions describing the expected value of columns in the main or auxiliary traces at the specified boundary. Column boundaries can be selected using boundary accessors. Valid boundary accessors are `.first`, which selects the first cell of the column to which it is applied, and `.last`, which selects the last cell of the column column to which it is applied.
+The `boundary_constraints` section consists of expressions describing the expected value of columns in the main trace at the specified boundary. Column boundaries can be selected using boundary accessors. Valid boundary accessors are `.first`, which selects the first cell of the column to which it is applied, and `.last`, which selects the last cell of the column column to which it is applied.
```
Al-Kindi-0, line 5: Not from this PR, but there is a repeated word

docs/src/description/constraints.md
```diff
@@ -34,56 +31,52 @@ public_inputs {
...
+Boundary constraints can access public input values provided by the verifier in their value expressions.
 
 To use public inputs, the public input must be declared in the `public_inputs` source section. They can be accessed using array indexing syntax, as described by the [accessor syntax rules](./syntax.md#section-specific-accessors).
 
-To use random values, they must be declared in the `random_values` source section. Random values are defined by an identifier which is bound to a fixed-length array (e.g. `rand: [16]`), but it is additionally possible to bind identifiers to specific random values or groups of values (e.g. `rand: [a, b[14], c]`). Random values can be accessed by using `$` followed by the array identifier and the index of the value (e.g. `$rand[10]`) or by using the name of the bound random value or group along with the index of the value within the group (e.g. `a` or `b[5]`), as described by the [accessor syntax rules](./syntax.md#section-specific-accessors).
-
-### Example of boundary constraints with public inputs and random values
+### Example of boundary constraints with public inputs and buses
```
Al-Kindi-0, line 49: I would removed/reformulate ("... with ... and buses") as logically it is buses that are with or have boundary constraints

docs/src/description/constraints.md
```diff
@@ -34,56 +31,52 @@ public_inputs {
...
 
-The following is an example of a valid `boundary_constraints` source section that uses public inputs and random values:
+The following is an example of a valid `boundary_constraints` source section that uses public inputs and buses:
```
Al-Kindi-0, line 52: same comment here i.e., it is buses that have boundary constraints

docs/src/description/constraints.md
```diff
@@ -165,30 +154,24 @@ integrity_constraints {
...
     enf a' = a + 1;
     enf b' - b - 1 = 0;
-
-    # this is an auxiliary constraint, since it uses an auxiliary trace column.
-    enf p = p' * a;
 }
 ```
 
-### Periodic columns and random values
+### Periodic columns and buses
```
Al-Kindi-0, line 160: we can probably remove "and buses"

mir/src/passes/bus_op_expand.rs
```diff
@@ -24,7 +24,7 @@ impl Pass for BusOpExpand<'_> {
     type Error = CompileError;
 
     fn run<'a>(&mut self, mut ir: Self::Input<'a>) -> Result<Self::Output<'a>, Self::Error> {
-        let mut max_num_random_values = ir.num_random_values as usize;
+        let mut max_num_random_values = 0;
```
Al-Kindi-0, line 27: Is this a temporary fix?

Al-Kindi-0 (approved):
Looks good, thank you @Leo-Besancon !
Left a few questions and comments.

mir/src/passes/bus_op_expand.rs
Leo-Besancon: As currently the frontend does not have random values but the backend does (and as we expand the bus constraints in the algebraic graph), we compute how many variables we need there.

This won't be needed and will be removed entirely when handling https://github.com/0xPolygonMiden/air-script/issues/378

codegen/masm/tests/test_boundary.rs
```diff
@@ -145,7 +145,9 @@ integrity_constraints {
     enf a + b = 0;
 }";
 
+// ignore `aux` tests
 #[test]
+#[ignore]
```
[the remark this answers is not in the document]
Leo-Besancon: Oh my bad, thank you! As we are removing MASM support it's not very impactful but indeed I forgot to update this test. I reinstated the test in commit 4467911ce6fbcda6277188527ca4b28381a7ba25

docs/src/description/constraints.md
```diff
@@ -2,19 +2,17 @@
 
 ## Boundary constraints (`boundary_constraints`)
 
-The `boundary_constraints` section consists of expressions describing the expected value of columns in the main or auxiliary traces at the specified boundary. Column boundaries can be selected using boundary accessors. Valid boundary accessors are `.first`, which selects the first cell of the column to which it is applied, and `.last`, which selects the last cell of the column column to which it is applied.
+The `boundary_constraints` section consists of expressions describing the expected value of columns in the main trace at the specified boundary. Column boundaries can be selected using boundary accessors. Valid boundary accessors are `.first`, which selects the first cell of the column to which it is applied, and `.last`, which selects the last cell of the column to which it is applied.
```
bobbinth, line 5: Boundary constraints can also specify constraints for busses, right? I think we should mention this here.

docs/src/description/constraints.md
```diff
@@ -34,56 +31,52 @@ public_inputs {
...
-    # these are auxiliary constraints that use public input values.
-    enf p0.first = stack_inputs[1]
-    enf p0.last = stack_outputs[1]
+    enf a.first = stack_inputs[0];
+    enf a.last = stack_outputs[0];
 
-    # these are auxiliary constraints that use random values from the verifier.
-    enf b.first = a + $rand[0]
-    enf p1.first = (stack_inputs[2] + $rand[0]) * (stack_inputs[3] + $rand[1])
+    # these are bus constraints that use the `null` identifier to define empty bus constraints
```
bobbinth, line 75: I would write this as "these are bus constraints that specify that busses must be empty at the beginning and the end of the execution trace"

docs/src/description/constraints.md
```diff
@@ -165,30 +154,24 @@ integrity_constraints {
...
 
-Integrity constraints can access the value of any periodic column in the current row, as well as random values provided by the verifier.
+Integrity constraints can access the value of any periodic column in the current row.
 
 To use periodic column values, the periodic column must be declared in the `periodic_columns` source section. The value in the current row can then be accessed by using the defined identifier of the periodic column.
 
-Random values can be accessed by using array indexing syntax on the `$rand` built-in, as described by the [accessor syntax rules](./syntax.md#section-specific-accessors).
-
-### Example of integrity constraints with periodic columns and random values
+### Example of integrity constraints with periodic columns
```
bobbinth, line 166: Should we have a section with integrity constraints with busses? Or maybe just add busses to this section?

bobbinth (approved):
Looks good! Thank you! Not a very deep review from me, but I left a few small comments inline.

Also, it seems like the tests are not currently passing?

docs/src/description/organization.md
```diff
@@ -51,9 +49,7 @@ where the name of the module must:
 - Start with a letter.
 - End with a newline.
 
-Besides the name declaration, library modules my contain definitions of constants, evaluators, and periodic columns. Constants and evaluators defined in a library module may be imported by a root or other library modules.
-
-Library modules inherit random value declarations of the root module. That is, evaluators defined in a library module can reference random values declared in the root module.
```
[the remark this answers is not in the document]
Leo-Besancon: Currently libraries do not inherit bus declarations. We also don't currently handle them in evaluators.
We can change both behaviors in a followup PR if needed, it may need some work in how we handle Buses and evaluator inlining in the Mir but I don't expect too much trouble.

The only design choice we would need is whether we limit the bus accessible in evaluators. Similarly to how evaluators were limited to only use specific aux trace columns given as arguments, we could do:

```
buses {
    multiset p,
    logup q,
}

ev foo([a,s], [b]) {
	# Valid, b is the bus given as parameter
	b.insert(a) when s;
	
	# Invalid
	q.insert(a) when s;
}

integrity_constraints {
    foo([a,s], [p])
}
```

Leo-Besancon:
> Also, it seems like the tests are not currently passing?

This was due to the merge of ACE backend work into next branch.
I rebased on next and fixed the impacted tests on the Ace backend.

As I fixed what I assume was a bug, I would welcome some feedback (maybe @paracetamolo can clear this up?) on this change: https://github.com/0xPolygonMiden/air-script/pull/375/commits/b17eaf87d8798d73a751f99be152770089c1a4f2#diff-631c6bc3d361a76e8c4d0c100ef8cc796e585132c76289604910cdc155c5f600R295
Previously there was some cases with overlap of e.g. `TraceAccess { segment: 0, column: 0, offset: 1 }` and `TraceAccess { segment: 1, column: 0, offset: 0 }`

Leo-Besancon:
@Al-Kindi-0 quick question after the last rebase following the merge of https://github.com/0xPolygonMiden/air-script/pull/383, I couldn't find what I did wrong at first, and I noticed that if I put `b_prime - b` in the integrity constraint Vecv, it lead to a non zero evaluation, while `b - b_prime` worked as expected. Is this expected to be different evaluations?

Al-Kindi-0:
The key idea is this, although both `b_prime - ((random[0] + random[1] * a) * s + ONE - s) * b` and `((random[0] + random[1] * a) * s + ONE - s) * b - b_prime` will evaluate to zero on the main execution trace, they will evaluate to a random value that is non-zero with high probability, say $\gamma$ for the first and $-\gamma$ for the second, respectively, during constraints evaluation (as part of the STARK/SNARK verification procedure). The is because each entry of the above expressions, say for example `b_prime` will be substituted by the evaluation of a polynomial, representing the column off all `b_prime` values, at some random point $z$ or $\zeta$ (using Plonk notation).

The above means that one has to use the same expression used by the circuit generated by the ACE backend in order to get the correct value.

Leo-Besancon:
That makes sence, thank you for the explanation!

codegen/masm/tests/test_divisor.rs
```diff
@@ -116,7 +116,9 @@ integrity_constraints {
     enf a = 0;
 }";
 
+// ignore `aux` tests
```
paracetamolo, line 119: why ignore and not remove?

docs/src/description/constraints.md
```diff
@@ -199,24 +182,46 @@ periodic_columns {
...
+}
+```
+
+### Buses
+
+Integrity constraints can constrain insertions and removal of elements into / from a given bus. The bus must first be declared in the `buses` source section. More information on bus types and the associated constraints can be found in the [buses](./buses.md) section.
 
-    # this is an auxiliary constraint that uses a periodic column.
-    enf p0' = k * p0;
+### Example of integrity constraints with buses
```
paracetamolo, line 199: Maybe I missed it but is there a section explaining how to do random values with buses?

paracetamolo (commented):
Nice work, a lot of cleanup!

codegen/masm/tests/test_divisor.rs
Leo-Besancon: Thnaks, removed one of the tests and fixed the other one

docs/src/description/organization.md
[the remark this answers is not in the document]
bobbinth: Makes sense. Let's create an issue for this and tackle it in a follow-up PR.

The way I'm thinking about busses is that they could work similarly to how random values used to work. That is, we declare them ones in the main module, but then can access them in evaluators in library modules (i.e., no need to pass them via the evaluator signature). Though, maybe there is an argument to be made that passing them via signatures is desirable.

bobbinth (approved):
Thank you! Looks good! I think this is good to merge, right?

docs/src/description/organization.md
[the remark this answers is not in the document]
Leo-Besancon: Ok that would work! https://github.com/0xMiden/air-script/issues/387

Leo-Besancon:
> Thank you! Looks good! I think this is good to merge, right?

Thanks! On my side it's good, I'll let @paracetamolo or others react if needed!

paracetamolo (approved):

PR #383, opened 2025-04-28
Fix formula final formula for ACE codegen and improve tests a bit

Al-Kindi-0:
The final formula computed in the ACE codegen didn't take into account the exception point used to exclude the last row of the execution trace. This fixes this problem and also some minor issues.
Tests have been improved a bit.


adr1anh:
It's not entirely clear why we are completely excluding the last row. I think it might be for zero-knowledge since it allows the last row to be entirely randomized, leading to independence between the out-of-domain sample to be independent of the trace. However, wouldn't we need to exclude the last **two** rows due to the openings that $$z$$ and $$gz$$?

I'll review the code shortly, though, and if the above holds it should be an easy fix.

codegen/ace/src/codegen.rs
```diff
@@ -75,16 +99,19 @@ pub fn build_ace_circuit(ir: &Air) -> anyhow::Result<(NodeIndex, Circuit)> {
         .chain(boundary_last_roots.clone())
         .collect();
 
-    // A circuit builder is instantiated with the inputs of the circuits plus the 13 needed by the ACE chiplet
+    // A circuit builder is instantiated with the inputs of the circuits plus the `NUM_ACE_AUX_INPUTS` needed by the ACE chiplet
     let airscript_inputs = n_random(ir) + n_public_inputs(ir) + n_trace_accesses(ir);
-    let auxiliary_inputs = 13;
+    let auxiliary_inputs = NUM_ACE_AUX_INPUTS;
```
paracetamolo, line 104: you can remove this intermediate variable and just use the constant everywhere, not important.

paracetamolo (approved):
All good, thanks for the fixes.

codegen/ace/src/codegen.rs
Al-Kindi-0: Changed

PR #388, opened 2025-05-02
Bump winterfell dependancy

Leo-Besancon:
The generated .rs Winterfell Air constraints descriptions files are now up to date with Winterfell 0.12 crates.
I was e.g. able to validate that a given trace follows the constraints given in the Air.

However, the prover panics (mismatch in degrees) during my tests. I'll dig a bit into why (it seems that this was the same before this PR targetting Winterfell 0.6), if there is a problem during codegen or if the programs I've tested in `air-script/tests` should not be proven.

Left to include in this PR:
- [x] Include the generated Winterfell code in the test builds (to check that it compiles)
- [x] Include traces generation snippets for each `air-script/tests` test
- ~Include tests with Winterfell's prover and verifier for each `air-script/tests` test~ Will do as a followup, see https://github.com/0xMiden/air-script/issues/390


bobbinth:
> However, the prover panics (mismatch in degrees) during my tests. I'll dig a bit into why (it seems that this was the same before this PR targetting Winterfell 0.6),

If this was a problem before as well, I'd create an issue for this and tackle it in a separate PR.

Leo-Besancon:
> > However, the prover panics (mismatch in degrees) during my tests. I'll dig a bit into why (it seems that this was the same before this PR targetting Winterfell 0.6),
> 
> If this was a problem before as well, I'd create an issue for this and tackle it in a separate PR.

Great, in this PR I've added the check of the generated Air implementations, defined a trace for all test cases and validates the trace against the constraints. Most traces are very simple though (e.g. all zeros), I'm not sure how important it is to have complex traces.

I've created an issue to include proving / verifying here: https://github.com/0xMiden/air-script/issues/390

air-script/tests/functions/test_air.rs
```diff
@@ -0,0 +1,81 @@
...
+            32, // number of queries
+            16, // blowup factor
+            0,  // grinding factor
+            FieldExtension::None,
+            8,                      // FRI folding factor
+            31,                     // FRI max remainder polynomial degree
+            BatchingMethod::Linear, // method of batching used in computing constraint composition polynomial
+            BatchingMethod::Linear, // method of batching used in computing DEEP polynomial
+        )
+    }
```
Al-Kindi-0, line 65: nit: this is probably redundant?

air-script/tests/helpers/mod.rs
```diff
@@ -0,0 +1,70 @@
+use winter_air::{BatchingMethod, EvaluationFrame, FieldExtension, ProofOptions, TraceInfo};
+use winter_math::fields::f64::BaseElement as Felt;
+use winterfell::{matrix::ColMatrix, AuxTraceWithMetadata, Trace, TraceTable};
+
+// We need to encapsulate the trace table in a struct to fix the aux trace width, that always return zero otherwise
```
Al-Kindi-0, line 5: nit: Could we re-write the sentence, I am not sure I understand the last part completely.

air-script/tests/helpers/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub trait AirTester {
+    type PubInputs;
+
+    fn build_main_trace(&self, length: usize) -> MyTraceTable;
+    fn public_inputs(&self) -> Self::PubInputs;
+    #[allow(unused_variables)]
+    fn build_aux_trace(&self, length: usize) -> Option<AuxTraceWithMetadata<Felt>> {
+        None
+    }
```
Al-Kindi-0, line 45: nit: maybe we can use `#[allow(unused_variables)]` and change to `_length`?

Al-Kindi-0 (commented):
Did an initial quick quick review and it looks good.
While review, I was thinking if we might need to have a bit more realistic examples i.e., with proper constraints and filled traces. This would probably require a lot work as it will involve updating most of the AirScript files.
Maybe we can start by including a couple of proper examples, e.g., Fibonacci, and then leave the rest for future PRs.

air-script/tests/functions/test_air.rs
Leo-Besancon: We need to set the blowup factor to 16 here (I assume because the constraints are of higher degrees), and unfortunatly there are not setters or public fields in `ProofOptions`, so we specify a whole new struct.

air-script/tests/helpers/mod.rs
Leo-Besancon: Updated with more details, please tell me if it's clearer!

air-script/tests/helpers/mod.rs
Leo-Besancon: Done

Leo-Besancon:
> Did an initial quick quick review and it looks good. While review, I was thinking if we might need to have a bit more realistic examples i.e., with proper constraints and filled traces. This would probably require a lot work as it will involve updating most of the AirScript files. Maybe we can start by including a couple of proper examples, e.g., Fibonacci, and then leave the rest for future PRs.

Thank you! Indeed it seemed too much work for this PR, and I updated the minimum to make the test pass without thinking too much about what constraints would be more realistic (also to avoid removing the testing of specific features.

I've added a fibonacci test case like you suggested in the latest commits!

air-script/tests/functions/test_air.rs
Al-Kindi-0: Got it, thank you for clarifying it in the comment!

Al-Kindi-0 (approved):
Looks good, thank you Leo!

bobbinth (approved):
Thank you! Looks good! Though, this is a very high level review from my end.

Issue #372, opened 2025-03-27
Refactor / remove the MASM backend


Leo-Besancon:
The work on ACE-based constraint evaluation (see issue https://github.com/0xPolygonMiden/air-script/issues/366 or PR https://github.com/0xPolygonMiden/air-script/pull/370) changes the relevances of the MASM backend.

Once this changes are complete we should remove or refactor the MASM backend.

_Originally posted by @bobbinth in https://github.com/0xPolygonMiden/air-script/pull/371#discussion_r2009290329_

bobbinth:
Closed by #389.

PR #389, opened 2025-05-02
Refactor: remove MASM backend and tests

Leo-Besancon:
This PR closes issue https://github.com/0xMiden/air-script/issues/372.

Instead of just removing the MASM backend, would it make sense to replace it with ACE (in CLI / docs)? I think the only thing lacking is a writer implementation to serialize it to file. If this feature is expected we maybe can wait before merging this PR.


bobbinth (approved):
Looks good! Thank you!

Issue #142, opened 2023-02-10
Add support for nested list comprehension


tohrnii:
We should allow users to define nested list comprehension like:

```
let transpose = [[row[i] for row in matrix] for i in 0..5]
```

Related issue: #132

bobbinth:
@tohrnii, @bitwalker - is this still relevant or has it been addressed with the new structure?

bitwalker:
We currently restrict comprehension bodies to scalar expressions, but we could easily relax that. I was under the impression we didn't want to support nested comprehensions, but there's no reason why we can't.

bobbinth:
Yes, I think one place where this would be useful is in computing MDS matrix multiplication for RPO constraints, which would look something like this:
```
const MDS = [
    [7, 23, 8, 26, 13, 10, 9, 7, 6, 22, 21, 8],
    [8, 7, 23, 8, 26, 13, 10, 9, 7, 6, 22, 21],
    [21, 8, 7, 23, 8, 26, 13, 10, 9, 7, 6, 22],
    [22, 21, 8, 7, 23, 8, 26, 13, 10, 9, 7, 6],
    [6, 22, 21, 8, 7, 23, 8, 26, 13, 10, 9, 7],
    [7, 6, 22, 21, 8, 7, 23, 8, 26, 13, 10, 9],
    [9, 7, 6, 22, 21, 8, 7, 23, 8, 26, 13, 10],
    [10, 9, 7, 6, 22, 21, 8, 7, 23, 8, 26, 13],
    [13, 10, 9, 7, 6, 22, 21, 8, 7, 23, 8, 26],
    [26, 13, 10, 9, 7, 6, 22, 21, 8, 7, 23, 8],
    [8, 26, 13, 10, 9, 7, 6, 22, 21, 8, 7, 23],
    [23, 8, 26, 13, 10, 9, 7, 6, 22, 21, 8, 7]
]

fn apply_mds(state: vector[12]) -> vector[12]:
    return [sum([s * m for (s, m) in (state, mds_row)]) for mds_row in MDS]
```

bobbinth:
Closed by #400.

PR #400, opened 2025-05-15
feat: Implement nested list-comprehension

### LINKED ISSUE #340: Panic when compiling AirScript
Hi all,

I was trying to compile some AirScript today, and after narrowing things down a little, I think there may be a bug. I tried the following AirScript example:
[fail_example.txt](https://github.com/0xPolygonMiden/air-script/files/12113703/fail_example.txt)

and get this:

```
============================================================
Transpiling...
thread 'main' panicked at 'internal error: entered unreachable code: unexpected constraint node: Binary(
    BinaryExpr {
        op: Mul,
        lhs: SymbolAccess(
            SymbolAccess {
                name: Local(
                    Identifier(
                        "sign_bit",
                    ),
                ),
                access_type: Default,
                offset: 0,
                ty: Some(
                    Felt,
                ),
            },
        ),
        rhs: Const(
            2048,
        ),
    },
)', parser/src/transforms/inlining.rs:535:24
```

On the other hand, if I move the invocation of the evaluator out of a match into its own `enf`, it seems to compile ok: 
[success_example.txt](https://github.com/0xPolygonMiden/air-script/files/12113716/success_example.txt)

Am I combining the match/evaluator statements incorrectly? Or might this be a bug? Thanks, and let me know if there's any other info I can provide!

Soulthym:
This PR implements [Nested list comprehensions](https://github.com/0xMiden/air-script/issues/142).

Some tasks have been post-poned, see [^2]

Task list:
- [x] lift semantic analysis restrictions
- [x] fix binding issues during constant propagation:
    - [x] propagate inside list comprehensions' body.
    - [x] fix type inference when sizes are involved.
- [x] fix new unrolling edge cases.
- [ ] ~fix #340 [^1] constant exponents not getting propagated.~
- [ ] change list comprehensions' body from ScalarExpr to Expr[^2]

[^1]: It currently re-introduces a bug from https://github.com/0xMiden/air-script/issues/340, where constant exponents no longer get constant propagated.
I'm working on fixing it and/or migrate it to the new pipeline.
[^2]: not needed for this case: https://github.com/0xMiden/air-script/issues/142#issuecomment-1617433601
This task is currently post-poned for a later release due to some complexities and no immediate need in Miden.


air-script/tests/list_comprehension/list_comprehension_nested.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    fn get_aux_assertions<E: FieldElement<BaseField = Felt>>(&self, aux_rand_elements: &AuxRandElements<E>) -> Vec<Assertion<E>> {
+        let mut result = Vec::new();
+        result
+    }
+
+    fn evaluate_transition<E: FieldElement<BaseField = Felt>>(&self, frame: &EvaluationFrame<E>, periodic_values: &[E], result: &mut [E]) {
+        let main_current = frame.current();
+        let main_next = frame.next();
+        result[0] = E::ZERO + main_current[0] * E::ONE + main_current[1] * E::from(Felt::new(2_u64)) - E::from(Felt::new(3_u64));
```
bobbinth, line 85: Should this not be equivalent to:
```
(1 * main[0] + 2 * main[1]) - 3
```
But instead it seems to be:
```
0 + main[0] + 1 * main[1] * (2 - 3)
```

air-script/tests/list_comprehension/list_comprehension_nested.rs
```diff
@@ -0,0 +1,99 @@
...
+        result[1] = E::ZERO + main_current[0] * E::from(Felt::new(2_u64)) + main_current[1] * E::from(Felt::new(3_u64)) - E::from(Felt::new(5_u64));
+        result[2] = E::ZERO + main_current[0] * E::from(Felt::new(3_u64)) + main_current[1] * E::from(Felt::new(4_u64)) - E::from(Felt::new(7_u64));
```
bobbinth, line 87: Similarly, if my understanding is correct, these should be:
```
(2 * main[0] + 3 * main[1]) - 5
```
And
```
(3 * main[0] + 4 * main[1]) - 7
```
But I think the results are different.

bobbinth (commented):
Thank you! Looks good. A very limited review from me (so, would be great for @Al-Kindi-0 and @bitwalker to take a look as well), but I left a couple of comments inline. Unless I misunderstood something, it seems like the results we are getting (at least with Winterfell backend) are somewhat off.

air-script/tests/list_comprehension/list_comprehension_nested.rs
Leo-Besancon: Hello, thank you for the review @bobbinth ! Unless I'm mistaken, the results should correspond to what you expect.
There are two differences:
- The parenthesis, which are not needed, Wintefell will do the multiplication before the addition/substractions, so
```
(1 * main[0] + 2 * main[1]) - 3
```
is equivalent to 
```
1 * main[0] + 2 * main[1] - 3
```
Since we've bumped Winterfell, the expression is less clear as previous `E::from(2_u64)` is now written `E::from(Felt::new(2_u64))`, but checking again the parenthesis seems good to me.
- The initial `E::ZERO + `, which is comming from how we handle the fold to sum the vector. In the Mir, we have a generic initial_value for the sum, so if we want to improve this bit we need to test explicitely whether this value is 0 and remove it if it's the case.

Al-Kindi-0 (approved):
Looks good to me, thank you!

parser/src/transforms/constant_propagation.rs
```diff
@@ -238,6 +243,15 @@ impl VisitMut<SemanticAnalysisError> for ConstantPropagation<'_> {
...
                         ControlFlow::Continue(())
                     }
+                    Err(SemanticAnalysisError::InvalidExpr(
+                        InvalidExprError::NonConstantExponent(_),
+                    )) if self.in_list_comprehension => {
+                        // If we are in a list comprehension, we do not know iterators' lengths
+                        // yet.
+                        // TODO: Add semantic analysis pass in Mir after all passes are done to
+                        // check for non-constant exponents in list comprehensions
+                        ControlFlow::Continue(())
```
bobbinth, line 253: Should we create an issue for this?

bobbinth (approved):
Thank you! Looks good! I left one small comment inline, but beyond that I'd wait for @bitwalker's review before merging. Also:

- It would be good to add an example of nested list comprehension to the docs (e.g., somewhere [here](https://github.com/0xMiden/air-script/blob/next/docs/src/description/convenience.md#list-comprehension)), and explain the current limitations.
- Create an issue for "change list comprehensions' body from ScalarExpr to Exp" point.

Soulthym:
Thanks for the reviews @bobbinth and @Al-Kindi-0!

> * It would be good to add an example of nested list comprehension to the docs (e.g., somewhere [here](https://github.com/0xMiden/air-script/blob/next/docs/src/description/convenience.md#list-comprehension)), and explain the current limitations.

Done in d1f8243

> * Create an issue for "change list comprehensions' body from ScalarExpr to Exp" point.

I've opened this issue for it: https://github.com/0xMiden/air-script/issues/407

bitwalker (approved):
Looks good overall! There are some bits of how we have to work with MIR that feel a bit awkward to me, but there may not really be a way to improve that currently. In any case, it's certainly not a blocker, just something we should keep our eyes on.

Issue #396, opened 2025-05-08
Multi auxiliary columns bounday constraints


Al-Kindi-0:
We sometimes might need to define boundary constraints involving more than one bus. As we support two types of buses, we can look at how these are defined for each one separately:

1. Multi-set: For some collection of buses $`\{\mathsf{b}_i\}_{i\in S}`$, we define the following equation on the last row of the the execution trace $`\prod_{i \in J} \mathsf{b}_i = \prod_{i \in S\setminus J} \mathsf{b}_i`$ where $`J \subset S`$ and the convention that $`\prod_{i \in \emptyset} \mathsf{b}_i = 1`$.
2. LogUp: For some collection of buses $`\{\mathsf{b}_i\}_{i\in S}`$, we define the following equation on the last row of the the execution trace $`\sum_{i \in J} \mathsf{b}_i = \sum_{i \in S\setminus J} \mathsf{b}_i`$ where $`J \subset S`$ and the convention that $`\sum_{i \in \emptyset} \mathsf{b}_i = 0`$.

A few points worth mentioning:
1. The buses $`\{\mathsf{b}_i\}_{i\in S}`$ should all be of the same type i.e., either all are `multiset` or all are `logup`.
2. For our current use case, defining these kinds of constraints on the last trace row should be enough. However, I think that adding support also against the first row should not be too much work and these might come in useful in the medium term when implementing ideas related to `continuations` in its broadest sense.

bobbinth:
I'm trying to think how this could look like syntax-wise. One option could be something like this:
```
buses {
    logup p,
    logup q,
}

public_inputs {
    foo: [[2]],
}

boundary_constraints {
    p.last + q.last = foo
}
```
But this may be more than we need because:

- It is not clear if we need to support arbitrary arithmetic operations when combining busses.
- This makes it seem like `q.first + p.last` is possible - but I'm not sure we want this.

So, maybe something like this could be better syntax:
```
boundary_constraints {
    [p & q].last = foo
}
```
I'm open to other suggestions as well, and also wonder if this would be affected by the route we take in #379.

adr1anh:
This is a great idea. SP1 has(/had?) something similar, which enabled lookup sharing between different proof, which in the context of memory meant continuations. The issue was the increased latency, since they had to wait for the traces of all CPU chunks before generating the lookup randomness. Something to keep in mind.

Before going into my comments, I just want to lay down my assumptions to make sure we're talking about the same thing.
[...]

Issue #377, opened 2025-04-15
Working example Miden VM with minimal set of constraints


Al-Kindi-0:
We need a minimal Air-Script example for the Miden VM AIR in order to test the recent work on the ACE chiplet.
We should probably include a main trace transition constraint and boundary constraint (e.g., the clock cycle) and an auxiliary transition constraint and boundary constraint.
Additional context, and as this is not yet in the docs, the main trace is of width $73$ columns and the auxiliary trace is of width $8$ columns.

Al-Kindi-0:
Here is an attempt:

```
def MidenVMMinimal

trace_columns {
    main: [m[73]],
    aux: [a[8]],
}

public_inputs {
    stack_inputs: [16],
    stack_outputs: [16],
    program_digest: [4],
}

boundary_constraints {
    # Clock should start at 0
    enf m[0].first = 0;
    # Range checker bus should end at 1 (or any value it started with)
    enf a[4].last = 1;
}

periodic_columns {
    k0: [0, 0, 0, 0, 0, 0, 0, 1], 
    k1: [0, 0, 0, 0, 0, 0, 1, 0], 
    k2: [1, 0, 0, 0, 0, 0, 0, 0],
    k3: [5789762306288267392, 12987190162843096997, 18072785500942327487, 5674685213610121970, 4887609836208846458, 16308865189192447297, 7123075680859040534, 0],
    k4: [6522564764413701783, 653957632802705281, 6200974112677013481, 5759084860419474071, 3027115137917284492, 11977192855656444890, 1034205548717903090, 0],
    k5: [17809893479458208203, 4441654670647621225, 17682092219085884187, 13943282657648897737, 9595098600469470675, 12532242556065780287, 7717824418247931797, 0],
    k6: [107145243989736508, 4038207883745915761, 10599526828986756440, 1352748651966375394, 10528569829048484079, 14594890931430968898, 3019070937878604058, 0],
    k7: [6388978042437517382, 5613464648874830118, 975003873302957338, 17110913224029905221, 7864689113198939815, 7291784239689209784, 11403792746066867460, 0],
    k8: [15844067734406016715, 13222989726778338773, 8264241093196931281, 1003883795902368422, 17533723827845969040, 5514718540551361949, 10280580802233112374, 0],
    k9: [9975000513555218239, 3037761201230264149, 10065763900435475170, 4141870621881018291, 5781638039037710951, 10025733853830934803, 337153209462421218, 0],
[...]

PR #408, opened 2025-06-18
feat: Allow unconstrained boundaries for buses

Leo-Besancon:
As currently boundary constraints can only be in the form `b.boundary = null | input`, it is not currently possible to represent some constraints needed for MidenVM, so we may want to allow unconstraint boudaries for now:

 _Originally posted by @Al-Kindi-0 in [#377](https://github.com/0xMiden/air-script/issues/377#issuecomment-2886593538):_

> Note that before the changes in #375 , we could have un-constrained columns in the auxiliary trace which allowed us to not be blocked by the need for having a fix for #396 and hence be able to generate the ACE circuit for the reduced set of Miden VM constraints.
> As far as I can see this is no longer possible i.e., once we define a bus we have to constrain it and without #375 we cannot define the correct constraints on the last row of both the chiplets virtual table and chiplet bus. With the wrong constraints, the ACE circuit will not evaluate to zero.

We introduce the `unconstrained` keyword to explicitely allow a certain bus boundary to be unconstrained. If we simply forget to put a constraint for a bus, we raise a diagnostic.

### DIFF
--- a/codegen/winterfell/src/air/boundary_constraints.rs
+++ b/codegen/winterfell/src/air/boundary_constraints.rs
@@ -128,7 +128,7 @@ fn add_aux_trace_assertions(func_body: &mut codegen::Function, ir: &Air) {
 
                     func_body.line(assertion);
                 }
-                air_ir::BusBoundary::Null => {}
+                air_ir::BusBoundary::Null | air_ir::BusBoundary::Unconstrained => {}
             }
         }
     }
--- a/parser/src/sema/semantic_analysis.rs
+++ b/parser/src/sema/semantic_analysis.rs
@@ -1390,8 +1390,11 @@ impl SemanticAnalysis<'_> {
                             };
 
                         match (found.clone().item, expr.rhs.as_mut()) {
-                            // Buses boundaries can be constrained by null
-                            (BindingType::Bus(_), ScalarExpr::Null(_)) => {}
+                            // Buses boundaries can be constrained by null or unconstrained
+                            (
+                                BindingType::Bus(_),
+                                ScalarExpr::Null(_) | ScalarExpr::Unconstrained(_),
+                            ) => {}
                             (BindingType::Bus(_), ScalarExpr::SymbolAccess(access)) => {
                                 self.visit_mut_resolvable_identifier(&mut access.name)?;
                                 self.visit_mut_access_type(&mut access.access_type)?;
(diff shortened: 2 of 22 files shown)


parser/src/sema/semantic_analysis.rs
```diff
@@ -1390,8 +1390,11 @@ impl SemanticAnalysis<'_> {
                             };
 
                         match (found.clone().item, expr.rhs.as_mut()) {
-                            // Buses boundaries can be constrained by null
-                            (BindingType::Bus(_), ScalarExpr::Null(_)) => {}
+                            // Buses boundaries can be constrained by null or unconstrained
```
Al-Kindi-0, line 1393: nit: "Buses boundaries can be constrained by null or set to be unconstrained"

parser/src/sema/semantic_analysis.rs
```diff
@@ -1438,18 +1439,18 @@ impl SemanticAnalysis<'_> {
                                     )
                                     .emit();
                             }
-                            (_, ScalarExpr::Null(_)) => {
-                                // Only buses can be constrained to null
+                            (_, ScalarExpr::Null(_) | ScalarExpr::Unconstrained(_)) => {
+                                // Only buses can be constrained to null or unconstrained
```
Al-Kindi-0, line 1443: nit: ".. or set to be unconstrained"

parser/src/sema/semantic_analysis.rs
```diff
@@ -1438,18 +1439,18 @@ impl SemanticAnalysis<'_> {
...
                                     expr.lhs.span(),
                                     "this constrains a column",
                                 )
                                 .with_secondary_label(
                                     expr.rhs.span(),
-                                    "but this expression is only valid to constrain buses",
+                                    "but this expression is only valid to for bus boundaries",
                                 )
-                                .with_note("The null value is only valid for defining empty buses")
+                                .with_note("The null / unconstrained value is only valid for defining empty buses")
```
Al-Kindi-0, line 1453: it should probably be ".. for constraining buses" instead

Al-Kindi-0 (approved):
Looks great, thank you!
Just a couple of nits/corrections

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -128,7 +128,7 @@ fn add_aux_trace_assertions(func_body: &mut codegen::Function, ir: &Air) {
 
                     func_body.line(assertion);
                 }
-                air_ir::BusBoundary::Null => {}
+                air_ir::BusBoundary::Null | air_ir::BusBoundary::Unconstrained => {}
```
bobbinth, line 131: We don't do anything for `Null` constraints here because they a part of the algebraic graph, right? If so, I'd add a comment to explain this.

bobbinth (approved):
Looks good! Thank you! I left one small doc-related comment inline. I'll merge this as and we can address this comment in future PRs.

Issue #399, opened 2025-05-14
Add handling of variable length public inputs to ACE codegen


Al-Kindi-0:
Given the current solution that we have settled for for handling variable length public inputs in the context of the recursive verifier, we should add support of this type of public inputs in the ACE [codegen](https://github.com/0xMiden/air-script/blob/1ea9fc6eb119d11a0b994ca35af76a3fe86569f5/codegen/ace/src/builder.rs#L139).

The inputs section, as part of the `READ` section, for the arithmetic circuit is now modified so that, in order and contiguously, we have:

1. fixed length public inputs. In the case of the recursive verifier, the I/O operand stack and the program digest,
2. the reduced values of the variable length public inputs. This means that we have one value, in the challenge field, per boundary constraint involving variable length public inputs. These values are stored next to each other and are word-aligned (with zero padding if needed),
3. the auxiliary randomness $\alpha$ and $\beta$, stored as one word, so that any bus message $m := (m_0, \cdots, m_{|m| - 1})$ can be hashed/fingerprinted as $\alpha + \sum_{i=0}^{|m|-1} \beta^i m_i$. Note that using two auxiliary random values is intentional as it is both the right way to present the (two) operations involved in the resulting expression, and also because it makes the soundness analysis cleaner.
4. the OOD evaluations as before without anything changed.

adr1anh:
This requires some modifications to the IR so we can represent the bus boundary as a separate `Value`. At the moment, these boundary constraints are handled by the codegen, which could lead to incompatible constraint polynomials between winterfell and ACE.

I'll create an issue for this, and hopefully have a PR ready early next week. A PR for this will follow.

PR #409, opened 2025-06-19
Extend ACE codegen with MASM code

Al-Kindi-0:
This is an initial version for defining MASM code accompanying the ACE circuit outputted by the ACE codegen.

There still a few points to close:

1. Variable length public inputs support needs to be add before we can test that part i.e., #399 needs to be addressed,
2. Only support of variable length public inputs to `multiset` buses is supported now, we should also add that for `logup`


Al-Kindi-0:
Updated according to the changes in https://github.com/0xMiden/miden-vm/pull/2008

Issue #387, opened 2025-05-02
Allow bus operations in libraries and evaluators


Leo-Besancon:
Currently libraries do not inherit bus declarations. We also don't currently handle them in evaluators. The discussion in https://github.com/0xMiden/air-script/pull/375#discussion_r2070939425 suggests we may not need to pass buses in the evaluator signature before using them.

Task list:
- [x] Ensure buses are accessible in libraries and evaluator bodies context
- [x] Adapt Mir passes handling buses (as for now we get bus operations in Mir during the lowering from the AST and before evaluator inlining, we may adapt things a bit.
- [x] Add tests and adapt documentation

PR #410, opened 2025-06-20
feat: Allow bus operations in libraries and evaluators

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/387

Buses are defined in the root (not allowed to be defined in another module), and are accessible in all modules and evaluators.

Task list:
- [x] Ensure buses are accessible in libraries and evaluator bodies context
- [x] Adapt Mir passes handling buses (as for now we get bus operations in Mir during the lowering from the AST and before evaluator inlining, we may adapt things a bit.
- [x] Add tests and adapt documentation

### DIFF
--- a/air/src/passes/translate_from_mir.rs
+++ b/air/src/passes/translate_from_mir.rs
@@ -75,10 +75,6 @@ impl Pass for MirToAir<'_> {
         let graph = mir.constraint_graph();
 
         // We insert all the constraints into the AIR graph.
-        // Note: In the MIR, buses operations are kept in integrity constraints to
-        // allow them to be handled in the graph (e.g. inlined via evaluators). This is why
-        // we need to first visit the integrity constraints, update the corresponding bus
-        // when encountering a BusOp, and then visit the buses to build them.
         for bc in graph.boundary_constraints_roots.borrow().deref().iter() {
             builder.build_boundary_constraint(bc)?;
         }
@@ -87,6 +83,10 @@ impl Pass for MirToAir<'_> {
             builder.build_integrity_constraint(ic)?;
         }
 
+        // Note: In the MIR, buses operations are kept in integrity constraints to
+        // allow them to be handled in the graph (e.g. inlined via evaluators). This is why
+        // we need to first visit the integrity constraints, update the corresponding bus
+        // when encountering a BusOp, and then visit the buses to build them.
         for bus in buses.values() {
             builder.build_bus(bus)?;
         }


air-script/tests/buses/buses_complex.air
```diff
@@ -25,8 +25,12 @@ integrity_constraints {
     enf s1^2 = s1;
     enf s2^2 = s2;
 
-    p.insert(1, a) when s1;
-    p.remove(1, b) when s2;
+    let vec = [x for x in 0..3];
+    let c = 41;
```
Al-Kindi-0, line 29: Q: is this line really needed?

mir/src/passes/inlining.rs
```diff
@@ -347,13 +347,37 @@ impl Visitor for InliningSecondPass<'_> {
...
+
+                        let new_node = self
+                            .nodes_to_replace
+                            .get(&body_node.get_ptr())
+                            .unwrap()
+                            .1
+                            .clone();
+
+                        if let Some(bus_op) = new_node.clone().as_bus_op() {
+                            //let args = bus_op.args.clone();
```
Al-Kindi-0, line 359: Should this be removed?

Al-Kindi-0 (approved):
Looks good to me (assuming I understood the referenced issue), thank you!

air-script/tests/buses/buses_complex.air
Leo-Besancon: It was not, thank you!

air/src/passes/translate_from_mir.rs
```diff
@@ -74,17 +74,18 @@ impl Pass for MirToAir<'_> {
...
             builder.build_boundary_constraint(bc)?;
         }
 
         for ic in graph.integrity_constraints_roots.borrow().deref().iter() {
             builder.build_integrity_constraint(ic)?;
         }
+
+        for bus in buses.values() {
+            builder.build_bus(bus)?;
+        }
```
bobbinth, line 92: Question: why did this have to move here? Seems like there may be some significance to the order in which we build different components. If so, I'd add some comments explaining this.

bobbinth (approved):
Looks good! Thank you! Not a super detailed review from me - but I left one small comment inline.

air/src/passes/translate_from_mir.rs
Leo-Besancon: Added a comment in f73dbb2b1c5d51761645403b8507869932e2f7c6, let me know if this is clearer!

Basically, the alternative if we want to avoid the ordering to matter is to add a tiny MIR pass after unrolling that just populates the BusOp in the corresponding buses. IMO it seemed simple enough to keep here, but I agree adding a comment explaining it is important.

air/src/passes/translate_from_mir.rs
```diff
@@ -74,6 +74,11 @@ impl Pass for MirToAir<'_> {
 
         let graph = mir.constraint_graph();
 
+        // We insert all the constraints into the AIR graph.
+        // Note: In the MIR, buses operations are kept in integrity constraints to
+        // allow them to be handled in the graph (e.g. inlined via evaluators). This is why
+        // we need to first visit the integrity constraints, update the corresponding bus
+        // when encountering a BusOp, and then visit the buses to build them.
```
bobbinth: nit: I'd maybe move this part of the comment to be above line 90 below.

air/src/passes/translate_from_mir.rs
bobbinth: Yes, I think adding a comment is sufficient. I left a small nit about comment location below.

air/src/passes/translate_from_mir.rs
Leo-Besancon: Indeed, done!

Issue #405, opened 2025-06-16
Remove the deprecated --without-mir pipeline


Leo-Besancon:
Now that the MIR compilation pipeline has more complete features (e.g. buses) which were not ported to the previous pipeline, we should remove the deprecated AST > AIR pipeline.

It entails the following tasks:
- [x] Remove the AST > AST Inlining pass
- [x] Remove the AstToAir pass
- [x] Remove WithoutMir CLI option
- [x] Remove tests done without the MIR

PR #413, opened 2025-06-27
Feat: Combine constraints generated by selector expressions

### LINKED ISSUE #332: Combine constraints generated by selector expressions
Currently, constraints generated for `enf match` expressions are not combined in any way. This results in a large number of constraints which will significantly impact verifier performance. There are two ways of combining constraints for `enf match`:

1. A relatively straight-forward way described in https://github.com/0xPolygonMiden/air-script/issues/125#issuecomment-1615503114.
2. A more sophisticated way which would give us a near-ideal result. This approach is described below.

### A simple example
Let's say we have the following `enf match` statement:
```
enf match:
    case c: op_add(s[0..3])
    case !c: op_mul(s[0..3])

ev op_add([s[3])
  enf s[0]' = s[0] + s[1]
  enf s[1]' = s[2]

ev op_mul([s[3])
  enf s[1]' = s[2]
  enf s[0]' = s[0] * s[1]
```
Currently, this will be reduced to 4 individual constraints:

$$
c \cdot (s_0' - s_0 - s_1) = 0
$$

$$
c \cdot (s_1' - s_2) = 0
$$

$$
(1 - c) \cdot (s_0' - s_0 \cdot s_1) = 0
$$

$$
(1 - c) \cdot (s_1' - s_2) = 0
$$

But we'd like to reduce it to two constraints which are much simpler:

$$
c \cdot (s_0' - s_0 - s_1) + (1 - c) \cdot (s_0' - s_0 \cdot s_1) = 0
$$

$$
(c + 1 - c) \cdot (s_1' - s_2) = 0
$$

It is easy to see that the second constraint is just $s_1' - s_2 = 0$, but we'll leave this optimization for later.

### General methodology
The key tool we'll use in this methodology is evaluation at a random point to test for equivalence between constraints. This works in our case because we deal with bounded-degree polynomials, and if we evaluate on a large enough domain, the probability that two distinct polynomials will evaluate to the same value is negligible. For us, I think degree 2 extension of our base field should be sufficiently large, but we can also go to degree 3 extension to be extra safe.

Let's say we have the following `enf match` statement:
```
enf match:
  s0: a, b, c, d
  s1: e, f, g
  s2: h, i
```
Where `s0`, `s1`, `s2` are selectors and `a, .., h` are individual constraints.

Let's also say that when evaluated at a random point, we get the following values for each constraint:
- `a`, `c`, and `f` evaluate to `r0`.
- `b` evaluates to `r1`
- `d` evaluates to `r2`
- `e` evaluates to `r3`
- `g`, `h` evaluate to `r4`.
- `i` evaluates to `r5`

Then, we start processing arms of `enf match` statement one-by-one, keeping track of aggregated results. We want to do the following two things:
1. Eliminate equivalent constraints in the same arm (e.g., `a` and `c`).
2. Establish links between equivalent constraints across different arms (e.g., `a` and `f`).

At the end of the process we want to end up with a data structure which records the following:
```
r0: (a, s0), (f, s1)
r1: (b, s0)
r2: (d, s0)
r3: (e, s1)
r4: (g, s1), (h, s2)
r5: (i, s2)
```
Then, we combine thee expressions together in a way that each constraint ends up having distinct selectors. Strategies for this could be different. One result could look like so:
```
(a, s0), (f, s1), (i, s2)
(b, s0), (e, s1)
(d, s0), (g, s1), (h, s2)
```
And then, in the end, we get 3 constraints which look like so:
```
a * (s0 + s1) + s2 * i = 0
s0 * b + s1 * e = 0
s0 * d + g * (s1 + s2) = 0
```
Notice that `f` and `h` are eliminated because they are the same constraints as `a` and `g` respectively.

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/332.

Task list:
- [x] Update parser to parse match as EnforceIf instead of EnforceAll (as list comprehensions). This aims to make it easier to optimize match constraints without handling list comprehensions.
- [x] Update MIR to create If nodes for EnforceIf statements
- [x] Figure out a way to get all branches of a given match in a single node? Otherwise we cannot optimize the constraints if we don't have a common parent to run the optimization on.
  For this point I'm considering updating:
    - EnforceIf to take a `Vec<MatchBranch>` (MatchBranch being a `(cond, expr)`)
    - If nodes in the MIR to take a similar structure. Then when we unroll them we can consider all conditions of the given match
- [x] Add helper function to evaluate a given (Mir) node at a random point
- [x] Evaluate all match constraints to random points during Unrolling
- [x] Optimize constraints using the method described in https://github.com/0xMiden/air-script/issues/332
- [x] Add test (using the same example as described in the issue, check that the resulting graph has the optimized constraints in an end to end test)
- [x] Sanity check the selector combine heuristic (it seems that current algo may not be optimal depending on how it chooses the next constraints to add)
- [x] Check the resulting graphs of E2E tests
- [x] Have the MIR node evaluator handle more cases (we don't handle many types of nodes that could be valid at this stage)
- [ ] Potential additional testing for node evaluations (check that we handle all the cases that can arise at that point)

Note: eval.rs has a manual implem in eval3.rs for 3 points. The manual implem is IMO more readable but less maintainable if we want to update the number of evaluations.


air-script/tests/selectors/selectors_combine.rs
```diff
@@ -82,9 +82,8 @@ impl Air for SelectorsAir {
     fn evaluate_transition<E: FieldElement<BaseField = Felt>>(&self, frame: &EvaluationFrame<E>, periodic_values: &[E], result: &mut [E]) {
         let main_current = frame.current();
         let main_next = frame.next();
-        result[0] = (main_next[3] - E::ZERO) * main_current[0] * (E::ONE - main_current[1]);
-        result[1] = (main_next[3] - main_current[3]) * main_current[0] * main_current[1] * main_current[2];
-        result[2] = (main_next[3] - E::ONE) * (E::ONE - main_current[1]) * (E::ONE - main_current[2]);
+        result[0] = (main_current[3] + E::ONE - main_current[3]) * (main_next[1] - main_current[2]) - E::ZERO;
```
adr1anh: Shouldn't the `- E::ZERO` be optimized out?

mir/src/ir/eval.rs
```diff
@@ -0,0 +1,308 @@
...
+
+pub const NUM_EVALS: usize = 3; // Number of random evaluation points
+
+/// A struct representing the evaluation of an operation at `NUM_EVALS` random points.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Eval<const NUM_EVALS: usize>(pub [Felt; NUM_EVALS]);
+
+impl<const NUM_EVALS: usize> Eval<NUM_EVALS> {
+    fn exp(self, other: Self) -> Self {
+        Eval(
```
adr1anh: We can write this more succinctly as this (assuming `Index` is implemented) 
```rust
Eval(array::from_fn(|i| self[i].exp(other[i].as_int())))
```
The other arithmetic implementations can do the same

mir/src/ir/eval3.rs
```diff
@@ -0,0 +1,258 @@
+use std::{
```
adr1anh: This file should be removed right?

mir/src/passes/unrolling.rs
```diff
@@ -487,49 +456,149 @@ impl UnrollingFirstPass<'_> {
 
         {
```
adr1anh: What is the point of the braces?

mir/src/passes/unrolling.rs
```diff
@@ -487,49 +456,149 @@ impl UnrollingFirstPass<'_> {
...
-            let else_branch = if_ref.else_branch.clone();
-
-            let mut new_vec = vec![];
-
-            if let Op::Vector(then_branch_vector) = then_branch.clone().borrow().deref() {
-                let then_branch_vec = then_branch_vector.children().borrow().deref().clone();
+            let match_arms = if_ref.match_arms.borrow();
+
+            // 1. We evaluate all constraints of this match node at random points
+            let mut current_evals = CurrentEvals::default();
```
adr1anh: Would there be any benefit to making this part of the struct so we can cache evaluations across the entire pass?

parser/src/ast/statement.rs
```diff
@@ -64,6 +64,67 @@ pub enum Statement {
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "match: ")?;
+        for arm in self.match_arms.iter() {
+            write!(f, "{arm}")?;
+        }
+        Ok(())
+    }
+}
+
+#[derive(Clone, Spanned, Debug)]
```
adr1anh: We do not need to implement `Eq` manually
```suggestion
#[derive(Clone, Spanned, Debug, Eq)]
```

parser/src/ast/statement.rs
```diff
@@ -64,6 +64,67 @@ pub enum Statement {
...
+    pub expr: ScalarExpr,
+}
+
+impl MatchArm {
+    pub fn new(span: SourceSpan, expr: ScalarExpr, condition: ScalarExpr) -> Self {
+        Self { span, expr, condition }
+    }
+}
+
+impl Eq for MatchArm {}
```
adr1anh: Not needed

parser/src/transforms/constant_propagation.rs
```diff
@@ -679,7 +679,15 @@ impl VisitMut<SemanticAnalysisError> for ConstantPropagation<'_> {
                     self.in_constraint_comprehension = false;
                 },
                 // This statement type is only present in the AST after inlining
-                Statement::EnforceIf(..) => unreachable!(),
+                Statement::EnforceIf(match_expr) => {
+                    self.in_constraint_comprehension = true;
+                    for match_arm in match_expr.match_arms.iter_mut() {
+                        // Visit the expression and selector of the match arm
+                        self.visit_mut_scalar_expr(&mut match_arm.expr)?;
```
adr1anh: Nit: isn't it more intuitive to visit the condition before the expr?

parser/src/ast/statement.rs
```diff
@@ -64,6 +64,67 @@ pub enum Statement {
     /// Declares a bus related constraint
     BusEnforce(ListComprehension),
 }
+
+#[derive(Clone, Spanned, Debug)]
```
adr1anh:
```suggestion
#[derive(Clone, Spanned, Debug, Eq)]
```

mir/src/passes/translate.rs
```diff
@@ -423,10 +423,18 @@ impl<'a> MirBuilder<'a> {
...
         &mut self,
-        _enf: &ast::ScalarExpr,
-        _cond: &ast::ScalarExpr,
+        match_expr: &'a ast::Match,
     ) -> Result<Link<Op>, CompileError> {
-        unreachable!("all EnforceIf should have been transformed into EnforceAll")
+        let mut match_arms = Vec::new();
+
+        for match_arm in match_expr.match_arms.iter() {
+            let expr_node = self.translate_scalar_expr(&match_arm.expr)?;
```
adr1anh: Nit: `condition` before `expr`?

mir/src/ir/eval.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+/// Helper function to either query an existing evaluation or create a new random one if the element
+/// is not present in the map.
+fn query_hashed_cur_eval<R: Rng + ?Sized, H: Hash + Eq + Clone>(
+    rng: &mut R,
+    cur_eval_map: &mut HashMap<H, Eval<NUM_EVALS>>,
+    element: &H,
+) -> Eval<NUM_EVALS> {
+    if cur_eval_map.get(element).is_none() {
```
adr1anh:
```suggestion
    *cur_eval_map.entry(element.clone()).or_insert_with(|| {
        Eval::new_random(rng)
    })
```

mir/src/ir/eval.rs
```diff
@@ -0,0 +1,308 @@
...
+        ConstantValue, Link, MirValue, Op, PeriodicColumnAccess, PublicInputAccess,
+        PublicInputTableAccess,
+    },
+};
+
+pub const NUM_EVALS: usize = 3; // Number of random evaluation points
+
+/// A struct representing the evaluation of an operation at `NUM_EVALS` random points.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Eval<const NUM_EVALS: usize>(pub [Felt; NUM_EVALS]);
```
adr1anh: We could implement `Index`:

```rust
impl<const NUM_EVALS: usize> Index<usize> for Eval<NUM_EVALS> {
    type Output = Felt;

    fn index(&self, index: usize) -> &Self::Output {
        self.0.index(index)
    }
}
```

mir/src/ir/eval.rs
```diff
@@ -0,0 +1,308 @@
...
+                .try_into()
+                .unwrap(),
+        )
+    }
+
+    /// Creates a new [Eval] with the same constant value for all evaluations.
+    fn new_const(felt: Felt) -> Self {
+        Eval([felt; NUM_EVALS])
+    }
+}
```
adr1anh: Consider adding 
```rust
    fn as_ints(&self) -> [u64; NUM_EVALS] {
        self.0.each_ref().map(Felt::as_int)
    }
```

mir/src/ir/eval.rs
```diff
@@ -0,0 +1,308 @@
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        for felt in &self.0 {
+            felt.as_int().hash(state);
+        }
+    }
+}
+
+impl<const NUM_EVALS: usize> Ord for Eval<NUM_EVALS> {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.0
```
adr1anh: Consider 
```rust
        self.as_ints().cmp(&other.as_ints())
```

air-script/tests/selectors/selectors_combine.rs
adr1anh: Never mind, I saw the later comment. Do we have an issue for this?

adr1anh (commented):
Nice work! Left some small nits to make things clearer. A more general nit is whether we should keep consistency with `condition -> expr`.

air/src/passes/translate_from_ast.rs
```diff
@@ -102,12 +102,9 @@ impl AirBuilder<'_> {
...
                 ..
             })) => self.build_integrity_equality(lhs, rhs, None),
-            ast::Statement::EnforceIf(
-                ast::ScalarExpr::Binary(ast::BinaryExpr {
-                    op: ast::BinaryOp::Eq, lhs, rhs, ..
-                }),
-                condition,
-            ) => self.build_integrity_equality(lhs, rhs, Some(condition)),
+            ast::Statement::EnforceIf(_) => {
+                unreachable!("`EnforceIf` not supported on this Pipeline")
```
bobbinth: nit: "on this Pipeline" -> "in this pipeline".

Also, it would be great to add a brief comment explaining why. My understanding is that by the time we get here `EnforceIf` (which designate `match` expressions) have already been reduced to more primitive expressions - correct?

air/src/tests/selectors.rs
```diff
@@ -18,7 +19,11 @@ fn single_selector() {
         enf clk' = clk when s[0];
     }";
 
-    assert!(compile(source, Pipeline::WithoutMIR).is_ok());
+    expect_diagnostic(
+        source,
+        "matches are not implemented for this Pipeline",
+        Pipeline::WithoutMIR,
+    );
```
bobbinth: Would it make sense to first address https://github.com/0xMiden/air-script/issues/405 so that we don't have to make these updates here? (but if https://github.com/0xMiden/air-script/issues/405 is not trivial, we can do it later as well).

mir/src/passes/unrolling.rs
bobbinth, line 1: Probably not for this PR, but this file is quite large with functions that have very deep nesting - and so, it is not easy to follow. We should try to split it up into multiple simpler files (still under `unrolling` module) and maybe have more helper functions that are called form `match` statements.

mir/src/passes/unrolling.rs
bobbinth: If we make `CurrentEvals` a part of the struct, we could probably put the RNG inside of it. This way, we may be able to avoid passing around `rng` as a separate parameter.

mir/src/passes/unrolling.rs
```diff
@@ -487,49 +456,149 @@ impl UnrollingFirstPass<'_> {
...
+            for match_arm in match_arms.iter() {
+                let condition = match_arm.condition.clone();
+                let expr = match_arm.expr.clone();
+
+                // Get all the individual constraints corresponding to this arm
+                let constraints_to_eval = if let Op::Vector(expr_vector) = expr.borrow().deref() {
+                    expr_vector.children().borrow().deref().clone()
+                } else {
+                    vec![expr.clone()]
+                };
```
bobbinth: Question: if we are treating `Op::Vector` here separately, do we also need to do the same for `Op::Matrix`?

mir/Cargo.toml
```diff
@@ -18,4 +18,6 @@ anyhow = { workspace = true }
 derive-ir = { package = "air-derive-ir", path = "./derive-ir", version = "0.5" }
 miden-diagnostics = { workspace = true }
 pretty_assertions = "1.4"
-thiserror = { workspace = true }
\ No newline at end of file
+thiserror = { workspace = true }
+rand = "0.9"
+winter-math = { package = "winter-math", version = "0.12", default-features = false }
```
bobbinth: nit: I would combine these with other dependencies and alphabetize them.

mir/src/ir/eval.rs
```diff
@@ -0,0 +1,308 @@
...
+
+impl<const NUM_EVALS: usize> PartialOrd for Eval<NUM_EVALS> {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// Represents the current existing evaluations to persist random values taken by the same values.
+#[derive(Debug, Clone, Default)]
+pub struct CurrentEvals<const NUM_EVALS: usize> {
```
bobbinth: I found the name `CurrentEvals` a bit confusing. These feel more like random inputs rather then evaluations. Should the name be `RandomInputs` or something like that?

mir/src/ir/eval.rs
bobbinth: We are keeping a vector here to increase soundness, right? If so, a simpler approach could be to keep a single value in the extension field. For example, using an element in the quadratic (or if we really want, cubic - though, not sure this is needed) extension field should give us the same soundness and should simplify the code.

mir/src/ir/eval.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+    *cur_eval_map.get(element).unwrap()
+}
+
+/// Evaluates a given MIR node at random points.
+pub fn eval_random_point<R: Rng + ?Sized>(
+    rng: &mut R,
+    current_evals: &mut CurrentEvals<NUM_EVALS>,
+    op: Link<Op>,
+) -> Result<Eval<NUM_EVALS>, CompileError> {
```
bobbinth: If we move `rng` into `CurrentEvals` - this could be just a method on `CurrentEvals`, right? If we also use extension field element (instead of array) and do some of the renaming mentioned in the previous comments, this could look something like:
```rust
impl RandomInputs {
    pub fn eval(&mut self, op: Link<Op>) -> Result<QuadFelt, CompileError> {
        ...
    }
}
```

mir/src/ir/eval.rs
```diff
@@ -0,0 +1,308 @@
...
+            let rhs = eval_random_point(rng, current_evals, e.rhs.clone())?;
+            Ok(lhs.exp(rhs))
+        },
+        Op::BusOp(b) => {
+            let mut hasher = DefaultHasher::new();
+            "BusOp".hash(&mut hasher);
+            b.hash(&mut hasher);
+            let hash = hasher.finish();
+            let felt = Felt::new(hash);
+            Ok(Eval::new_const(felt))
```
bobbinth: Could we add a comment about what we are doing here?

bobbinth (commented):
Thank you! Looks good. Not a full review, but I left some comments inline.

Also, would be great to add a more sophisticated test (in addition to the simple `selectors_combine.air`).

Lastly, we may want to remove the "without MIR" pipeline in a separate PR and then rebase this based on that. This should reduce the footprint of this PR a bit.

bobbinth (commented):
Thank you! Looks good. Not a full review, but I left some comments inline.

Also, would be great to add a more sophisticated test (in addition to the simple `selectors_combine.air`).

Lastly, we may want to remove the "without MIR" pipeline in a separate PR and then rebase this based on that. This should reduce the footprint of this PR a bit.

air/src/passes/translate_from_ast.rs
Leo-Besancon: Before this PR, EnforceIf contained only one match arm. I've updated this variant to handle all the match arms for a given match.
Because this pipeline will be removed, I didn't want to bother adapting the existing code to this new structure, but there is no reason preventing it, in the deprecated inlining match statements where reduced to primitive EnforceIf we could handle.

I'll remove the without-mir pipeline in a separate PR and rebase like you suggested.

mir/src/passes/unrolling.rs
Leo-Besancon: This correspond to the resulting constraints for a given match arm.
It could be a single element (`case c: enf a = b`) or a vector of constraints if it's the result of an evaluator (`case c: op_add([s[0..3]])`), but I don't think it can be a Matrix for now.

air/src/tests/selectors.rs
Leo-Besancon: I think it does make sense as it will have to be done at some point and it simplifies this PR.

I opened the PR here: https://github.com/0xMiden/air-script/pull/424

air-script/tests/selectors/selectors_combine.rs
Leo-Besancon: This should be handle in the PR introducing Common Sub-Expression Elimination: https://github.com/0xMiden/air-script/pull/418 (note that it has not yet been updated based on the comments of this PR for Eval handling, etc.)

mir/src/ir/eval.rs
Leo-Besancon: Thank you for the trick! I didn't know the `array::from_fn` function, it's pretty nice that typing inference auto detects how many items it should generate!

Note: As suggested by @bobbinth below, as I use Quad extension instead of multiple evaluation points, this is no longer needed.

mir/src/ir/eval3.rs
Leo-Besancon: Thank you, removed.

mir/src/passes/unrolling.rs
Leo-Besancon: It was useful in other cases for the borrow checker, but it's useless in this cases, I've removed them.

mir/src/passes/unrolling.rs
Leo-Besancon: Done

parser/src/ast/statement.rs
Leo-Besancon: Thank you, removed the Eq implementations here and below

parser/src/transforms/constant_propagation.rs
Leo-Besancon: Why not, I've refactor here and below

mir/src/ir/eval.rs
Leo-Besancon: Thank you, used this suggestion in the quad_eval.rs file

mir/src/ir/eval.rs
Leo-Besancon: No longer needed with Quad extensions

mir/src/ir/eval.rs
Leo-Besancon: Implemented these two suggestions

mir/src/passes/unrolling.rs
Leo-Besancon: True, I've created the issue here: https://github.com/0xMiden/air-script/issues/425

mir/src/ir/eval.rs
Leo-Besancon: Indeed, I've made the switch in the quad_eval.rs file.

mir/src/ir/eval.rs
Leo-Besancon: Renamed to RandomInputs and added it as a method

Leo-Besancon:
@adr1anh and @bobbinth thank you for your reviews, you raised many good points! I should have resolved everything.

I've added a `selectors_combine_complex` test that for now only aims to reproduce the general example showcased in https://github.com/0xMiden/air-script/issues/332#issue-1784289797. The produced graph does correspond to the 3 constraints in the linked comment. I'm not sure If we have more complex cases (e.g. with nested match), I will think about it.

I've put a new PR for the WithoutMir pipeline removal (a lot of changes but pretty straightforward), will rebased when it is merged: https://github.com/0xMiden/air-script/pull/424

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,268 @@
...
+use crate::{
+    CompileError,
+    ir::{
+        ConstantValue, Link, MirValue, Op, PeriodicColumnAccess, PublicInputAccess,
+        PublicInputTableAccess,
+    },
+};
+
+#[derive(Eq, PartialEq, Clone, Debug, Copy)]
+pub struct QuadFelt(pub QuadExtension<Felt>);
```
bobbinth: nit: do we actually need to create this wrapper? We have a `QuadFelt` type alias defined somewhere in this repo - we may be able to use it directly. This may require creating small helper functions like `rand_quad_felt()` and `const_quad_felt()` but at the same time, we should be able to get rid this struct and all the op overrides (`Add`, `Sub`, and `Mul`).

But if this does cause complications - it is fine to keep this wrapper struct too.

Leo-Besancon: Seems good. The only thing I'm leaving for later is a direct dependency of mir crate in ace. But after https://github.com/0xMiden/air-script/pull/418 it may be re-exported by air-ir crate so it may be cleaner.

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,212 @@
...
+            },
+            Op::Mul(m) => {
+                let lhs = self.eval(m.lhs.clone())?;
+                let rhs = self.eval(m.rhs.clone())?;
+                Ok(lhs * rhs)
+            },
+            Op::Exp(e) => {
+                let lhs = self.eval(e.lhs.clone())?;
+                let rhs = self.eval(e.rhs.clone())?;
+                let power = rhs.to_base_elements()[0].as_int();
```
Al-Kindi-0: nit: should we test that the second component is ZERO?

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,212 @@
...
+            },
+            Op::Value(v) => {
+                match &v.value.value {
+                    MirValue::Constant(ConstantValue::Felt(c)) => {
+                        let felt = Felt::new(*c);
+                        Ok(const_quad_felt(felt))
+                    },
+                    MirValue::TraceAccess(trace_access) => match trace_access.segment {
+                        0 => {
+                            let index = trace_access.column * 2 + trace_access.row_offset;
```
Al-Kindi-0, line 102: nit: would be good to add a comment on how the indexing works i.e., the layout of what we are indexing into

mir/src/passes/unrolling.rs
```diff
@@ -437,105 +446,155 @@ impl UnrollingFirstPass<'_> {
...
-                    new_vec.push(new_node);
-                }
-                if_node.set(&Vector::create(new_vec));
-            }
-        };
-
-        Ok(())
-    }*/
-
     fn visit_if_bis(
```
Al-Kindi-0, line 449: should this be just `visit_if`?

Al-Kindi-0 (commented):
This is an initial review from my part, but looks good, thank you!
Would it make sense to review something else before going deeper into this PR? I think #424 is one of these, but are there others?

Leo-Besancon:
> This is an initial review from my part, but looks good, thank you! Would it make sense to review something else before going deeper into this PR? I think #424 is one of these, but are there others?

Thank you! No, looking into #424 and then this one is good!
The only other related PR is #418, but as it also uses random evaluation, I'll wait for all comments on this PR before rebasing and replicating the changes there.

mir/src/passes/unrolling.rs
Leo-Besancon: In the Unrolling pass, we implemented `visit_*_bis` methods to avoid coliding with `visit_*` methods of the [default Visitor trait](https://github.com/0xMiden/air-script/blob/next/mir/src/passes/visitor.rs#L114), which directly mutates the visited node but would lead to double borrow errors here.

In our case, the `_bis` methods return the wanted updated value for the node so it can be [set later](https://github.com/0xMiden/air-script/blob/next/mir/src/passes/unrolling.rs#L945). 
This probably can be improved, and could maybe be one of the things that are cleaned up in the following issue: https://github.com/0xMiden/air-script/issues/425

mir/src/ir/quad_eval.rs
Leo-Besancon: Improved by returning an Err if it is not

mir/src/ir/quad_eval.rs
Leo-Besancon: Sure, I've added a comment, can you tell me if it's clearer now?

Al-Kindi-0: Thank you

mir/src/passes/unrolling.rs
```diff
@@ -437,105 +446,155 @@ impl UnrollingFirstPass<'_> {
...
             }
 
-            let one_constant = SpannedMirValue {
-                span: Default::default(),
-                value: MirValue::Constant(ConstantValue::Felt(1)),
-            };
-
-            if let Op::Vector(else_branch_vector) = else_branch.clone().borrow().deref() {
-                let else_branch_vec = else_branch_vector.children().borrow().deref().clone();
+            // Remove the taken evaluation indices from all the structures
```
Al-Kindi-0: nit: remove from all structures

Al-Kindi-0:
@Leo-Besancon , in expressions of the form
```
enf match{
 case selector0: arm0,
 case selector1: arm1,
 case selector2: arm2,
}
```
Ensuring that the selectors are mutually exclusive, i.e., sum is equal to 1,  or that at least only one is active at a time, i.e., sum is less than or equal to 1, is left to the user to enforce.
As you might know,  the optimizations in this PR will break things if the above is not enforced.
Edit: Even without the optimizations, we would want to use `enf match`  only we the selectors satisfy the above conditions.
Otherwise, things look good to me, thank you!

Leo-Besancon:
> @Leo-Besancon , in expressions of the form
> 
> ```
> enf match{
>  case selector0: arm0,
>  case selector1: arm1,
>  case selector2: arm2,
> }
> ```
> 
> Ensuring that the selectors are mutually exclusive, i.e., sum is equal to 1, or that at least only one is active at a time, i.e., sum is less than or equal to 1, is left to the user to enforce. As you might know, the optimizations in this PR will break things if the above is not enforced. Edit: Even without the optimizations, we would want to use `enf match` only we the selectors satisfy the above conditions. Otherwise, things look good to me, thank you!

Indeed, in the [documentation](https://0xmiden.github.io/air-script/description/convenience.html), it is described as:

> AirScript makes the following assumptions about selector expressions, which are not yet enforced by the language:
> 1. All selector expressions are based on binary values. To enforce these, we must manually add constraints of the form x^2=x for  all values involved in selector expressions.
> 2. All selector expressions are mutually exclusive. That is, for a given set of inputs, only one of the selector expressions in an enf > match statement can evaluate to 1, and all other selectors must evaluate to 0. Note: it is OK if all selector expressions evaluate to 0.

I think that we can still be flexible and allow a sum equal to 0 after this PR.

But I think we could, in a follow-up PR, enforce that this holds and fail the compilation with an error message if it doesn't.
Especially once the type system is up, we would be able to enforce that selectors are composed of binary expressions, and we could try all possible evaluations of the truth table to check by bruteforce it holds. Should I create an issue for this?

Al-Kindi-0:
> But I think we could, in a follow-up PR, enforce that this holds and fail the compilation with an error message if it doesn't.
Especially once the type system is up, we would be able to enforce that selectors are composed of binary expressions, and we could try all possible evaluations of the truth table to check by bruteforce it holds. Should I create an issue for this?

That would be great!

mir/src/passes/unrolling.rs
```diff
@@ -437,105 +446,155 @@ impl UnrollingFirstPass<'_> {
...
+                    equivalent_constraints.first().unwrap().1.clone(),
+                    if_ref.span(),
                 );
-                new_vec.push(new_node);
+
+                cur_node = if let Some(cur_node) = cur_node {
+                    Some(Add::create(cur_node, new_node, if_ref.span()))
+                } else {
+                    Some(new_node)
+                };
```
Al-Kindi-0: nit: I would write this as

```rust
cur_node = match cur_node {
    Some(existing) => Some(Add::create(existing, new_node, if_ref.span())),
    None => Some(new_node),
};
```

Al-Kindi-0 (approved):
LGTM, thank you!

air/src/tests/selectors.rs
Leo-Besancon: I've just rebased on next!

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,227 @@
...
+                            Ok(query_indexed_cur_eval(&mut self.rng, &mut self.main_trace, index))
+                        },
+                        1 => {
+                            let index = trace_access.column * 2 + trace_access.row_offset;
+                            Ok(query_indexed_cur_eval(&mut self.rng, &mut self.aux_trace, index))
+                        },
+                        _ => {
+                            println!(
+                                "Unexpected segment in eval_random_point: {}",
```
bobbinth: nit: the function name should probably be updated here. This is also applicable to a few more instances below.

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,227 @@
...
+            | Op::Fold(_)
+            | Op::Boundary(_)
+            | Op::For(_)
+            | Op::If(_)
+            | Op::Vector(_)
+            | Op::Matrix(_)
+            | Op::None(_) => {
+                // These operations are not handled in this function
+                println!("Unexpected operation in eval_random_point: {op:?}");
+                Err(CompileError::Failed)
```
bobbinth: What are the implications of this? For example, what if we have a `match` expression where one of the arms uses an evaluator and this evaluator calls a pure function - would this result in an error now?

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,227 @@
...
+            Op::BusOp(b) => {
+                // For a given bus operation, we hash the operation and its parameters to create a
+                // unique identifier for this operation in the extension field,
+                // as we haven't expanded the bus constraints yet.
+                let mut hasher = DefaultHasher::new();
+                "BusOp".hash(&mut hasher);
+                b.hash(&mut hasher);
+                let hash = hasher.finish();
+                let felt = Felt::new(hash);
+                Ok(const_quad_felt(felt))
```
bobbinth: I'm still not sure I fully understand the implication of this. What if `match` arms use evaluators which have bus expressions in them - does this mean that buss expressions won't be evaluated at random points?

air-script/tests/selectors/selectors_combine_complex.air
```diff
@@ -0,0 +1,45 @@
...
+ev ev_s1([c]) {
+    enf c = R[3]; #e
+    enf c = R[0]; #f
+    enf c = R[4]; #g
+}
+
+ev ev_s2([c]) {
+    enf c = R[4]; #h
+    enf c = R[5]; #i
+}
```
bobbinth: Could we make the evaluators here a bit more complex. For example:

- We could have some bus expressions in evaluators.
- We could call pure functions within evaluators.

bobbinth (commented):
Thank you! Looks good! I reviewed almost everything (one exception is the `unrolling.rs` file) and left a few comments/questions inline.

mir/src/ir/quad_eval.rs
Leo-Besancon: Currently, I think this means that bus expressions would only be "evaluated" to the same value if they are exactly the same, for instance, the first two expressions would be the same but different from the third:
```air
p.insert(2 + 2) when s
p.insert(2 + 2) when s
p.insert(4) when s
```

But thinking about it, I think there is nothing preventing us from evaluating the arguments and the selectors of the bus at random points to be able to reduce them,
I'll add more complicated tests as you suggested below and see what I can do!

mir/src/ir/quad_eval.rs
Leo-Besancon: I've added some comments regarding this. I'll add a test later for this, but as we call this during Unrolling, evaluators and functions have already been inlined.
Depending on other use cases we want to use eval() for, we may expand to other operation types, it should not be an issue, but there was no need for that right now.

mir/src/ir/quad_eval.rs
Leo-Besancon: Coming back to it, I don't think we can have bus expression in match arms currently.
Do we want to handle them and combine the match condition and the selector?
```
enf match {
        case s0: p.insert(1, 2) when s1,
}
```
Should this be treated as `p.insert(1, 2) when s1 & s0`, or should we disallow that?

If we allow this, I think we have filter them out of selector combination, we cannot mix regular constraints and bus constraints in the same expression.

air-script/tests/selectors/selectors_combine_complex.air
Leo-Besancon: I've updated the test with both of these!
For bus operation, I handle them separately and combine the latch with the match statement condition.

air-script/tests/selectors/selectors_combine_complex.air
```diff
@@ -0,0 +1,64 @@
...
+ev ev_s1([c]) {
+    enf c = R[3]; #e
+    enf c = 16; #f
+    enf c = R[4]; #g
+}
+
+ev ev_s2([c]) {
+    enf c = R[4]; #h
+    enf c = R[5]; #i
+}
```
bobbinth: I know this is meant to be just a test, but most of the constraints here are contradictory. Could we make them a bit more "realistic"? Also, could we add bus operations in all the evaluators. Maybe something like this will work? (though I didn't check if this covers all cases you wanted to cover)
```
ev ev_s0([a, b, c]) {
    enf a = double_with_mult(R[0]); #a
    enf b = R[1]; #b
    enf c = R[2]; #d

    p.insert(1, 2) when c;
}

ev ev_s1([a, b, c]) {
    enf a = 16; #f
    enf b = R[3]; #e
    enf c = R[4]; #g

    p.insert(1, 2) when c;
}

ev ev_s2([a, b, c]) {
    enf a = double_with_add(R[0]); #c
    enf b = R[4]; #h
    enf c = R[5]; #i

    p.remove(3, 4) when b;
}
```
The above implies that the main trace has been declared as something like `main: [s[3], a, b, c]`.

-------

Separately, and not related to this PR, but this made me think that we may want to allow bus operations without the `when` clause. For example, we may want to write something like:
```
ev ev_s2([a, b, c]) {
    ...

    p.remove(3, 4);
}
```
Here, we'd have the implicit selector applied to the bus operation because `ev_s2` is invoked form a `match` expression - but I think that's currently not possible, right? If so, let's create an issue for this.

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,217 @@
...
+                } else {
+                    let power = rhs.to_base_elements()[0].as_int();
+                    Ok(lhs.exp(power))
+                }
+            },
+            Op::Parameter(_) => {
+                // We cannot easily detect that two parameters refer to the same For, so we consider
+                // them to be all different
+                Ok(rand_quad_felt(&mut self.rng))
+            },
```
bobbinth, line 86: What are the implications of this? My understanding is that this is relevant to list comprehensions - right? If so, could we unroll list comprehensions before we do this pass so that we don't need to handle this case?

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,217 @@
...
+                    // this trace access, but also for all previous indices.
+                    MirValue::TraceAccess(trace_access) => match trace_access.segment {
+                        0 => {
+                            let index = trace_access.column * 2 + trace_access.row_offset;
+                            Ok(query_indexed_cur_eval(&mut self.rng, &mut self.main_trace, index))
+                        },
+                        1 => {
+                            let index = trace_access.column * 2 + trace_access.row_offset;
+                            Ok(query_indexed_cur_eval(&mut self.rng, &mut self.aux_trace, index))
+                        },
```
bobbinth: This handles constraints for auxiliary segments, right? If so, is this really needed? My understanding was that the bus constraints are the only way to have constraints against the auxiliary segment, and these are handled separately.

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,217 @@
...
+                        let felt = Felt::new(*c);
+                        Ok(const_quad_felt(felt))
+                    },
+                    // For each trace segment, we associate a random value to the each trace access
+                    // indexed in the following way, each column having two
+                    // distinct evaluations to account for the two possible row offsets:
+                    // $main[0], $main[0]', $main[1], $main[1]', $main[2], ...
+                    // Note: if we encounter a trace access corresponding to an index we have not
+                    // yet evaluated, we will randomly generate values for
+                    // this trace access, but also for all previous indices.
```
bobbinth: I would also probably add a brief explanation of this next to the struct field (i.e., line 35 above).

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,217 @@
...
+                        Ok(query_hashed_cur_eval(&mut self.rng, &mut self.public_inputs, pi))
+                    },
+                    MirValue::PeriodicColumn(pc) => {
+                        Ok(query_hashed_cur_eval(&mut self.rng, &mut self.periodic_columns, pc))
+                    },
+                    MirValue::PublicInputTable(pita) => Ok(query_hashed_cur_eval(
+                        &mut self.rng,
+                        &mut self.public_input_tables,
+                        pita,
+                    )),
```
bobbinth: Could we explain what the implication of using `query_hashed_cur_eval()` is for these? I'm assuming that, for example, for a public input, we'd create one random value per input (because we'd use input name in the hash), right?

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,217 @@
...
+                    MirValue::Null
+                    | MirValue::BusAccess(_)
+                    | MirValue::Unconstrained
+                    | MirValue::TraceAccessBinding(_)
+                    | MirValue::Constant(_) => {
+                        // These values are not handled in this function
+                        println!("Unexpected values in RandomInputs::eval: {op:?}");
+                        Err(CompileError::Failed)
```
bobbinth: Could we explain why these values are not handled here? (I'm assuming this is because when this function is called, these values have already been unrolled and therefore shouldn't be in the graph).

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,217 @@
...
+            | Op::BusOp(_)
+            | Op::For(_)
+            | Op::If(_)
+            | Op::Vector(_)
+            | Op::Matrix(_)
+            | Op::None(_) => {
+                // These operations are not handled in this function, as we currently expect this
+                // function to be called only on during Unrolling.
+                println!("Unexpected operation in RandomInputs::eval: {op:?}");
+                Err(CompileError::Failed)
```
bobbinth: Similar comment as above - if we expect all the above operations to have been already unrolled by this point, I'd say this explicitly.

mir/src/ir/quad_eval.rs
```diff
@@ -0,0 +1,217 @@
...
+                                    index,
+                                ));
+                            },
+                            1 => {
+                                return Ok(query_indexed_cur_eval(
+                                    &mut self.rng,
+                                    &mut self.aux_trace,
+                                    index,
+                                ));
+                            },
```
bobbinth: Similar to one of the above comments: are we actually expecting to handle auxiliary trace constraints here?

mir/src/passes/unrolling.rs
```diff
@@ -437,105 +446,195 @@ impl UnrollingFirstPass<'_> {
...
-                    new_vec.push(new_node);
-                }
-                if_node.set(&Vector::create(new_vec));
-            }
-        };
-
-        Ok(())
-    }*/
-
     fn visit_if_bis(
```
bobbinth, line 449: Question: what does `bis` stand for?

Separately, I would probably create a helper struct to improve the readability of this function. We could instantiate this struct with a reference to `self.random_inputs` and then have methods on it to help with various stages of the algorithm (e.g., `add_match_arm()`, `reduce_constraints()` - just off the top of my head, may be a better way to break things up).

We could address it as a part of https://github.com/0xMiden/air-script/issues/425, if you think that's a more appropriate way to do it.

bobbinth (commented):
Looks good! Thank you. I left some more comments inline.

air-script/tests/selectors/selectors_combine_complex.air
Leo-Besancon: > I know this is meant to be just a test, but most of the constraints here are contradictory. 

True, I've updated with your example after adapting it a bit to handle everything I wanted (e.g. I do need some duplicate constraints in one arm to ensure they are optimized out).

I've also fixed an issue with undeterministic iteration of bus related constraints, putting them in a Vec instead of an HashMap to have reproducible tests.

The resulting graph makes sense to me (for the combined constraints as well as the bus related constraints), but it would be great if you can check them too!

> but I think that's currently not possible, right? If so, let's create an issue for this.

I've created an issue here: https://github.com/0xMiden/air-script/issues/430

Do you think it's suitable to have `p.remove(3, 4);` be handled the same as `p.remove(3, 4) when 1;`?
This way, if it's inside a match, we can handle it by multiplying the two selectors as we do now.

But I'm not sure if it's clean enough (for instance, we may want to warn explicitely that no selectors have been set if we use a bus operation with no selector outside a match).

mir/src/ir/quad_eval.rs
Leo-Besancon: This is a bit tricky. When we handle this If node (in UnrollingFirstPass), For nodes have been transformed into: `Vector<Parameter(0, for_node), ..., Parameter(n, for_node)>`.

It's only later, during UnrollingSecondPass, that we replace each `Parameter(i, for_node)` parameter with the correct value (the body of the corresponding for_node evaluated at index `i`).

The implication is that constraints that rely on list comprehensions would indeed always have different .eval() values, we would not be able to combine and optimize them.


The best way to handle this is I think to keep If nodes the same during UnrollingFirstPass, then unroll the list comprehensions in UnrollingSecondPass, and then handling If nodes in a third pass.

I'll just have to think better about:
- how list comprehensions and match statements can interact
- If moving the unroll of If nodes later has other consequences. For example, if If node is not a root node, we may need to apply another Unrolling pass afterwards to make sure that no Vector remains in the graph.

Depending on the work to be done, it may be cleaner to handle that in a followup PR, I'm not sure right now.

mir/src/ir/quad_eval.rs
Leo-Besancon: Indeed, I've removed these (as well as `PublicInputTable`), and improved the prints explaining why we can't handle different operation types in this function depending on the reason (bus to be handled separately, operations that should have been inlinined previously, operations that should have been unrolled previously, etc.).

mir/src/ir/quad_eval.rs
Leo-Besancon: I've added a comment for this (let me know if it's clearer!).

Exactly, and the index of the public input access is also taken into account in the hash, so in the following:
```
public_inputs {
  stack_inputs: [2],
  stack_outputs: [1],
}
```

- `stack_inputs[0]`, `stack_inputs[1]` and `stack_outputs[0]` would each receive a different random value.
- If we have in the graph multiple accesses to e.g. `stack_inputs[0]`, they will all return the same value.

mir/src/passes/unrolling.rs
Leo-Besancon: > Question: what does bis stand for?

I've answered this one here: https://github.com/0xMiden/air-script/pull/413#discussion_r2222587318. Short answer: We already have visit_if() in visitor.rs with a different signature, this should probably be improved in #425.

> Separately, I would probably create a helper struct to improve the readability of this function. We could instantiate this struct with a reference to self.random_inputs and then have methods on it to help with various stages of the algorithm (e.g., add_match_arm(), reduce_constraints() - just off the top of my head, may be a better way to break things up).

Indeed I think it's better to be handled in #425, thank you for the ideas!

mir/src/ir/quad_eval.rs
bobbinth: Makes sense - thank you for such a thorough explanation! Let's create an issue for this and handle it in a separate PR.

Leo-Besancon: Sure, I've created the issue here: https://github.com/0xMiden/air-script/issues/431.

air-script/tests/selectors/selectors_combine_complex.air
bobbinth: > Do you think it's suitable to have `p.remove(3, 4);` be handled the same as `p.remove(3, 4) when 1;`?
> This way, if it's inside a match, we can handle it by multiplying the two selectors as we do now.

I think this could work, but we could think this through more as a part of #430.

bobbinth (approved):
All looks good! Thank you! As discussed in the comments, I think we have a few follow-ups coming out from here. The most important ones, I think are #425 and #431.

PR #415, opened 2025-06-28
refactor(bus boundaries): Add `PublicInputTable` node type to IR to handle bus boundary constraints uniformly

### LINKED ISSUE #414: Bus boundaries: Add node to graph referencing reduced public input tables
**Context**:

> This requires some modifications to the IR so we can represent the bus boundary as a separate `Value`. At the moment, these boundary constraints are handled by the codegen, which could lead to incompatible constraint polynomials between winterfell and ACE.

 _Originally posted by @adr1anh in [#399](https://github.com/0xMiden/air-script/issues/399#issuecomment-3013913809)_

**Detail**:

The generation of bus boundary constraints for public input tables is currently delegated to the codegen. This creates tension between different backends which have to ensure they produce the same constraint polynomial. By introducing a new node type for a public input table, the IR can evaluate the expression for this boundary value once.

Moreover this representation is required to #399.

adr1anh:
Fixes #414

**Introduce new `ir::PublicInputTable` node**: This node corresponds to a randomly-reduced public input table.

**Optimize Winterfell bus boundary constraint generation**: Using the new `PublicInputTable` node, we can compute the reduced public input tables once, and handle all bus boundary constraints uniformly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Introduced initial version of the ACE backend (#370, #380, #386).
 - Updated Winterfell codegen to the latest version (#388).
 - Removed obsolete MASM codegen backend (#389).
+- Add node to graph referencing reduced public input tables ([#414](https://github.com/0xMiden/air-script/issues/414))
 
 ### Internal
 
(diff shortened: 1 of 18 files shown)


Al-Kindi-0 (approved):
Looks good! I think this was reviewed as part of the follow-up PR

Leo-Besancon (approved):
LGTM!

Issue #397, opened 2025-05-09
Busses: generate powers of auxiliary random values


adr1anh:
With the upcoming changes in [VM#1810](https://github.com/0xMiden/miden-vm/pull/1810), the codegen relating to bus constraints could be updated to use powers of the `alpha` challenge, rather than independent challenges.

For prover codegen, we would need to add an extra step to compute the powers, and then use them in the same way as we currently do with the independent challenges.

For the ACE circuit, we could either
- add operations to the `build_ace_circuit` method to compute the required powers. This would simplify the MASM verifier, but require the `Air` to provide additional information about how many powers are required. It's not yet clear what that would look like.
- change the Air generation to use Horner's method when combining the bus values with the challenge powers.

This would require some investigating, and may turn out to not be worth the trouble. In that case, the MASM verifier could manually compute the challenge powers, of which there are few anyway.

bobbinth:
I wonder if how random values are generated should be defined by some options (e.g., at codegen time). This would be similar to how we can specify different types of batching in Winterfell.

Also, it may make sense to tackle this after #378 as otherwise we'd need to know how to deal with random values during IR -> AIR lowering rather than at codegen time.

Al-Kindi-0:
> With the upcoming changes in [VM#1810](https://github.com/0xMiden/miden-vm/pull/1810), the codegen relating to bus constraints could be updated to use powers of the `alpha` challenge, rather than independent challenges.
> 
> For prover codegen, we would need to add an extra step to compute the powers, and then use them in the same way as we currently do with the independent challenges.
> 

Not sure I follow, but this should be an AIR specific thing. The way it is done in the referenced PR ([here](https://github.com/0xMiden/miden-vm/blob/f96e2144011e28efc2caf84f5fdc3652e13ce489/air/src/lib.rs#L279)) is by re-implementing the `get_aux_rand_elements` method.

> For the ACE circuit, we could either
> 
>     * add operations to the `build_ace_circuit` method to compute the required powers. This would simplify the MASM verifier, but require the `Air` to provide additional information about how many powers are required. It's not yet clear what that would look like.
> 
>     * change the Air generation to use Horner's method when combining the bus values with the challenge powers.
> 
> 
> This would require some investigating, and may turn out to not be worth the trouble. In that case, the MASM verifier could manually compute the challenge powers, of which there are few anyway.

[...]

PR #416, opened 2025-06-28
feature(ACE): Handle variable length public inputs and generate powers of randomness

### LINKED ISSUE #399: Add handling of variable length public inputs to ACE codegen

adr1anh:
Fixes #399 and #397 

**Changes**:
- Update the ACE codegen to handle bus boundary constraints which refer to public input tables.
- Update `ir::Air` to return the list of public table accesses with a canonical order.
  - Use this accessor to simplify winterfell codegen
- Improve layout of ACE inputs to satisfy requirements of the MASM verifier.
- Fix random challenges to α and β and derive powers of the latter in the circuit.

### DIFF
--- a/air/src/ir/mod.rs
+++ b/air/src/ir/mod.rs
@@ -103,6 +103,28 @@ impl Air {
         self.public_inputs.values()
     }
 
+    pub fn reduced_public_input_table_accesses(&self) -> Vec<PublicInputTableAccess> {
+        let mut accesses: Vec<_> = self
+            .buses
+            .values()
+            .flat_map(|bus| {
+                [bus.first, bus.last]
+                    .iter()
+                    .filter_map(|boundary| {
+                        if let BusBoundary::PublicInputTable(access) = boundary {
+                            Some(*access)
+                        } else {
+                            None
+                        }
+                    })
+                    .collect::<Vec<_>>()
+            })
+            .collect();
+        accesses.sort();
+        accesses.dedup();
+        accesses
+    }
+
     pub fn periodic_columns(&self) -> impl Iterator<Item = &PeriodicColumn> + '_ {
         self.periodic_columns.values()
     }
--- a/air/src/ir/mod.rs
+++ b/air/src/ir/mod.rs
@@ -103,6 +103,7 @@ impl Air {
         self.public_inputs.values()
     }
 
+    /// Returns a list of all access to reduced public input tables in canonical order.
     pub fn reduced_public_input_table_accesses(&self) -> Vec<PublicInputTableAccess> {
         let mut accesses: Vec<_> = self
             .buses
--- a/air/src/ir/mod.rs
+++ b/air/src/ir/mod.rs
@@ -103,7 +103,7 @@ impl Air {
         self.public_inputs.values()
     }
 
-    /// Returns a list of all access to reduced public input tables in canonical order.
+    /// Returns a list of all accesses to reduced public input tables in canonical order.
     pub fn reduced_public_input_table_accesses(&self) -> Vec<PublicInputTableAccess> {
         let mut accesses: Vec<_> = self
             .buses
--- a/codegen/ace/src/builder.rs
+++ b/codegen/ace/src/builder.rs
@@ -138,6 +138,9 @@ impl CircuitBuilder {
                 Value::PublicInput(pi) => self.layout.public_inputs[&pi.name]
                     .as_node(pi.index)
                     .expect("invalid public input access"),
+                Value::PublicInputTable(_) => {
+                    todo!("public input tables are not supported yet (see #399)")
+                },
                 Value::RandomValue(idx) => {
                     self.layout.random_values.as_node(*idx).expect("invalid random value index")
                 },
--- a/codegen/ace/src/builder.rs
+++ b/codegen/ace/src/builder.rs
@@ -30,6 +30,7 @@ pub struct CircuitBuilder {
     ops_cache: BTreeMap<OperationNode, Node>,
     // A cache of nodes already inserted in the circuit, used to avoid duplicates.
     air_node_cache: BTreeMap<AirOperation, Node>,
+    random_values: Vec<Node>,
     // Cache mapping a periodic column identifier to the evaluation of a column at `z`.
     periodic_columns_cache: BTreeMap<QualifiedIdentifier, Node>,
 }
@@ -38,6 +39,8 @@ impl CircuitBuilder {
     /// Initializes a [`CircuitBuilder`] for a given [`Air`].
     pub fn new(air: &Air) -> Self {
         let layout = Layout::new(air);
+        let beta_node = layout.random_value_node(0).unwrap();
+        let alpha_node = layout.random_value_node(1).unwrap();
         Self {
             layout,
             constants: vec![],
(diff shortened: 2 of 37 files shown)


codegen/ace/src/builder.rs
```diff
@@ -306,6 +308,25 @@ impl CircuitBuilder {
         self.periodic_columns_cache.insert(ident, result);
         Some(result)
     }
+
+    /// Returns a [`Node`] corresponding to the random challenge at the given index.
+    /// We assume that the challenges are [ β, 1, α, α², α³, … ].
```
Al-Kindi-0: Would be great if we can flip alpha and beta i.e., `[alpha, 1, beta, beta^2, ...]`. I think this ordering was suggested by the MASM code but the reason alpha and beta are out of order there is mainly for optimization purposes

codegen/ace/src/builder.rs
```diff
@@ -306,6 +308,25 @@ impl CircuitBuilder {
...
+    fn random(&mut self, index: usize) -> Node {
+        if index == 0 {
+            return self.layout.random_beta_node();
+        }
+        // TODO: Does rand[1] equal 1, or alpha? We assume 1 here.
```
Al-Kindi-0: Indeed, it should be one

codegen/ace/src/inputs.rs
```diff
@@ -18,10 +18,14 @@ pub struct AirInputs {
     pub log_trace_len: u32,
     /// Public inputs in the same order as [`Air::public_inputs`].
     pub public: Vec<Vec<QuadFelt>>,
+    /// Reduced public input table values used as boundaries for buses.
+    pub reduced_tables: Vec<QuadFelt>,
     /// Evaluations of the *main* trace.
     pub main: [Vec<QuadFelt>; 2],
-    /// Verifier challenges used to derive the *aux* trace.
-    pub rand: Vec<QuadFelt>,
+    /// Verifier challenge α used to fingerprint bus messages for the *aux* trace.
```
Al-Kindi-0: nit: Assuming the order alpha then beta, then alpha is used for the multi-set or logup argument i.e., $X \leftarrow \alpha$. $\beta$ is then used to fingerprint/reduce the messages.

codegen/ace/src/layout.rs
```diff
@@ -56,21 +66,51 @@ impl Layout {
...
-            *current_offset += width.next_multiple_of(HASH_ALIGNMENT);
+            *current_offset += width.next_multiple_of(alignment);
             InputRegion { offset, width }
         }
 
+        fn align(offset: &mut usize, alignment: usize) {
+            *offset += offset.next_multiple_of(alignment);
+        }
+
+        // TODO: should these be element or word aligned?
```
Al-Kindi-0: (fixed length) public inputs should be double word aligned as a whole. This means that within (fixed length) PI we should have element-alignment
variable length public inputs should also be double word aligned per message. This would make Horner evaluation more optimized.

codegen/ace/src/layout.rs
```diff
@@ -56,21 +66,51 @@ impl Layout {
     /// Returns a new [`Layout`] from a description of an [`Air`]. All regions are padded according
     /// to `HASH_ALIGNMENT`, ensuring that each section starts at a word-aligned memory pointer.
```
Al-Kindi-0: should probably be updated

codegen/ace/src/layout.rs
```diff
@@ -56,21 +66,51 @@ impl Layout {
...
+        // Ensure the entire region containing the public inputs is double-word aligned
+        // since it is hashed as one contiguous array.
+        align(offset, DOUBLE_WORD_ALIGNMENT);
+
+        // List of all reduced public input table accesses in canonical order.
+        let reduced_table_accesses = air.reduced_public_input_table_accesses();
+
+        // Region containing all reduced public input table values.
+        let reduced_tables_region =
+            next_region(offset, reduced_table_accesses.len(), WORD_ALIGNMENT);
```
Al-Kindi-0: Could we store one reduced table per word? This simplifies the MASM codegen work significantly

codegen/ace/src/layout.rs
```diff
@@ -98,17 +138,29 @@ impl Layout {
...
+        // during the FRI query phase.
+        // At the moment, we do so by padding each trace with zero-valued columns.
+        let trace_segments = [0, 1].map(|_row_offset| {
+            segment_widths.map(|width| next_region(offset, width, DOUBLE_WORD_ALIGNMENT))
+        });
+
+        let stark_vars = next_region(offset, StarkVar::num_vars(), WORD_ALIGNMENT);
+
+        // Ensure the entire input region is double-word aligned
+        // TODO: Not sure if this is actually needed by the MASM verifier.
```
Al-Kindi-0: don't think so either

Al-Kindi-0 (commented):
Did an initial review and it LGTM!

codegen/ace/src/builder.rs
adr1anh: Done

air/src/ir/mod.rs
```diff
@@ -103,6 +103,29 @@ impl Air {
         self.public_inputs.values()
     }
 
+    /// Returns a list of all access to reduced public input tables in canonical order.
```
Al-Kindi-0: nit: access -> accesses ?

codegen/ace/src/layout.rs
```diff
@@ -40,37 +53,71 @@ pub struct Layout {
...
+        // Mapping of each access to its index within `reduced_tables_region`
+        // The index is doubled to match the "one variable per word" requirement.
+        let reduced_tables: BTreeMap<_, _> = reduced_table_accesses
+            .into_iter()
+            .enumerate()
+            .map(|(index, access)| (access, 2 * index))
             .collect();
 
-        let random_values = next_region(&mut inputs_offset, air.num_random_values as usize);
+        // Random challenges α, β used to fingerprint bus messages.
```
Al-Kindi-0: nit: // Random challenges α, β used to fingerprint bus messages. -> // Random challenges α, β.

Al-Kindi-0 (approved):

PR #418, opened 2025-07-09
Feat: Add a pass for common subexpression elimination in the constraint graph

### LINKED ISSUE #279: Add a pass for common subexpression elimination in the constraint graph
Currently when adding subgraphs for evaluators to the constraints graph, redundant nodes are added. We should ideally reuse existing nodes whenever possible.

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/279.

It uses random point evaluation of nodes in the Algebraic Graph to remove duplicate nodes.

Task List:
- [x] Use evaluation of the AlgebraicGraph at random points to implement CSE
- [x] Update E2E tests generated code
- [ ] De-duplicate (maybe through a trait that both MIR and AIR nodes implement?) the Eval code, for now it is duplicated in both crates
- [ ] Potentially better handle the boundary constraints for Winterfell (see below)

Note: as it uses a similar Eval struct to what was introduced in https://github.com/0xMiden/air-script/pull/413, this PR is based on this branch and will be rebased on next when merged.

I did have a question (maybe you have an idea @bobbinth ?), Winterfell expects split boundary constraints (in the form `x = y`), and as a result we always had in the graph `Sub(x, y)` for boundary constraints. After CSE, this assumption can break (e.g. we do not insert `Sub(x, 0)` in the graph). For now, I have a hacky solution to consider that `y = 0` if we do not have a Sub operation, but I'm not sure of the soundness.
What would be the prefered way to handle this?
- Maybe refuse to do CSE on boundary constraint roots?
- Or adding both `x` and `y` in the Algebraic graph, and storing both indices as the boundary constraints root?


Leo-Besancon:
Note: #419 is a duplicate of this PR with only the commits related to CSE, without the commits of #413

Leo-Besancon:
Closing in favor of #419

Issue #354, opened 2024-08-09
Optimize Wintefell codegen in some expressions


Leo-Besancon:
Following https://github.com/0xPolygonMiden/air-script/pull/352, the codegen could be improved in some instance, for example when we reduce exponentiations into multiplications.

- We can introduce temporary variables to hold results of common sub-expressions (e.g. powers of two of exponentiated elements)
- We can factorize extension field conversion in expressions that contain multiple instances of it

Additional context in the originally  comment posted by @bobbinth in https://github.com/0xPolygonMiden/air-script/pull/352#discussion_r1711059613_

PR #419, opened 2025-07-11
Feat: Add a pass for common subexpression elimination in the constraint graph

### LINKED ISSUE #279: Add a pass for common subexpression elimination in the constraint graph

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/279.

It uses random point evaluation of nodes in the Algebraic Graph to remove duplicate nodes.

Task List:
- [x] Use evaluation of the AlgebraicGraph at random points to implement CSE
- [x] Update E2E tests generated code
- [x] Update code based on reviews made for https://github.com/0xMiden/air-script/pull/413 
- [x] ~De-duplicate (maybe through a trait that both MIR and AIR nodes implement?) the Eval code, for now it is duplicated in both crates~
There might still be some improvements to be made, but the previous refactor made this less of an issue IMO, as e.g. QuadFelt type and helpers are reused, only the RandomInput struct is duplicated and there is some key differences between the two structs.
- [ ] Potentially better handle the boundary constraints for Winterfell (see below)

I did have a question (maybe you have an idea @bobbinth ?), Winterfell expects split boundary constraints (in the form `x = y`), and as a result we always had in the graph `Sub(x, y)` for boundary constraints. After CSE, this assumption can break (e.g. we do not insert `Sub(x, 0)` in the graph). For now, I have a hacky solution to consider that `y = 0` if we do not have a Sub operation, but I'm not sure of the soundness.
What would be the prefered way to handle this?
- Maybe refuse to do CSE on boundary constraint roots?
- Or adding both `x` and `y` in the Algebraic graph, and storing both indices as the boundary constraints root?


air/src/graph/mod.rs
```diff
@@ -82,6 +84,72 @@ impl AlgebraicGraph {
...
+
+                let op = self.node(node_index).op();
+                let new_op = match op {
+                    Operation::Value(_) => {
+                        // Values do not need renumbering, they are leaf nodes
+                        *op
+                    },
+                    Operation::Add(lhs, rhs) => {
+                        let new_lhs = *renumbering_map.get(lhs).unwrap();
+                        let new_rhs = *renumbering_map.get(rhs).unwrap();
```
Al-Kindi-0: nit: I would probably change the `unwrap` to an `expect` and state why it shouldn't fail

air/src/graph/mod.rs
```diff
@@ -82,6 +84,72 @@ impl AlgebraicGraph {
         self.nodes.len()
     }
 
+    pub fn evaluate_all_nodes(
```
Al-Kindi-0: Should we add a short description since this is a public function?

air/src/graph/mod.rs
```diff
@@ -82,6 +84,72 @@ impl AlgebraicGraph {
...
+        random_inputs: &mut RandomInputs,
+    ) -> Result<BTreeMap<NodeIndex, QuadFelt>, CompileError> {
+        for index in 0..self.num_nodes() {
+            let node_index = NodeIndex(index);
+            random_inputs.eval(self, &node_index)?;
+        }
+        Ok(random_inputs.evals_map.clone())
+    }
+
+    pub fn eliminate_common_subexpressions(
```
Al-Kindi-0: same comment here

air/src/ir/constraints.rs
```diff
@@ -48,6 +48,24 @@ impl Constraints {
         }
     }
 
+    pub fn renumber_constraints(&mut self, renumbering_map: &HashMap<NodeIndex, NodeIndex>) {
```
Al-Kindi-0: same comment here

air/src/ir/random_inputs.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use mir::ir::{QuadFelt, const_quad_felt, query_hashed_cur_eval, query_indexed_cur_eval};
+use rand::prelude::*;
+use winter_math::fields::f64::BaseElement as Felt;
+
+use crate::{
+    AlgebraicGraph, CompileError, NodeIndex, Operation, PeriodicColumnAccess, PublicInputAccess,
+    PublicInputTableAccess, Value,
+};
+/// Represents the current existing evaluations to persist random values taken by the same values.
```
Al-Kindi-0: nit: I would add a new line here

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -133,24 +139,32 @@ fn domain_to_str(domain: ConstraintDomain) -> String {
 /// boundary constraint expression must hold, as well as the node index that represents the root
 /// of the constraint expression that must equal zero during evaluation.
 ///
+/// Note: If, after the CSE pass, the boundary constraint is a single trace access, we return None
+/// and assume that the expression is zero.
```
Al-Kindi-0: nit: I would be more explicit here e.g., "... we return None for the constraint expression" or something like that

Al-Kindi-0 (approved):
Looks good to me, thank you!

air/src/graph/mod.rs
Leo-Besancon: Thank you! I did this and I've added a comment above the 3 cases (Add, Sub, Mul) to help contextualize it.

I've also added/improved the comments you've asked in other review comments, don't hesitate to tell me if they can still be improved!

bobbinth:
> I did have a question (maybe you have an idea @bobbinth ?), Winterfell expects split boundary constraints (in the form `x = y`), and as a result we always had in the graph `Sub(x, y)` for boundary constraints. After CSE, this assumption can break (e.g. we do not insert `Sub(x, 0)` in the graph). For now, I have a hacky solution to consider that `y = 0` if we do not have a Sub operation, but I'm not sure of the soundness. What would be the prefered way to handle this?
> 
> * Maybe refuse to do CSE on boundary constraint roots?
> * Or adding both `x` and `y` in the Algebraic graph, and storing both indices as the boundary constraints root?

I haven't thought too deeply about this yet, but it feels like this should be handled by the Winterfell backend somehow. Is it possible to detect such "single-node" boundary constraints in the backend and add `- 0` there? Also, curious what @Al-Kindi-0 thinks on this.

Leo-Besancon:
> > I did have a question (maybe you have an idea @bobbinth ?), Winterfell expects split boundary constraints (in the form `x = y`), and as a result we always had in the graph `Sub(x, y)` for boundary constraints. After CSE, this assumption can break (e.g. we do not insert `Sub(x, 0)` in the graph). For now, I have a hacky solution to consider that `y = 0` if we do not have a Sub operation, but I'm not sure of the soundness. What would be the prefered way to handle this?
> > 
> > * Maybe refuse to do CSE on boundary constraint roots?
> > * Or adding both `x` and `y` in the Algebraic graph, and storing both indices as the boundary constraints root?
> 
> I haven't thought too deeply about this yet, but it feels like this should be handled by the Winterfell backend somehow. Is it possible to detect such "single-node" boundary constraints in the backend and add `- 0` there? Also, curious what @Al-Kindi-0 thinks on this.

This is what I'm currently doing yes, and it's working in my tests.
But I feel like it could break and was wondering if there could be a better way.
 
For instance, if we've already inserted the node `Add(Sub(a, 2), 1)` in the graph, and then we try to insert our constraint `Sub(a,1)`.
As it evaluates to the same value as a known node, during CSE it would not be inserted in the graph, keeping the "Add", and our Winterfell backend will not be able to find `a`or `1` easily in the node.

As a side note:
- Currently I always keep the first node inserted in the graph with a given evaluation. We could think about checking "how complex" it is and keeping the simpler of the two nodes, in this case, `Sub(a,1)` would be simpler than `Add(Sub(a, 2), 1)`. But this may be a bit hard to implement as replacing one node already in the graph with another one does not look trivial in my current implementation.
- If we insert boundary constraints first in the graph as we do now, I don't think the issue I've describe can arise.

Al-Kindi-0:
If I understand your concern correctly Leo, then I think processing the boundary constraints first is probably the way to go. It is a sub-optimal solution in the sense that it makes things rigid, but if we document why this choice was made then this would be "just" an implementation detail of the CSE pass.

The other solution of adding a complexity measure to choose between equivalent expressions is also plausible as a solution but will require further thinking when compared to the first solution.

Let me know if I missed your point, wholly or in part.

Leo-Besancon:
> If I understand your concern correctly Leo, then I think processing the boundary constraints first is probably the way to go. It is a sub-optimal solution in the sense that it makes things rigid, but if we document why this choice was made then this would be "just" an implementation detail of the CSE pass.
> 
> The other solution of adding a complexity measure to choose between equivalent expressions is also plausible as a solution but will require further thinking when compared to the first solution.
> 
> Let me know if I missed your point, wholly or in part.

It makes sence to me, I've added a few comments regarding this. Thank you!

air-script/tests/selectors/selectors_combine_complex.rs
```diff
@@ -97,6 +97,6 @@ impl Air for SelectorsAir {
...
+        result[0] = ((aux_rand_elements.rand_elements()[0] + aux_rand_elements.rand_elements()[1] + E::from(Felt::new(2_u64)) * aux_rand_elements.rand_elements()[2]) * E::from(main_current[0]) * E::from(main_current[5]) + E::ONE - E::from(main_current[0]) * E::from(main_current[5])) * ((aux_rand_elements.rand_elements()[0] + aux_rand_elements.rand_elements()[1] + E::from(Felt::new(2_u64)) * aux_rand_elements.rand_elements()[2]) * (E::ONE - E::from(main_current[0])) * E::from(main_current[1]) * E::from(main_current[5]) + E::ONE - (E::ONE - E::from(main_current[0])) * E::from(main_current[1]) * E::from(main_current[5])) * aux_current[0] - ((aux_rand_elements.rand_elements()[0] + E::from(Felt::new(3_u64)) * aux_rand_elements.rand_elements()[1] + E::from(Felt::new(4_u64)) * aux_rand_elements.rand_elements()[2]) * (E::ONE - E::from(main_current[0])) * (E::ONE - E::from(main_current[1])) * E::from(main_current[4]) + E::ONE - (E::ONE - E::from(main_current[0])) * (E::ONE - E::from(main_current[1])) * E::from(main_current[4])) * aux_next[0];
     }
```
adr1anh, line 101: Above, we have `1-current[0]` twice. Is this something that's easy to combine?

adr1anh: We should also be able to compute this (and other similar instances) over the base field, and then cast to the extension field. This should be slightly faster.

Leo-Besancon: > Above, we have `1-current[0]` twice. Is this something that's easy to combine?

In the algebraic graph, they refer to the same node (I think before CSE it was already the case).
During the winterfell codegen, we insert the text representation of the node where it appears.

I think factorizing the expression would be a bit hard arithmetically, but doable.
Otherwise, we can also use tempory variables for nodes that are are in multiple part of an expression.

This last idea and your second point are part of the following Winterfell optimization issue: https://github.com/0xMiden/air-script/issues/354, maybe it needs to be updated a bit.

air/src/passes/translate_from_mir.rs
```diff
@@ -75,6 +75,8 @@ impl Pass for MirToAir<'_> {
         let graph = mir.constraint_graph();
 
         // We insert all the constraints into the AIR graph.
+        // Note: We need to insert the boundary constraints before the integrity constraints
+        // as it's a requirement for the CommonSubexpressionElimination pass.
```
bobbinth: As discussed offline, this may not be needed. The reason is that a boundary constraint expressions (after constant folding and common sub-expression elimination passes) should always get reduced to one of the following:

- `sub(column, constant)` - this case is already handled.
- `column` - this means that the constant is `0` and we should output the assertion accordingly.

All other cases should probably be considered errors - but it is worth double checking (and maybe creating some tests) to make sure we can't get a boundary constraint in any other form (at least for now).

Leo-Besancon: I have found a minimal example that fails if we insert integrity constraints before boundary constraints:

```
trace_columns {
    main: [a, b],
}

boundary_constraints {
   enf b.first = 1;
}

integrity_constraints {
    enf (b - a - 1 + a) = 0;
}
```

Here, the current CSE pass will not reduce `b-a-1+a` to `b-1`, because in the following nodes:
```
0:  b
1:  a
2:  Sub(b, a)
3:  1
4:  Sub(Sub(b, a), 1)
5:  Add(Sub(Sub(b, a), 1), a)
```

Each node does not evaluate to a node previously inserted in the graph.

This means that the boundary constraints will then be replaced with the 5th node here, and we have an `Add` instead of a `Sub` or a column.

I think this would only work if we have a more involved arithmetic optimization.

adr1anh: Just a quick thought, but could we not represent the above by linear combination? That is, a vector of pairs `(coefficient, column)`, which can then be ordered and optimized as we add or subtract elements. 

We could later apply an optimization for small LCs where we check if any tuple of elements can be replaced by an existing node.

Leo-Besancon: > Just a quick thought, but could we not represent the above by linear combination? That is, a vector of pairs `(coefficient, column)`, which can then be ordered and optimized as we add or subtract elements.
> 
> We could later apply an optimization for small LCs where we check if any tuple of elements can be replaced by an existing node.

I have two questions on this:
- Do you mean to describe all nodes as a linear combination? or only constraint roots?
- Wouldn't we also need to have a power element to handle constraints of higher degrees?

bobbinth: > I have found a minimal example that fails if we insert integrity constraints before boundary constraints

Ah nice! Now indeed it it obvious that there is an issue.

This is probably not for this PR (and so we can keep the ordering assumption for now as is) - but eventually, we should probably do more sophisticated algebraic analysis on the nodes. It could be something like what @adr1anh suggested or something else.

We could even try to do an exhaustive search here to come up with optimal expressions. We may be able to do this because:

- The expressions we deal with are not too big. I think these would be a couple hundred nodes deep in the worst case (and frequently just a few nodes deep).
- The compilation time is not a huge concern: we can afford to wait a minute or two (or even more) to come up with optimal constraints.

adr1anh: > Do you mean to describe all nodes as a linear combination? or only constraint roots?

My thinking is that constant, addition, and subtraction node types could all be represented by a pair `(Const, Vec<(Coeff, NodeIndex)>)`, leading to a linear expression. With this approach in the above example, we would end up with `-1 + [1 * b]`. You can multiply an LC with other LCs or other more complex nodes. 

We would map the existing IR to a simplified one, collapsing the 3 linear nodes (constant, addition, and subtraction) into a single one.

> Wouldn't we also need to have a power element to handle constraints of higher degrees?

Not really, that would be handled by the multiplication nodes in the Air IR.

Leo-Besancon: > > Do you mean to describe all nodes as a linear combination? or only constraint roots?
> 
> My thinking is that constant, addition, and subtraction node types could all be represented by a pair `(Const, Vec<(Coeff, NodeIndex)>)`, leading to a linear expression. With this approach in the above example, we would end up with `-1 + [1 * b]`. You can multiply an LC with other LCs or other more complex nodes.
> 
> We would map the existing IR to a simplified one, collapsing the 3 linear nodes (constant, addition, and subtraction) into a single one.
> 
> > Wouldn't we also need to have a power element to handle constraints of higher degrees?
> 
> Not really, that would be handled by the multiplication nodes in the Air IR.

I see, thank you!

air/src/passes/translate_from_mir.rs
```diff
@@ -89,6 +89,8 @@ impl Pass for MirToAir<'_> {
...
```
bobbinth: I would probably mention that this is needed specifically for the winterfell codegen.

air/src/ir/constraints.rs
```diff
@@ -48,6 +48,26 @@ impl Constraints {
...
     }
 
+    /// Updates the root boundary and integrity constraints to use the new node indices
+    /// values, given in the `renumbering_map`.
+    pub fn renumber_constraints(&mut self, renumbering_map: &HashMap<NodeIndex, NodeIndex>) {
+        // Renumber the boundary constraints
+        for (_, segment_constraints) in self.boundary_constraints.iter_mut() {
+            for constraint in segment_constraints.iter_mut() {
+                constraint
+                    .update_node_index(*renumbering_map.get(constraint.node_index()).unwrap());
```
bobbinth: nit: here and on line 66 below, I'd use `expect()` instead of `unwrap()`.

air/src/graph/mod.rs
```diff
@@ -82,6 +84,85 @@ impl AlgebraicGraph {
...
+    /// and returns a map of node indices to their evaluations.
+    pub fn evaluate_all_nodes(
+        &mut self,
+        random_inputs: &mut RandomInputs,
+    ) -> BTreeMap<NodeIndex, QuadFelt> {
+        for index in 0..self.num_nodes() {
+            let node_index = NodeIndex(index);
+            random_inputs.eval(self, &node_index);
+        }
+        random_inputs.evals_map.clone()
```
bobbinth: nit: I'd add a method on `RandomInputs` so that here we could do something like:
```rust
random_inputs.into_evaluations()
```

air/src/graph/mod.rs
```diff
@@ -82,6 +84,85 @@ impl AlgebraicGraph {
...
+    }
+
+    /// Given the map of evaluations of all nodes in the graph,
+    /// eliminates common subexpressions by replacing nodes with identical evaluations.
```
bobbinth: nit: comment wrapping.

air/src/graph/mod.rs
```diff
@@ -82,6 +84,85 @@ impl AlgebraicGraph {
         self.nodes.len()
     }
 
+    /// Evaluates all the nodes in the graph at random points
+    /// and returns a map of node indices to their evaluations.
+    pub fn evaluate_all_nodes(
+        &mut self,
+        random_inputs: &mut RandomInputs,
+    ) -> BTreeMap<NodeIndex, QuadFelt> {
```
bobbinth: nit: comment wrapping.

Also: why do we need to return a `BTreeMap` from here? If node indexes are sequential, we should be able to return just `Vec<QuadFelt>` where we should be able to just index into the vector using node index, no?

Also also: do we need to pass in `random_inputs` into this? From what I can tell, we are not re-using them here across multiple invocations, and if so, we could create `random_inputs` inside of this method. But maybe I missed something?

To summarize, maybe this could be:
```rust
pub fn evaluate_on_random_inputs(&self) -> Vec<QuadFelt> {
    ...
}
```

air/src/graph/mod.rs
```diff
@@ -82,6 +84,85 @@ impl AlgebraicGraph {
...
+                let new_op = match op {
+                    Operation::Value(_) => {
+                        // Values do not need renumbering, they are leaf nodes
+                        *op
+                    },
+                    // Note: for Add, Sub, and Mul operations, we assume it's children have already
+                    // been handled. This holds because when building the graph,
+                    // we always insert children before parents, so their
+                    // indices will always be lower and thus have been already processed and added
+                    // to the `renumbering_map`.
```
bobbinth: nit: comment wrapping seems off.

air/src/ir/random_inputs.rs
```diff
@@ -0,0 +1,118 @@
...
+    rand_values: Vec<QuadFelt>,
+    public_inputs: HashMap<PublicInputAccess, QuadFelt>,
+    periodic_columns: HashMap<PeriodicColumnAccess, QuadFelt>,
+    public_inputs_tables: HashMap<PublicInputTableAccess, QuadFelt>,
+    // A map to hold the the current evaluations of nodes at random points
+    pub evals_map: BTreeMap<NodeIndex, QuadFelt>,
+}
+
+impl RandomInputs {
+    /// Evaluates a given MIR node at random points.
```
bobbinth: nit: "MIR` should probably be "AIR"? Or maybe it should be "algebraic graph node"?

air/src/ir/random_inputs.rs
```diff
@@ -0,0 +1,118 @@
...
+    ///
+    /// Note that we currently assume this will be called only during the unrolling phase, some
+    /// operation types are not handled.
+    pub fn eval(&mut self, graph: &AlgebraicGraph, node_index: &NodeIndex) -> QuadFelt {
```
bobbinth: Is this comment still relevant? (i.e., do we have an unrolling phase for this?)

air/src/ir/eval.rs
bobbinth, line 1: Is this file relevant here?

mir/src/ir/quad_eval.rs
```diff
@@ -22,10 +22,33 @@ fn rand_quad_felt<R: Rng + ?Sized>(rng: &mut R) -> QuadFelt {
...
     QuadFelt::new(felt, Felt::ZERO)
 }
 
+/// Helper function to either query an existing evaluation or create a new random one if the index
+/// is out of bounds.
+pub fn query_indexed_cur_eval<R: Rng + ?Sized>(
+    rng: &mut R,
+    cur_eval_vec: &mut Vec<QuadFelt>,
+    index: usize,
+) -> QuadFelt {
```
bobbinth: nit: what does "cur" here stand for? i.e., why not name this just `query_indexed_eval()`?

I would also probably rename `cur_eval_vec` to just `evaluations`.

air/src/ir/random_inputs.rs
```diff
@@ -0,0 +1,118 @@
...
+                },
+                Value::PublicInputTable(public_input_table_access) => {
+                    let eval = query_hashed_cur_eval(
+                        &mut self.rng,
+                        &mut self.public_inputs_tables,
+                        public_input_table_access,
+                    );
+                    self.evals_map.insert(*node_index, eval);
+                    eval
+                },
```
bobbinth: To double-check: all public input, periodic column, and public input table accesses would get their own random values, right? And for this, we are relying on the `Hash` trait implementation for these nodes - correct?

bobbinth (commented):
Looks good! Thank you! Not a full review yet (I haven't looked through the test results), but I left some comments/questions inline.

air/src/ir/random_inputs.rs
Leo-Besancon: That is correct, if for instance two nodes in the graph are the same public inputs, then they will be `==`, and their hash will be the same as required by HashMap's assumptions. Then, a single random value will be used for both nodes.

If the two public inputs are not `==`, their hash should be different (and the implementation is derived so this holds), and two different random values will be generated.

The same is true for periodic columns and public input tables.

Leo-Besancon:
> Looks good! Thank you! Not a full review yet (I haven't looked through the test results), but I left some comments/questions inline.

Thank you @bobbinth for your suggestions, I've updated the code based on them!

air/src/ir/constraints.rs
```diff
@@ -48,6 +48,34 @@ impl Constraints {
         }
     }
 
+    /// Updates the root boundary and integrity constraints to use the new node indices
+    /// values, given in the `renumbering_map`.
+    ///
+    /// Panics if a constraint's node index is not found in the renumbering map.
+    pub fn renumber_constraints(&mut self, renumbering_map: &HashMap<NodeIndex, NodeIndex>) {
+        // Renumber the boundary constraints
+        for (_, segment_constraints) in self.boundary_constraints.iter_mut() {
```
adr1anh: Nit: we can merge these by zipping the constraint iterators

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -101,7 +104,10 @@ fn add_aux_trace_assertions(func_body: &mut codegen::Function, ir: &Air) {
             split_boundary_constraint(ir.constraint_graph(), constraint.node_index());
         debug_assert_eq!(trace_access.segment, TraceSegmentId::Aux);
 
-        let expr_root_string = expr_root.to_string(ir, ElemType::Ext, TraceSegmentId::Aux);
+        let expr_root_string = match expr_root {
```
adr1anh, line 111: Not entirely sure, but I think we should panic here rather than providing a default value. Moreover, zero is invalid for a multi-set boundary.

adr1anh (commented):
Looks good! Left some small comments.

One final thought is whether we can re-use the plumbing to remove any duplicate constraints which evaluate to the same value?

codegen/winterfell/src/air/boundary_constraints.rs
Leo-Besancon: This part handles empty buses constraints that are part of the algebraic graph (inserted in `expand_buses.rs`), for instance:
```
buses {
    multiset p,
    logup q,
}
boundary_constraints {
    enf p.first = null;
    enf q.first = null;
    enf p.last = null;
    enf q.last = null;
}
```

For multiset the constraints are indeed `Sub(p, 1) = 0` so we won't get zero. But for logup they are `Sub(q, 0) = 0`, and in that case, CSE will reduce the `Sub` node to just `q` and the `expr_root` returned by `split_boundary_constraint` will be None.

air/src/ir/constraints.rs
Leo-Besancon: Thank you! I've refactored that part and included the de-duplication of constraints in the same function (it seemed easier to handle everything at once even though it reduces readability a bit).

air/src/ir/constraints.rs
```diff
@@ -131,6 +163,10 @@ impl ConstraintRoot {
         &self.index
     }
 
+    pub fn update_node_index(&mut self, new_index: NodeIndex) {
```
bobbinth, line 171: Let's add a doc comment here.

air/src/ir/constraints.rs
```diff
@@ -48,6 +49,37 @@ impl Constraints {
         }
     }
 
+    /// Updates the root boundary and integrity constraints to use the new node indices
+    /// values, given in the `renumbering_map`.
+    /// This functions also removes duplicate constraints (that share the same root and domain).
+    ///
+    /// Panics if a constraint's node index is not found in the renumbering map.
+    pub fn renumber_and_deduplicate_constraints(
```
bobbinth: nit: let's add a newline after line 53. Also, we usually put panics into their own section. Basically, the doc comment here could look like so:
```rust
/// Updates the root boundary and integrity constraints to use the new node indices
/// values, given in the `renumbering_map`.
///
/// This functions also removes duplicate constraints (that share the same root and domain).
///
/// # Panics
/// Panics if a constraint's node index is not found in the renumbering map.
```

air/src/ir/constraints.rs
```diff
@@ -1,4 +1,5 @@
 use core::fmt;
+use std::collections::HashSet;
```
bobbinth: nit: we usually prefer to use `BTreeMap` and `BTreeSet` unless really critical for performance reasons to use `HashMap` and `HashSet`. The main advantages are `BTreeMap` and `BTreeSet` are:

- Predictable iteration order.
- Can be used in `no_std` context.

air/src/ir/random_inputs.rs
```diff
@@ -0,0 +1,118 @@
...
+    // A vector to hold the random values taken for the main trace, indexed in the following way:
+    // $main[0], $main[0]', $main[1], $main[1]', $main[2], ...
+    main_trace: Vec<QuadFelt>,
+    aux_trace: Vec<QuadFelt>,
+    rand_values: Vec<QuadFelt>,
+    public_inputs: HashMap<PublicInputAccess, QuadFelt>,
+    periodic_columns: HashMap<PeriodicColumnAccess, QuadFelt>,
+    public_inputs_tables: HashMap<PublicInputTableAccess, QuadFelt>,
+    // A map to hold the the current evaluations of nodes at random points
+    pub evals_map: BTreeMap<NodeIndex, QuadFelt>,
```
bobbinth: Does this field need to be public?

air/src/ir/random_inputs.rs
```diff
@@ -0,0 +1,118 @@
...
+                    );
+                    self.evals_map.insert(*node_index, eval);
+                    eval
+                },
+            },
+        }
+    }
+
+    /// Returns all the evaluations, ordered by `NodeIndex`.
+    pub fn into_evaluations(&self) -> Vec<QuadFelt> {
```
bobbinth: nit: by convention `into_` methods usually consume `self`. And I think in our context that should work out too.

air/src/ir/random_inputs.rs
```diff
@@ -0,0 +1,118 @@
...
+use winter_math::fields::f64::BaseElement as Felt;
+
+use crate::{
+    AlgebraicGraph, NodeIndex, Operation, PeriodicColumnAccess, PublicInputAccess,
+    PublicInputTableAccess, Value,
+};
+
+/// Represents the current existing evaluations to persist random values taken by the same values.
+#[derive(Debug, Clone, Default)]
+pub struct RandomInputs {
```
bobbinth: The doc comment here is a bit confusing. Could we expand this a bit to explain that this struct both random inputs needed by leaf nodes of the algebraic graph as well as evaluations of every node?

air/src/ir/random_inputs.rs
```diff
@@ -0,0 +1,118 @@
+use std::collections::{BTreeMap, HashMap};
```
bobbinth: nit: similar to the other nit - I'd probably stick to using `BTreeMap`s and import them from `core`.

air/src/graph/mod.rs
```diff
@@ -82,6 +84,84 @@ impl AlgebraicGraph {
...
+    /// Given the evaluations of all nodes in the graph, ordered by their node indices, eliminates
+    /// common subexpressions by replacing nodes with identical evaluations.
+    ///
+    /// In the process, as some nodes will be removed from the graph, node indices need to be
+    /// remapped to keep consistent values (from `NodeIndex(0)` to `NodeIndex(self.num_nodes())`).
+    /// This function returns the node indices remapping map.
+    pub fn eliminate_common_subexpressions(
+        &mut self,
+        evals: &[QuadFelt],
+    ) -> HashMap<NodeIndex, NodeIndex> {
```
bobbinth: Thinking about this more, it is not entirely clear to me that there is much value in splitting these into two public functions. An alternative would be to make `evaluate_on_random_inputs()` private and then just call it as the first thing in the `eliminate_common_subexpressions()`.

Also, I'd probably move both of these functions into a sub-module - e.g., in `src/graph/cse.rs` file. Though, I'd still keep them attached to the `AlgebraicGraph` struct. So, in `cse.rs` it would look something like this:
```rust
impl AlgebraicGraph {
    pub fn eliminate_common_subexpressions(
        &mut self
    ) -> HashMap<NodeIndex, NodeIndex> {
        let evals = self.evaluate_on_random_inputs();
        ...
    }

    fn evaluate_on_random_inputs(&self) -> Vec<QuadFelt> {
        ...
    }
}
```

bobbinth (commented):
Looks good! Thank you! I've reviewed all the code (though still haven't checked the test results) and left some more comments inline.

codegen/winterfell/src/air/boundary_constraints.rs
adr1anh: Thanks for the explanation. Can we make the comment a bit clearer though?

air/src/ir/constraints.rs
Leo-Besancon: I've replaced HashMap/HashSet by BTreeMap/BTreeSet (for the structs related to this PR only). I've imported them from alloc crate with:
```rust
extern crate alloc;
use alloc::collections::{BTreeMap, BTreeSet};
```

Leo-Besancon:
> Looks good! Thank you! I've reviewed all the code (though still haven't checked the test results) and left some more comments inline.

Thank you, the latest commits should resolve your comments.

codegen/winterfell/src/air/boundary_constraints.rs
Leo-Besancon: Sure, I've explained what's happening in 1554db65404a38d85e516cff4e8e7c61add28503, let me know if it's still unclear!

bobbinth (approved):
Looks good! Thank you! We just need to merge the latest `next`, and I think we are good to go.

Leo-Besancon:
> Looks good! Thank you! We just need to merge the latest `next`, and I think we are good to go.

Great! I've rebased on next.

adr1anh (approved):
Looks good to me!

PR #420, opened 2025-07-14
Fixes some bugs and clean up

Al-Kindi-0:
As the title says.
Also adds a check to the bus pass in order to better handle unconstrained buses. Before this, if a bus didn't have any constraints then constraints of the form `b_next = b_current` would be added. Clearly this is not the desired behavior.


codegen/ace/src/layout.rs
```diff
@@ -89,19 +92,25 @@ impl Layout {
...
+                } else {
+                    None
+                }
+            })
             .collect();
 
         // Ensure the entire region containing the public inputs is double-word aligned
         // since it is hashed as one contiguous array.
-        align(offset, Alignment::DoubleWord);
+        align(offset, Alignment::DoubleDoubleWord);
```
adr1anh: The alignment is in terms of extension field elements, so this would effectively pad the entire region to a multiple of 16 felts. 

For example, if we have 3 actual public input felts encoded as 6 extension field elements, then this would add 5 dummy inputs and we would need 16 felts in total.

codegen/ace/src/layout.rs
```diff
@@ -110,7 +119,7 @@ impl Layout {
         // For MASM efficiency, we store one reduced table per word.
         // Each variable therefore occupies two "variable slots".
         let reduced_tables_region =
-            next_region(offset, 2 * reduced_table_accesses.len(), Alignment::DoubleWord);
+            allocate_region(offset, 2 * reduced_table_accesses.len(), Alignment::Word);
```
adr1anh, line 125: This is because each reduced table is computed rather than unhashed right?

codegen/ace/src/layout.rs
```diff
@@ -89,19 +92,25 @@ impl Layout {
         }
 
         fn align(offset: &mut usize, alignment: Alignment) {
-            *offset += offset.next_multiple_of(alignment as usize);
+            *offset = offset.next_multiple_of(alignment as usize);
```
adr1anh, line 98: Good catch!

codegen/ace/src/layout.rs
```diff
@@ -259,21 +268,21 @@ impl InputRegion {
 /// List of STARK variables and challenges, derived from the public parameters and proof transcript.
 #[derive(Copy, Clone, Debug)]
 pub enum StarkVar {
```
adr1anh, line 273: We need to update `TryFrom` below
```rust
    fn try_from(value: usize) -> Result<Self, Self::Error> {
        match value {
            0 => Ok(Self::Alpha),
            1 => Ok(Self::Z),
            2 => Ok(Self::ZPowN),
            3 => Ok(Self::GenLast),
            4 => Ok(Self::ZMaxCycle),
            5 => Ok(Self::GenPenultimate),
            _ => Err(value),
        }
    }
```

adr1anh: Also update `[α, z, zⁿ, g⁻¹, zᵐᵃˣ, g⁻²`]` in [this comment](https://github.com/0xMiden/air-script/blob/06279871779adcaab682bf1beca889559c9c6fad/codegen/ace/src/lib.rs#L60) (there was a small typo there before)

adr1anh (commented):
Looks good, and thanks for catching these bugs!

Left some small nits about correcting comments and questions about alignment

codegen/ace/src/layout.rs
Al-Kindi-0: Yes, this is the desired behavior. The reason for this is that we want to load (fixed) public inputs in groups of 8 and hence store them in groups of 8 extension field elements.
Will update the comment as I see it is out of date

codegen/ace/src/layout.rs
Al-Kindi-0: Yes, exactly. We store each reduced value in a separate word and hence we only need word-alignment

codegen/ace/src/layout.rs
Al-Kindi-0: Fixed, thanks!

codegen/ace/src/layout.rs
adr1anh: And we still assume that if there are multiple fixed public input vectors, then these will be concatenated?

Al-Kindi-0: If I understand correctly, we do assume that all fixed public inputs are concatenated and, as a group, they are double-double-word aligned

adr1anh (approved):
Nice work!

codegen/ace/src/layout.rs
```diff
@@ -89,19 +95,25 @@ impl Layout {
...
+                }
+            })
             .collect();
 
-        // Ensure the entire region containing the public inputs is double-word aligned
-        // since it is hashed as one contiguous array.
-        align(offset, Alignment::DoubleWord);
+        // Ensure the entire region containing the public inputs is double-double-word aligned
+        // since it is hashed as one contiguous array and processed in batches of 8 inputs.
+        align(offset, Alignment::DoubleDoubleWord);
```
bobbinth: nit: would something like `QuadWord` be better than `DoubleDoubleWord`?

bobbinth (approved):
Looks good! Thank you! This is a very shallow review from me, but I left one nit inline.

Also, more general comment: it looks like we fixed some bugs in this PR which were not covered by tests. Should we add tests to make sure we cover these cases for the future? (could be done in a separate PR).

codegen/ace/src/layout.rs
Al-Kindi-0: Good point, changed

Al-Kindi-0:
> Looks good! Thank you! This is a very shallow review from me, but I left one nit inline.
> 
> Also, more general comment: it looks like we fixed some bugs in this PR which were not covered by tests. Should we add tests to make sure we cover these cases for the future? (could be done in a separate PR).

Created #421

PR #424, opened 2025-07-21
Refactor: Remove the deprecated --without-mir pipeline

### LINKED ISSUE #405: Remove the deprecated --without-mir pipeline

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/405.


air-script/tests/functions/functions_complex.rs
```diff
@@ -83,7 +83,7 @@ impl Air for FunctionsAir {
         let main_current = frame.current();
         let main_next = frame.next();
         result[0] = main_next[16] - main_current[16] * ((main_current[3] * main_current[3] * main_current[3] * main_current[3] * main_current[3] * main_current[3] * main_current[3] * main_current[1] * main_current[2] + main_current[3] * main_current[3] * (E::ONE - main_current[1]) * main_current[2] + main_current[3] * main_current[1] * (E::ONE - main_current[2]) + (E::ONE - main_current[1]) * (E::ONE - main_current[2])) * main_current[0] - main_current[0] + E::ONE);
-        result[1] = main_next[3] - (main_current[4] + main_current[5] + main_current[6] + main_current[7] + main_current[8] + main_current[9] + main_current[10] + main_current[11] + main_current[12] + main_current[13] + main_current[14] + main_current[15] + E::ONE) * E::from(Felt::new(2_u64));
+        result[1] = main_next[3] - (E::ZERO + main_current[4] + main_current[5] + main_current[6] + main_current[7] + main_current[8] + main_current[9] + main_current[10] + main_current[11] + main_current[12] + main_current[13] + main_current[14] + main_current[15] + E::ONE) * E::from(Felt::new(2_u64));
```
bobbinth, line 86: Question: why did this change?

air-script/tests/list_comprehension/list_comprehension.rs
```diff
@@ -83,7 +83,7 @@ impl Air for ListComprehensionAir {
         let main_current = frame.current();
         let main_next = frame.next();
         result[0] = main_current[0] - main_current[2];
-        result[1] = main_current[4] - main_current[0] * E::from(Felt::new(8_u64)) * main_current[11];
+        result[1] = main_current[4] - main_current[0] * E::from(Felt::new(2_u64)) * E::from(Felt::new(2_u64)) * E::from(Felt::new(2_u64)) * main_current[11];
```
bobbinth, line 86: Similar question here: seems like maybe something is not going through constant folding?

codegen/winterfell/README.md
```diff
@@ -16,11 +16,19 @@ Example usage:
...
+// Create the compilation pipeline needed to translate the AST to AIR
+let pipeline = air_parser::transforms::ConstantPropagation::new(&diagnostics)
+  .chain(mir::passes::AstToMir::new(&diagnostics))
+  .chain(mir::passes::Inlining::new(&diagnostics))
+  .chain(mir::passes::Unrolling::new(&diagnostics))
+  .chain(air_ir::passes::MirToAir::new(&diagnostics))
+  .chain(air_ir::passes::BusOpExpand::new(&diagnostics));
+
+// process the AST and the MIR to get a Result containing the AIR or a CompileError
+let air = pipeline.run(ast)
```
bobbinth, line 28: Not for this PR, but we should probably simplify the usage here somehow - maybe by allowing the user to instantiate a default pipeline which would do all of the things here. Maybe it could look something like:
```rust
let ast = parse(source.as_str()).expect("Parsing failed");
let pipeline = mir::default_pipeline();
let air = pipeline.run(ast);
```

air/README.md
```diff
@@ -15,20 +15,15 @@ Example usage:
 let ast = parse(source.as_str()).expect("Parsing failed");
 
 // Create the compilation pipeline needed to translate the AST to AIR
-let pipeline_with_mir = air_parser::transforms::ConstantPropagation::new(&diagnostics)
+let pipeline = air_parser::transforms::ConstantPropagation::new(&diagnostics)
   .chain(mir::passes::AstToMir::new(&diagnostics))
   .chain(mir::passes::Inlining::new(&diagnostics))
   .chain(mir::passes::Unrolling::new(&diagnostics))
   .chain(air_ir::passes::MirToAir::new(&diagnostics))
   .chain(air_ir::passes::BusOpExpand::new(&diagnostics));
```
bobbinth, line 23: Similar comment as above: it is great that we show the flexibility here, but would also be good to show how this can be done in a way that doesn't require the user to know about all the different pipelines and their required order.

air-script/README.md
```diff
@@ -9,14 +9,15 @@ An in-depth description of AirScript is available in the full AirScript [documen
 The compiler has three stages, which can be imported and used independently or together.
```
bobbinth: nit: this should probably say "four stages" now?

bobbinth (approved):
Looks good! Thank you! I left some comments/questions inline.

air-script/tests/functions/functions_complex.rs
Leo-Besancon: Oh, here and below the changelog does not make it clear. The two pipelines had slightly different generated graphs, so we had two different files to compare to, named `functions_complex.rs` and `functions_complex_with_mir.rs`.

I've deleted the former and renamed the later.

As to why there is a difference (e.g. for constant folding), it's because some computations are different (e.g. powers of exp operations corresponding to list comprehension indices are not yet know when doing the constant propagation. Another pass of constant prop may be needed to solve that).

codegen/winterfell/README.md
Leo-Besancon: Makes sence, added an issue here: https://github.com/0xMiden/air-script/issues/426

Al-Kindi-0 (approved):
Looks good to me!
Though I have the same remark regarding constant folding which seems to not go through in some places. Would be good to create an issue if it is not straightforward to fix in this PR.

Leo-Besancon:
> Looks good to me! Though I have the same remark regarding constant folding which seems to not go through in some places. Would be good to create an issue if it is not straightforward to fix in this PR.

Indeed this as been raised by @bobbinth as well, there is a regression in MIR as, for instance, powers of exp operations corresponding to list comprehension indices are not yet know when doing the AST constant propagation. Another pass of constant propagation after MIR would solve that, here is the corresponding issue: https://github.com/0xMiden/air-script/issues/427

bitwalker (approved):
LGTM!

PR #434, opened 2025-07-29
[MIR] Refactor / Split Unrolling pass

### LINKED ISSUE #425: Refactor / Split MIR Unrolling pass
The `unrolling.rs` file is quite large with functions that have very deep nesting - and so, it is not easy to follow. We should try to split it up into multiple simpler files (still under `unrolling` module) and maybe have more helper functions that are called form `match` statements.

_Originally posted by @bobbinth in https://github.com/0xMiden/air-script/pull/413#discussion_r2218291335_

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/425

Task list:
- [x] Split the unrolling pass into two main files (unrolling_first_pass.rs and unrolling_second_pass.rs, with the main pass handled in the associated mod.rs
- [x] Extract some match arms of the first pass into helper functions to improve readability
- [x] Create a separate structure with helper methods for `If` nodes handling
- [x] Create a separate structure with helper methods for `For` nodes handling
- [ ] ~~Improve visitor logic if possible (e.g. to avoid `visit_*_bis` functions)~~
I did not find any alternative way to handle this without added awkwardness (that I'm not sure would even work). Instead, I've added a comment in the visit_node() method to explain a bit better what we do.


mir/src/passes/unrolling/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+use unrolling_first_pass::UnrollingFirstPass;
+use unrolling_second_pass::UnrollingSecondPass;
+
+/// This pass follows a similar approach as the Inlining pass.
+/// It requires that this Inlining pass has already been done.
+///
+/// * In the first step, we visit the graph, unrolling each node type except For nodes. Instead, for
+///   these node types we gather the context to inline them in the second pass. In this first pass,
+///   we also optimize constraints found in match statements.
+/// * In the second pass, we inline the bodies of For nodes.
```
bobbinth: nit: I would put "For" in backticks - i.e.,
```
... type except `For` nodes ...
```
Rather than:
```
... type except For nodes ...
```
This comment applies in other relevant places as well.

mir/src/passes/unrolling/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// This structure is used to keep track of what is needed to inline a For node
+#[derive(Clone, Debug)]
+pub struct ForInliningContext {
+    body: Link<Op>,
+    iterators: Vec<Link<Op>>,
+    selector: Option<Link<Op>>,
+    ref_node: Link<Op>,
+}
+impl ForInliningContext {}
```
bobbinth: Why do we need this empty `impl`?

mir/src/passes/unrolling/match_optimizer.rs
```diff
@@ -0,0 +1,237 @@
...
+impl MatchOptimizer {
+    /// For all match arms, splits main and bus-related constraints, and evaluates the main
+    /// constraints
+    pub fn evaluate_match_arms(
+        &mut self,
+        node_evals: &mut Vec<QuadFelt>,
+        constraints_evaluation_indices: &mut ConstraintEvaluationMap,
+        bus_related_constraints: &mut Vec<(Link<Op>, Vec<Link<Op>>)>,
+        match_arms: Ref<Vec<MatchArm>>,
+    ) -> Result<(), CompileError> {
```
bobbinth: Why do we need to pass `node_evals` into this? From what I can tell, we don't actually do anything with this vector outside of this function. If so, `node_evals` could probably be just a local variable?

Also, based on the usage, it feels like `bus_related_constraints` should actually be an output of this function (rather than an input).

Lastly, should `constraints_evaluation_indices` be a field in `MatchOptimizer`?

To summarize, maybe it could look like so:
```rust
pub struct MatchOptimizer {
    random_inputs: RandomInputs,
    constraints_evaluation_indices: ConstraintEvaluationMap,
}

impl MatchOptimizer {
    pub fn evaluate_match_arms(
        &mut self,
        match_arms: Ref<Vec<MatchArm>>,
    ) -> Result<Vec<(Link<Op>, Vec<Link<Op>>), CompileError> {
        ...
    }
}
```

mir/src/passes/unrolling/match_optimizer.rs
```diff
@@ -0,0 +1,237 @@
...
+        new_vec
+    }
+
+    /// Adds bus-related constraints to the resulting constraints vector.
+    pub fn add_bus_constaints(
+        &self,
+        bus_related_constraints: &mut [(Link<Op>, Vec<Link<Op>>)],
+        new_vec: &mut Vec<Link<Op>>,
+        span: SourceSpan,
+    ) {
```
bobbinth: Why does this take `&self` parameter? Unless I'm missing something, I don't think we are using it in this function.

mir/src/passes/unrolling/match_optimizer.rs
```diff
@@ -0,0 +1,237 @@
...
+        eval_lens
+    }
+
+    /// Combines the constraints in an optimized way, based on their evaluations.
+    pub fn reduce_main_constraints(
+        &self,
+        constraints_evaluation_indices: &mut ConstraintEvaluationMap,
+        eval_lens: &mut BTreeMap<usize, Vec<usize>>,
+        span: SourceSpan,
+    ) -> Vec<Link<Op>> {
```
bobbinth: Same question as above: it doesn't seems like we are using `&self` in this function.

mir/src/passes/unrolling/unrolling_first_pass.rs
```diff
@@ -0,0 +1,821 @@
...
+};
+
+pub struct UnrollingFirstPass<'a> {
+    #[allow(unused)]
+    diagnostics: &'a DiagnosticsHandler,
+
+    // general context
+    work_stack: Vec<Link<Node>>,
+    // current evaluations of nodes at random points
+    match_optimizer: MatchOptimizer,
```
bobbinth: If I understood the design correctly, `MatchOptimizer` is just a wrapper around `RandomInputs` and we use it across all match expressions. An alternative could be to instantiate a `MatchOptimizer` for each match expression. Here, we'd still keep track of `RandomInputs`, and `MatchOptimizer` could then look like so:

```rust
pub struct MatchOptimizer<'a> {
    random_inputs: &'a mut RandomInputs,
    // match-specific fields
}
```
Assuming this works, I think it may be cleaner to keep the state of the optimizer in the struct rather than in local variables of the `visit_if_bis()` function.

mir/src/passes/unrolling/unrolling_first_pass.rs
```diff
@@ -0,0 +1,821 @@
...
+        // We use indices to the node_evals vector as keys, to avoid non-determinism for
+        // iterating across
+        let mut constraints_evaluation_indices: BTreeMap<usize, Vec<_>> = BTreeMap::new();
+
+        self.match_optimizer.evaluate_match_arms(
+            &mut node_evals,
+            &mut constraints_evaluation_indices,
+            &mut bus_related_constraints,
+            match_arms,
+        )?;
```
bobbinth: Related to the previous couple of comments: if we make `MatchOptimizer` expression-specific, `constraints_evaluation_indices` should probably become a filed of the `MatchOptimizer`.

Also, I'd probably move the loop outside of `evaluate_match_arms()` so that `MatchOptimizer` processes one arm at a time.

mir/src/passes/unrolling/unrolling_first_pass.rs
```diff
@@ -0,0 +1,821 @@
...
+        // evaluations first, in order to minimize the number of constraints in the
+        // final vector
+        let mut eval_lens = self.match_optimizer.compute_eval_lens(&constraints_evaluation_indices);
+
+        // 3. Construct the new vector of combined constraints
+        let mut new_vec = self.match_optimizer.reduce_main_constraints(
+            &mut constraints_evaluation_indices,
+            &mut eval_lens,
+            if_ref.span,
+        );
```
bobbinth: Given that `compute_eval_lens()` is pretty simple, I'd probably combine these two functions and just call `compute_eval_lens()` from within `reduce_main_constraints()`.

bobbinth (commented):
Looks good! Thank you. Not a full review yet (I mostly focused on the `MatchOptimizer` for now), but I left a some comments inline.

mir/src/passes/unrolling/match_optimizer.rs
Leo-Besancon: > Why do we need to pass `node_evals` into this? From what I can tell, we don't actually do anything with this vector outside of this function. If so, `node_evals` could probably be just a local variable?

I moved `node_evals` as a field of MatchOptimizer instead, because I'm moving the loop outside of evaluate_match_arms() as suggested below.

Leo-Besancon:
Thank you @bobbinth, I believe I've taken into account all your comments for now in the latest commits.

mir/src/passes/unrolling/match_optimizer.rs
Leo-Besancon: Note: we do now that `constraints_evaluation_indices` is in self's fields.

mir/src/passes/unrolling/unrolling_first_pass.rs
Leo-Besancon: It makes sence. I've kept `bus_related_constraints` and `combined_main_constraints` in `visit_if_bis()` as they are more related to the constraints than evaluations, but I could move them as fields as well, tell me what you think is best!

mir/src/passes/inlining.rs
```diff
@@ -12,15 +12,15 @@ use crate::{
...
 };
 
-/// This pass handles inlining of Call nodes at there call sites.
+/// This pass handles inlining of `Call` nodes at their call sites.
 ///
 /// It works in three steps:
 /// * Firstly, we visit the graph to build the call dependency graph.
 /// * This dependency graph is then used to compute the wanted inlining order (we first replace
-///   calls to callees that do not have Calls in their body). If it is not possible create this
+///   calls to callees that do not have `Call` in their body). If it is not possible create this
```
Al-Kindi-0: ...possible to create this order ...

mir/src/passes/unrolling/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+
+mod match_optimizer;
+mod unrolling_first_pass;
+mod unrolling_second_pass;
+
+use unrolling_first_pass::UnrollingFirstPass;
+use unrolling_second_pass::UnrollingSecondPass;
+
+/// This pass follows a similar approach as the Inlining pass.
+/// It requires that this Inlining pass has already been done.
```
Al-Kindi-0: Would this be better ?

/// This pass follows a similar approach as the Inlining pass and requires that the latter has already been done.

mir/src/passes/unrolling/unrolling_first_pass.rs
```diff
@@ -0,0 +1,804 @@
...
+                };
+                acc_node = new_acc_node;
+            }
+            updated_fold = Some(acc_node);
+        }
+
+        Ok(updated_fold)
+    }
+
+    /// Visiting an `If` node consists into evaluating all the main trace constraints contained in
```
Al-Kindi-0: ...consists into -> consists of

mir/src/passes/unrolling/unrolling_first_pass.rs
```diff
@@ -0,0 +1,804 @@
...
+            .clone()
+            .into_iter()
+            .map(|bc| bc.as_node())
+            .chain(integrity_constraints_roots_ref.clone().into_iter().map(|ic| ic.as_node()))
+            .chain(bus_roots.into_iter().map(|b| b.as_node()));
+        combined_roots.collect()
+    }
+
+    fn visit_node(&mut self, graph: &mut Graph, node: Link<Node>) -> Result<(), CompileError> {
+        // We keep a reference to all For nodes to avoid dropping the backlinks stored in Parameters
```
Al-Kindi-0: For -> `For` ?

mir/src/passes/unrolling/unrolling_second_pass.rs
```diff
@@ -0,0 +1,173 @@
...
+            all_for_nodes,
+        }
+    }
+}
+impl Visitor for UnrollingSecondPass<'_> {
+    fn work_stack(&mut self) -> &mut Vec<Link<Node>> {
+        &mut self.work_stack
+    }
+    // The root nodes visited during the second pass are the children of `For` nodes to inline
+    fn root_nodes_to_visit(&self, _graph: &Graph) -> Vec<Link<Node>> {
```
Al-Kindi-0: Q: Is there a reason for not adding newlines?

Al-Kindi-0 (approved):
Looks good to me.

Leo-Besancon:
> Looks good to me.

Thank you! I've added newlines and improved comments based on your review.

mir/src/passes/unrolling/match_optimizer.rs
```diff
@@ -0,0 +1,241 @@
...
+        // 1.2. Filter out all BusOp nodes from the constraints, we will handle them separately
+        for constraint in all_constraints {
+            match constraint.borrow().deref() {
+                Op::BusOp(_) => bus_related_constraints_for_match_arm.push(constraint.clone()),
+                Op::Enf(enf) => match enf.expr.borrow().deref() {
+                    Op::BusOp(_) => bus_related_constraints_for_match_arm.push(enf.expr.clone()),
+                    _ => constraints_to_eval.push(constraint.clone()),
+                },
+                _ => constraints_to_eval.push(constraint.clone()),
+            }
```
bobbinth: I would maybe add a comment explaining that there are only two ways from bus operations to appear in the source code - `enf bus_op` (in boundary constraints?) and just `bus_op` (in integrity constraints, I think).

Also question: why in case of `Op::BusOp` we push the original constraint into `bus_related_constraints_for_match_arm`, but in case of `Op::Enf( Op::BusOp)` we push the inner expression of the constraint?

mir/src/passes/unrolling/match_optimizer.rs
```diff
@@ -0,0 +1,241 @@
...
+    // A reference to the random inputs used to evaluate the constraints
+    // (common for all matches)
+    random_inputs: &'a mut RandomInputs,
+    // Used to keep track of the evaluations values we've seen so far
+    // in the current `If` node.
+    node_evals: Vec<QuadFelt>,
+    // Used to keep track of the evaluation of each constraint
+    // We use indices to the node_evals vector as keys, to avoid non-determinism for
+    // iterating across random evaluations.
+    constraints_evaluation_indices: ConstraintEvaluationMap,
```
bobbinth, line 65: These two fields are related, right? Is the main reason to keep them separate that we can't use `QuadFelt` as a key in a map? If so, it may be more clear to create a simple wrapper type here that implements `PartialOrd` on `QuadFelt`.

mir/src/passes/unrolling/match_optimizer.rs
```diff
@@ -0,0 +1,241 @@
...
+use crate::{
+    CompileError,
+    ir::{
+        Add, ConstantValue, Link, MatchArm, MirValue, Mul, Op, Parent, QuadFelt, RandomInputs,
+        SpannedMirValue, Sub, Value,
+    },
+    passes::duplicate_node,
+};
+
+type ConstraintEvaluationMap = BTreeMap<usize, Vec<(Link<Op>, Link<Op>)>>;
```
bobbinth, line 50: Could we add a doc comment here explaining what the tuples in the vector represent? My understanding is that it is `(condition, constraint)` - but may be good to give an example.

mir/src/passes/unrolling/match_optimizer.rs
bobbinth, line 1: It would be good to add general explanation of the methodology (maybe as module-level comment) - kind of similar to what we had in the issue description (though, maybe somewhat simplified).

mir/src/passes/unrolling/match_optimizer.rs
```diff
@@ -0,0 +1,241 @@
...
+            constraints_evaluation_indices: ConstraintEvaluationMap::new(),
+        }
+    }
+
+    /// For a given match arm, splits main and bus-related constraints, and evaluates the main
+    /// constraints
+    pub fn evaluate_match_arm(
+        &mut self,
+        match_arm: &MatchArm,
+    ) -> Result<Vec<Link<Op>>, CompileError> {
```
bobbinth: I would also mention that:

- We return a list of bus-related constraints (and it seems like we may be transforming them as well?).
- That we group all main constraints by the evaluations - i.e., two constraints that evaluate to the same value, will be put in the same group.

mir/src/passes/unrolling/match_optimizer.rs
```diff
@@ -0,0 +1,241 @@
...
+                },
+            }
+        }
+        Ok(bus_related_constraints_for_match_arm)
+    }
+
+    /// For each evaluation, computes the number of constraints (with different selectors)
+    /// evaluating to it.
+    // Returns eval_lens: BTreeMap<len, Vec<evaluation_index>>
+    pub fn compute_eval_lens(&self) -> BTreeMap<usize, Vec<usize>> {
```
bobbinth: A couple of nits: 
- this could be just a helper function, right? (i.e., not `pub`)
- Should it be `///` online 117 rather than `//`?

mir/src/passes/unrolling/match_optimizer.rs
```diff
@@ -0,0 +1,241 @@
...
+        }
+        new_vec
+    }
+
+    /// Adds bus-related constraints to the resulting constraints vector.
+    pub fn gather_all_constraints(
+        bus_related_constraints: &mut [(Link<Op>, Vec<Link<Op>>)],
+        combined_main_constraints: Vec<Link<Op>>,
+        span: SourceSpan,
+    ) -> Vec<Link<Op>> {
```
bobbinth: I would also maybe mention that we are computing latches for the constraints here.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - once these are addressed, I think we can merge.

mir/src/passes/unrolling/match_optimizer.rs
Leo-Besancon: Not exactly, in a previous version of the code I used the ordering of QuadFelt by converting it to base_elements and then to ints, but the problem is that the order depends on the (random) evaluations, so is not deterministic across different runs of the program unless we rely on a seed.

Now, I use the `node_evals` field to have an ordering based on the graph of the constraints. Given the `If` node with two match arms `(s0, [A, C])` and `(s1, B)`, we would have node_evals be `[eval(A), eval(C), eval(B)]` regardless of the values (well except if two of them evaluate to the same value, we would not add redundant evaluations!)

I think we could instead have only one struct (to "hide" the node_evals utility vec) and implement methods on it (e.g. `insert(val: QuadFelt) -> Option<usize>` to insert a new value and return its index if present), please let me know if you think it's better!

mir/src/passes/unrolling/match_optimizer.rs
Leo-Besancon: Two things regarding this:
- In the current code we cannot have BusOp without Enf, so I've updated the code to have an unreachable there.
  - In boundary constraints, we do not have BusOp, we have binary operations "Eq" between bounded symbol accesses and values
  - In integrity constraints, `p.insert / p.remove` are translated into `Op::Enf(Op::BusOp( ... ))`
- For simplicity, I unwrap the `Enf` there, and re-wrap it with a new `Enf` in `gather_all_constraints`. Otherwise mutating the child of the Enf would be a (little) bit trickier. I've added comments explaining this, but if you think it's cleaner to push the whole constraint with the `Enf` I can do it!

Leo-Besancon:
> Looks good! Thank you! I left some comments inline - once these are addressed, I think we can merge.

Great, I've added comments and answered your questions here, don't hesitate to tell me if something is still unclear!

mir/src/passes/unrolling/match_optimizer.rs
bobbinth: Makes sense! Thank you for the explanation. I think this is fine for now and we can revisit this in the future, if needed.

PR #437, opened 2025-08-06
Add default compilation pipeline

### LINKED ISSUE #426: Provide a default compilation pipeline for ease of use
We should allow the user to instantiate a default pipeline which would do all of the compilation step without knowing what to do or in which order. Maybe it could look something like:
```rust
let ast = parse(source.as_str()).expect("Parsing failed");
let pipeline = mir::default_pipeline();
let air = pipeline.run(ast);
```

_Originally posted by @bobbinth in https://github.com/0xMiden/air-script/pull/424#discussion_r2220684757_

Leo-Besancon:
This PR aims to close https://github.com/0xMiden/air-script/issues/426.

In `air-script/src/lib.rs`, I export the function `ast_to_air_pipeline` that provides a default compilation pipeline.
In its implementation, I had two approaches:
- the first one that simply chained all the passes together (commented out) and returned them
- the second one, that abstracts the passes for each compilation stage (AST / MIR / AIR). It chains black-boxed `AstPasses`, `MirPasses`, `AirPasses`. This is useful as if we need to update the passes for a given crate, we do not need to update code outside it. However, the borrow check wouldn't let me neatly chain the passes there (as Pass is not dyn-compatible), so I had to create new structs (`AstPasses`, `MirPasses`, `AirPasses`) and implement the Pass trait on them manually.

Which implementation should we keep?

### DIFF
--- a/air-script/README.md
+++ b/air-script/README.md
@@ -17,8 +17,7 @@ The compiler has four stages, which can be imported and used independently or to
 Example usage:
 
 ```Rust
-use air_ir::ast_to_air_pipeline;
-use air_script::{parse, Pass, WinterfellCodeGenerator};
+use air_script::{parse, compile, WinterfellCodeGenerator};
 use miden_diagnostics::{
     term::termcolor::ColorChoice, CodeMap, DefaultEmitter, DiagnosticsHandler,
 };
@@ -31,11 +30,8 @@ let diagnostics = DiagnosticsHandler::new(Default::default(), codemap.clone(), e
 // Parse into AST
 let ast = parse(&diagnostics, codemap, source.as_str()).expect("parsing failed");
 
-// Lower to MIR, then AIR
-let air = {
-   let mut pipeline = ast_to_air_pipeline(&diagnostics);
-   pipeline.run(ast).expect("lowering failed")
-};
+// Compile AST into AIR
+let air = compile(&diagnostics, ast).expect("compilation failed");
 
 // Generate Rust code targeting the Winterfell prover
 let code = WinterfellCodeGenerator.generate(&air).expect("codegen failed");
--- a/air-script/README.md
+++ b/air-script/README.md
@@ -17,6 +17,7 @@ The compiler has four stages, which can be imported and used independently or to
 Example usage:
 
 ```Rust
+use air_ir::ast_to_air_pipeline;
 use air_script::{parse, Pass, WinterfellCodeGenerator};
 use miden_diagnostics::{
     term::termcolor::ColorChoice, CodeMap, DefaultEmitter, DiagnosticsHandler,
@@ -30,20 +31,10 @@ let diagnostics = DiagnosticsHandler::new(Default::default(), codemap.clone(), e
 // Parse into AST
 let ast = parse(&diagnostics, codemap, source.as_str()).expect("parsing failed");
 
-// Lower to MIR
-let mir = {
-   let mut pipeline = air_parser::transforms::ConstantPropagation::new(&diagnostics)
-      .chain(mir::passes::AstToMir::new(&diagnostics))
-      .chain(mir::passes::Inlining::new(&diagnostics))
-      .chain(mir::passes::Unrolling::new(&diagnostics));
-   pipeline.run(ast).expect("lowering failed")
-};
-
-// Lower to AIR
+// Lower to MIR, then AIR
 let air = {
-   let mut pipeline = air_ir::passes::MirToAir::new(&diagnostics)
-      .chain(air_ir::passes::BusOpExpand::new(&diagnostics));
-   pipeline.run(mir).expect("lowering failed")
+   let mut pipeline = ast_to_air_pipeline(&diagnostics);
+   pipeline.run(ast).expect("lowering failed")
 };
 
 // Generate Rust code targeting the Winterfell prover
--- a/air/README.md
+++ b/air/README.md
@@ -14,11 +14,8 @@ Example usage:
 // parse the source string to a Result containing the AST or an Error
 let ast = parse(source.as_str()).expect("Parsing failed");
 
-// Create the compilation pipeline needed to translate the AST to the MIR and then the AIR
-let mut pipeline = ast_to_air_pipeline(&diagnostics);
-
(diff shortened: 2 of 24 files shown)


air-script/README.md
```diff
@@ -30,20 +31,10 @@ let diagnostics = DiagnosticsHandler::new(Default::default(), codemap.clone(), e
...
-
-// Lower to AIR
+// Lower to MIR, then AIR
 let air = {
-   let mut pipeline = air_ir::passes::MirToAir::new(&diagnostics)
-      .chain(air_ir::passes::BusOpExpand::new(&diagnostics));
-   pipeline.run(mir).expect("lowering failed")
+   let mut pipeline = ast_to_air_pipeline(&diagnostics);
+   pipeline.run(ast).expect("lowering failed")
 };
```
bobbinth: Could we make it even simpler? Maybe just a single function, similar to how we have `parse()`. Then, these line would look like:
```rust
// Parse into AST
let ast = parse(&diagnostics, codemap, source.as_str()).expect("parsing failed");

// Compile AST into AIR
let air = compile(&diagnostics, ast).expect("compilation failed");
```
The `compile()` function could be imported from the `air_script` namespace.

air/README.md
```diff
@@ -14,13 +14,8 @@ Example usage:
...
-  .chain(mir::passes::AstToMir::new(&diagnostics))
-  .chain(mir::passes::Inlining::new(&diagnostics))
-  .chain(mir::passes::Unrolling::new(&diagnostics))
-  .chain(air_ir::passes::MirToAir::new(&diagnostics))
-  .chain(air_ir::passes::BusOpExpand::new(&diagnostics));
+// Create the compilation pipeline needed to translate the AST to the MIR and then the AIR
+let mut pipeline = ast_to_air_pipeline(&diagnostics);
 
 // process the AST and the MIR to get a Result containing the AIR or a CompileError
 let air = pipeline.run(ast)
```
bobbinth: Same comment about potentially using just a single `compile()` function as above.

codegen/winterfell/README.md
```diff
@@ -16,13 +16,8 @@ Example usage:
...
-  .chain(mir::passes::AstToMir::new(&diagnostics))
-  .chain(mir::passes::Inlining::new(&diagnostics))
-  .chain(mir::passes::Unrolling::new(&diagnostics))
-  .chain(air_ir::passes::MirToAir::new(&diagnostics))
-  .chain(air_ir::passes::BusOpExpand::new(&diagnostics));
+// Create the compilation pipeline needed to translate the AST to the MIR and then the AIR
+let mut pipeline = ast_to_air_pipeline(&diagnostics);
 
 // process the AST and the MIR to get a Result containing the AIR or a CompileError
 let air = pipeline.run(ast)
```
bobbinth: Same comment as above.

bobbinth (commented):
Looks good! Thank you! I left a couple of nits inline about how we could simplify it even further from the user's perspective.

> * the first one that simply chained all the passes together (commented out) and returned them
> * the second one, that abstracts the passes for each compilation stage (AST / MIR / AIR). It chains black-boxed `AstPasses`, `MirPasses`, `AirPasses`. This is useful as if we need to update the passes for a given crate, we do not need to update code outside it. However, the borrow check wouldn't let me neatly chain the passes there (as Pass is not dyn-compatible), so I had to create new structs (`AstPasses`, `MirPasses`, `AirPasses`) and implement the Pass trait on them manually.

I don't have a strong opinion here mostly because I'm not sure I understand the tradeoffs. What is the main downside of the approach we have now vs. the commented-out approach (and vice-versa)?

I think as long as we can provide a high-level function for users to call (e.g., something like the `compile()` function I mentioned in the inline comments), the internal implementation can be flexible.

Leo-Besancon:
> I don't have a strong opinion here mostly because I'm not sure I understand the tradeoffs. What is the main downside of the approach we have now vs. the commented-out approach (and vice-versa)?

For the user, it won't change anything (even more as we provide the `compile()` helper you suggested!)

On the dev side:
- the current approach is a bit more involved and requires to implement 3 more structs (`AstPasses`, `MirPasses`, `AirPasses`). For now `AstPasses` is even a bit useless as it only contains 1 pass, but the goal is to abstract it away.
- the commented-out approach requires that a dev that e.g. adds a Mir pass to the pipeline, adds it everywhere it is called (in mir tests, air tests, air-script crate, air-script tests..), whereas with the current approach they would just update "MirPasses" once.

air-script/README.md
Leo-Besancon: I've added the compile function, thank you for the suggestion!

bobbinth (approved):
Looks good! Thank you!

> * the current approach is a bit more involved and requires to implement 3 more structs (`AstPasses`, `MirPasses`, `AirPasses`). For now `AstPasses` is even a bit useless as it only contains 1 pass, but the goal is to abstract it away.
> * the commented-out approach requires that a dev that e.g. adds a Mir pass to the pipeline, adds it everywhere it is called (in mir tests, air tests, air-script crate, air-script tests..), whereas with the current approach they would just update "MirPasses" once.

My understanding is that the current approach is a bit more easily "extensible" - that is, we can add more passes and that would be encapsulated in one place rather than having to do many changes throughout the codebase. If so, I think the complexity is not that big - and so, I would keep it (and remove the commented-out portion).

Leo-Besancon:
> Looks good! Thank you!
> 
> > * the current approach is a bit more involved and requires to implement 3 more structs (`AstPasses`, `MirPasses`, `AirPasses`). For now `AstPasses` is even a bit useless as it only contains 1 pass, but the goal is to abstract it away.
> > * the commented-out approach requires that a dev that e.g. adds a Mir pass to the pipeline, adds it everywhere it is called (in mir tests, air tests, air-script crate, air-script tests..), whereas with the current approach they would just update "MirPasses" once.
> 
> My understanding is that the current approach is a bit more easily "extensible" - that is, we can add more passes and that would be encapsulated in one place rather than having to do many changes throughout the codebase. If so, I think the complexity is not that big - and so, I would keep it (and remove the commented-out portion).

Perfect, I've removed it.

adr1anh (approved):
LGTM, it's nice to have a single pipeline definition.

Issue #379, opened 2025-04-18
Design and implement support multiple main traces


Leo-Besancon:
In comments of PR https://github.com/0xPolygonMiden/air-script/pull/375, several use cases have been discussed in favor of supporting multiple main trace segments in AirScript.

bobbinth:
I don't have a clear idea yet of how this could look like in AirScript but i think we do need to be able to support this, especially once we move to the Plonky3 backend. One use case for this is to have separate tables for the "main" trace and for the "chiplets" trace. It could look something like this:

```
trace_columns {
    main: [clk, ...],
    chiplets: [s0, s1, s2, ....],
}
```
A couple of outstanding questions:

- What syntax can we use to tie these together? This can probably be done via busses, but we may need a new bus type for this. Or maybe this would need to be another higher level construct.
- In general, how would this affect the busses? We may need to define which trace a given bus belongs to. So, maybe there is a better overall syntax for this.
- How would this work in evaluators? I'm assuming that evaluators usually work with a single table - but maybe there is a need to pass two or more segments to evaluators?

I think @adr1anh is probably the person who's most familiar with multi-table traces - so, curious what you think.

bobbinth:
> * What syntax can we use to tie these together? This can probably be done via busses, but we may need a new bus type for this. Or maybe this would need to be another higher level construct.
> * In general, how would this affect the busses? We may need to define which trace a given bus belongs to. So, maybe there is a better overall syntax for this.

To support multiple main traces (we should call them "trace tables"), we'll need to re-work how busses work. Currently, we have two potential options:

1. Make buses purely logical constructs and then have the compiler figure out how to lay them out in in physical trace columns. This is described in https://github.com/0xMiden/air-script/issues/422#issue-3232262336.
2. Add ability to specify which trace table each bus can interact with https://github.com/0xMiden/air-script/issues/422#issuecomment-3146978590.

The first approach is a bit more involved but may be easier for the user to work with. The second approach is much more manual but should be simpler to implement.

> * How would this work in evaluators? I'm assuming that evaluators usually work with a single table - but maybe there is a need to pass two or more segments to evaluators?

My current thinking is that each evaluator still only works with a single table. Any constraints that involve multiple tables would go through the busses anyway.

Issue #422, opened 2025-07-15
Revised bus boundary constraints


adr1anh:
This issue describes an alternative implementation of buses which may lead to simplified code generation and verifier implementation. We will assume all our buses are based on logUp, as this is the most natural choice when dealing with multiple tables.

**TLDR**
- Expose a single bus in frontend used for all messages
- Multiple buses can still be supported if necessary, in which case the compiler can inject an additional domain separator at the start of all messages (i.e. the message is `[BUS_ID, m0, m1, ...]`).
- The compiler determines the number of bus columns depending on the desired blowup factor and the degree of the expressions.
- Bus columns are "batched". Each column represents the insertion or removal of multiple queries. 
- All bus columns constrain the value in the first row to be 0.
- The prover sends the final logUp value of each bus as part of the proof. 
- Each column is constrained in the last row to equal the final value provided by the prover.
- Public input tables no longer need to be treated as variables representing a bus boundary value. The IR still exposes them to facilitate code generation. 
- The verifier is responsible for checking that the sum of 
	- the randomly reduced logUp entries of all public input tables, and
	- all the final bus boundary values provided in the proof,
	evaluates to zero.

#### Connecting all busses

At the moment, busses by default are independent of one another, unless explicitly connected with a potential `enf busA.last = busB.first` type of syntax. This exposes unnecessary bus implementation details. 

Instead, we can assume that all buses are connected, even though they may contain messages used in different contexts. Consider the case where we have two separate buses
- **`range`**: The message `[value]` is a single `Felt` which is sent to the range-check table.
- **`chiplet_op`**: A message `[INSTRUCTION_LABEL, arg0, arg1, ...]` containing an instruction label and a variable number of arguments, which is sent to one of the chiplets.
We can combine these into a single bus by adding the bus name as part of the message, i.e. 
- `[RANGE_BUS_LABEL, value]`
- `[CHIPLET_OP_BUS_LABEL, INSTRUCTION_LABEL, arg0, arg1, ...]`
As long as the `*_BUS_LABEL`s are unique, messages from different buses will never collide.

#### Batching buses

Consider a chiplet which enforces `a + b = c`. 
```
trace_columns {  
    main: [  
        s,          // row activation  
        a, b, c,    // arguments   
        m,          // operation multiplicity  
    ],  
}  
  
buses {  
    range,  
    chip,  
}  
  
integrity_constraints {  
    // row is active or not  
    enf s^2 = s;  
  
    // enforce arguments are u32  
[...]

PR #438, opened 2025-08-06
Define TraceSegment enum for Main and Aux traces

### LINKED ISSUE #406: Define TraceSegment enum for `Main` and `Aux` traces
When handling constraints on given trace segments, instead of referencing its id value (0 or 1), we should instead use enums such as:

```rust
pub enum TraceSegment {
    Main,
    Aux,
}
```

Note: we may need to adapt it based on potential needs for multiple main traces: https://github.com/0xMiden/air-script/issues/379

_Originally posted by @bobbinth in https://github.com/0xMiden/air-script/pull/402#discussion_r2147382004_

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/406.

For now, I've kept only `Main` and `Aux` variants without a way to specify multiple main traces.
It should not cause issues to do it later when https://github.com/0xMiden/air-script/issues/379 is better scoped

There is some awkwardness as we still handle `Vec<TraceSegment>` in many structures where only the main trace would be valid, but I've tried to keep the changes of this PR to a minimum, to be adapted later depending on https://github.com/0xMiden/air-script/issues/379.

### DIFF
--- a/air/src/graph/mod.rs
+++ b/air/src/graph/mod.rs
@@ -103,29 +103,29 @@ impl AlgebraicGraph {
         // recursively walk the subgraph and infer the trace segment and domain
         match self.node(index).op() {
             Operation::Value(value) => match value {
-                Value::Constant(_) => Ok((DEFAULT_SEGMENT, default_domain)),
-                Value::RandomValue(_) => Ok((AUX_SEGMENT, default_domain)),
+                Value::Constant(_) => Ok((TraceSegmentId::Main, default_domain)),
+                Value::RandomValue(_) => Ok((TraceSegmentId::Aux, default_domain)),
                 Value::PeriodicColumn(_) => {
                     assert!(
                         !default_domain.is_boundary(),
                         "unexpected access to periodic column in boundary constraint"
                     );
                     // the default domain for [IntegrityConstraints] is `EveryRow`
-                    Ok((DEFAULT_SEGMENT, ConstraintDomain::EveryRow))
+                    Ok((TraceSegmentId::Main, ConstraintDomain::EveryRow))
                 },
                 Value::PublicInput(_) => {
                     assert!(
                         !default_domain.is_integrity(),
                         "unexpected access to public input in integrity constraint"
                     );
-                    Ok((DEFAULT_SEGMENT, default_domain))
+                    Ok((TraceSegmentId::Main, default_domain))
                 },
                 Value::PublicInputTable(_) => {
                     assert!(
                         !default_domain.is_integrity(),
                         "unexpected access to public input table in integrity constraint"
                     );
-                    Ok((DEFAULT_SEGMENT, default_domain))
+                    Ok((TraceSegmentId::Main, default_domain))
                 },
                 Value::TraceAccess(trace_access) => {
                     let domain = if default_domain.is_boundary() {
@@ -145,7 +145,10 @@ impl AlgebraicGraph {
                 let (lhs_segment, lhs_domain) = self.node_details(lhs, default_domain)?;
                 let (rhs_segment, rhs_domain) = self.node_details(rhs, default_domain)?;
 
-                let trace_segment = lhs_segment.max(rhs_segment);
+                let trace_segment = match (lhs_segment, rhs_segment) {
+                    (TraceSegmentId::Aux, _) | (_, TraceSegmentId::Aux) => TraceSegmentId::Aux,
+                    _ => TraceSegmentId::Main,
+                };
                 let domain = lhs_domain.merge(rhs_domain)?;
 
                 Ok((trace_segment, domain))
--- a/air/src/graph/mod.rs
+++ b/air/src/graph/mod.rs
@@ -145,10 +145,7 @@ impl AlgebraicGraph {
                 let (lhs_segment, lhs_domain) = self.node_details(lhs, default_domain)?;
                 let (rhs_segment, rhs_domain) = self.node_details(rhs, default_domain)?;
 
-                let trace_segment = match (lhs_segment, rhs_segment) {
-                    (TraceSegmentId::Aux, _) | (_, TraceSegmentId::Aux) => TraceSegmentId::Aux,
-                    _ => TraceSegmentId::Main,
-                };
+                let trace_segment = lhs_segment.max(rhs_segment);
                 let domain = lhs_domain.merge(rhs_domain)?;
 
                 Ok((trace_segment, domain))
--- a/air/src/passes/translate_from_mir.rs
+++ b/air/src/passes/translate_from_mir.rs
@@ -34,16 +34,20 @@ impl Pass for MirToAir<'_> {
 
         let buses = mir.constraint_graph().buses.clone();
 
-        if mir.trace_columns.len() != 1 {
-            panic!("Expected one trace segment, but found multiple: {:?}", mir.trace_columns);
-        }
+        assert!(
+            mir.trace_columns.len() == 1,
+            "Expected one trace segment, but found: {:?}",
+            mir.trace_columns
+        );
         let main_trace_segment = mir.trace_columns.first().unwrap();
-        if main_trace_segment.id != TraceSegmentId::Main {
-            panic!(
-                "Expected trace segment to be the main segment, but found: {:?}",
-                main_trace_segment.id
-            );
-        }
+
+        assert_eq!(
+            main_trace_segment.id,
+            TraceSegmentId::Main,
+            "Expected trace segment to be the main segment, but found: {:?}",
+            main_trace_segment.id
+        );
+
         let mut trace_columns = BTreeMap::new();
         trace_columns.insert(main_trace_segment.id, main_trace_segment.clone());
 
@@ -58,7 +62,10 @@ impl Pass for MirToAir<'_> {
             let aux_trace_segment = TraceSegment::new(
                 SourceSpan::default(),
                 TraceSegmentId::Aux,
-                Identifier::new(SourceSpan::default(), Symbol::new(TraceSegmentId::Aux as u32)),
+                Identifier::new(
+                    SourceSpan::default(),
+                    Symbol::new(TraceSegmentId::Aux.index() as u32),
+                ),
                 bus_raw_bindings,
             );
             for binding in aux_trace_segment.bindings.iter() {
@@ -445,7 +452,7 @@ impl AirBuilder<'_> {
                 if let Some(prev) = self
                     .trace_columns
                     .get_mut(&trace_access.segment)
-                    .unwrap()
+                    .expect("Boundary constraint on an unknown trace segment")
                     .mark_constrained(lhs_span, trace_access.column, boundary.kind)
                 {
                     self.diagnostics
--- a/air/src/passes/translate_from_mir.rs
+++ b/air/src/passes/translate_from_mir.rs
@@ -34,33 +34,40 @@ impl Pass for MirToAir<'_> {
 
         let buses = mir.constraint_graph().buses.clone();
 
-        let mut trace_columns = mir.trace_columns.clone();
+        if mir.trace_columns.len() != 1 {
+            panic!("Expected one trace segment, but found multiple: {:?}", mir.trace_columns);
+        }
+        let main_trace_segment = mir.trace_columns.first().unwrap();
+        if main_trace_segment.id != TraceSegmentId::Main {
+            panic!(
+                "Expected trace segment to be the main segment, but found: {:?}",
+                main_trace_segment.id
+            );
+        }
+        let mut trace_columns = BTreeMap::new();
+        trace_columns.insert(main_trace_segment.id, main_trace_segment.clone());
 
         let mut bus_bindings_map = BTreeMap::new();
         if !buses.is_empty() {
             let bus_raw_bindings: Vec<_> = buses
                 .keys()
-                .map(|k| Span::new(k.span(), (Identifier::new(k.span(), k.name()), AUX_SEGMENT)))
+                .map(|k| Span::new(k.span(), (Identifier::new(k.span(), k.name()), 1)))
                 .collect();
 
             // Add buses as `aux` trace columns
             let aux_trace_segment = TraceSegment::new(
                 SourceSpan::default(),
-                AUX_SEGMENT,
-                Identifier::new(SourceSpan::default(), Symbol::new(AUX_SEGMENT as u32)),
+                TraceSegmentId::Aux,
+                Identifier::new(SourceSpan::default(), Symbol::new(TraceSegmentId::Aux as u32)),
                 bus_raw_bindings,
             );
             for binding in aux_trace_segment.bindings.iter() {
                 bus_bindings_map.insert(binding.name.unwrap(), binding.offset);
             }
-            if trace_columns.len() == 1 {
-                trace_columns.push(aux_trace_segment);
-            } else {
-                panic!("Expected only one trace segment, but found multiple: {trace_columns:?}",);
-            }
+            trace_columns.insert(aux_trace_segment.id, aux_trace_segment);
         }
 
-        air.trace_segment_widths = trace_columns.iter().map(|ts| ts.size as u16).collect();
+        air.trace_segment_widths = trace_columns.values().map(|ts| ts.size as u16).collect();
         air.num_random_values = mir.num_random_values;
         air.periodic_columns = mir.periodic_columns.clone();
         air.public_inputs = mir.public_inputs.clone();
@@ -98,7 +105,7 @@ impl Pass for MirToAir<'_> {
 struct AirBuilder<'a> {
     diagnostics: &'a DiagnosticsHandler,
     air: &'a mut Air,
-    trace_columns: Vec<TraceSegment>,
+    trace_columns: BTreeMap<TraceSegmentId, TraceSegment>,
     bus_bindings_map: BTreeMap<Identifier, usize>,
 }
 
@@ -264,7 +271,7 @@ impl AirBuilder<'_> {
                         let name = bus_access.bus.borrow().deref().name();
                         let column = self.bus_bindings_map.get(&name).unwrap();
                         crate::ir::Value::TraceAccess(crate::ir::TraceAccess {
-                            segment: AUX_SEGMENT,
+                            segment: TraceSegmentId::Aux,
                             column: *column,
                             row_offset: bus_access.row_offset,
                         })
@@ -320,7 +327,7 @@ impl AirBuilder<'_> {
                         let name = bus_access.bus.borrow().deref().name();
                         let column = self.bus_bindings_map.get(&name).unwrap();
                         crate::ir::Value::TraceAccess(crate::ir::TraceAccess {
-                            segment: AUX_SEGMENT,
+                            segment: TraceSegmentId::Aux,
                             column: *column,
                             row_offset: offset,
                         })
@@ -422,8 +429,11 @@ impl AirBuilder<'_> {
                         let bus = bus_access.bus;
                         let name = bus.borrow().deref().name();
                         let column = self.bus_bindings_map.get(&name).unwrap();
-                        let trace_access =
-                            mir::ir::TraceAccess::new(AUX_SEGMENT, *column, bus_access.row_offset);
+                        let trace_access = mir::ir::TraceAccess::new(
+                            TraceSegmentId::Aux,
+                            *column,
+                            bus_access.row_offset,
+                        );
                         (trace_access, lhs_span)
                     },
                     _ => unreachable!(
@@ -432,11 +442,12 @@ impl AirBuilder<'_> {
                     ), // Raise diag
                 };
 
-                if let Some(prev) = self.trace_columns[trace_access.segment].mark_constrained(
-                    lhs_span,
-                    trace_access.column,
-                    boundary.kind,
-                ) {
+                if let Some(prev) = self
+                    .trace_columns
+                    .get_mut(&trace_access.segment)
+                    .unwrap()
+                    .mark_constrained(lhs_span, trace_access.column, boundary.kind)
+                {
                     self.diagnostics
                         .diagnostic(Severity::Error)
                         .with_message("overlapping boundary constraints")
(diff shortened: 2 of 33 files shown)


codegen/ace/src/tests/quotient.rs
```diff
@@ -66,7 +66,11 @@ pub fn eval_quotient(air: &Air, ace_vars: &AceVars, log_trace_len: u32) -> QuadF
             Operation::Value(v) => match v {
                 Value::Constant(c) => QuadFelt::from(Felt::new(c)),
                 Value::TraceAccess(access) => {
-                    ace_vars.segments[access.row_offset][access.segment][access.column]
+                    let segment_index = match access.segment {
```
adr1anh: Can we add a `TraceSegmentId::index()` method?

Leo-Besancon: Indeed, I've added it in 3fd4ad5b15d04a63e90590749a104d293d158f89.

I wanted your thoughts on it though. In other parts of the code, I've refactored it to index directly with TraceSegmentId to avoid the need of this method and mixing ids and usizes. For example, in the Air `Constraints` struct:
```rust
boundary_constraints: Vec<Vec<ConstraintRoot>>
```
becomes
```rust
boundary_constraints: BTreeMap<TraceSegmentId, Vec<ConstraintRoot>>
``` 

For ACE, I've left it as is and used this `index()` method, because it seems hardcoded into the layout (https://github.com/massalabs/air-script/blob/add_trace_segment_enum/codegen/ace/src/layout.rs#L60), and mixed with the quotient.

Would it be cleaner to instead have something like `AceSegmentId` below and index by it?

```rust
pub enum AceSegmentId {
	AirSegment(TraceSegmentId),
	Quotient,
}
```

And instead of `pub trace_segments: [[InputRegion; 3]; 2],`, do something like `[BTreeMap<AceSegmentId, InputRegion>; 2]`?

air/src/graph/mod.rs
```diff
@@ -145,7 +145,10 @@ impl AlgebraicGraph {
                 let (lhs_segment, lhs_domain) = self.node_details(lhs, default_domain)?;
                 let (rhs_segment, rhs_domain) = self.node_details(rhs, default_domain)?;
 
-                let trace_segment = lhs_segment.max(rhs_segment);
+                let trace_segment = match (lhs_segment, rhs_segment) {
+                    (TraceSegmentId::Aux, _) | (_, TraceSegmentId::Aux) => TraceSegmentId::Aux,
+                    _ => TraceSegmentId::Main,
+                };
```
bobbinth: Could we encapsulate this into a method on `TraceSegmentId`? Maybe something like `TraceSegmentId::max()` (though, maybe there is a better name for this).

parser/src/ast/trace.rs
```diff
@@ -5,7 +5,30 @@ use miden_diagnostics::{SourceSpan, Spanned};
 use super::*;
 
 /// The id of a trace segment is its index in the trace_columns declaration
-pub type TraceSegmentId = usize;
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum TraceSegmentId {
+    Main,
+    Aux,
+}
```
bobbinth: Should we call this just `TraceSegment`? Or would this collide with some other name we have?

parser/src/parser/grammar.lalrpop
```diff
@@ -195,14 +195,22 @@ EvaluatorFunction: EvaluatorFunction = {
 
 EvaluatorBindings: Vec<TraceSegment> = {
     <l:@L> <trace: Comma<EvaluatorSegmentBindings>> <r:@R> =>? {
-        let mut segments = Vec::with_capacity(trace.len());
+        if trace.len() > 1 {
```
bobbinth: Should this be `if trace.len() != 1`? Basically, shouldn't we always have exactly one trace segment?

parser/src/parser/grammar.lalrpop
```diff
@@ -195,14 +195,22 @@ EvaluatorFunction: EvaluatorFunction = {
...
+            diagnostics.diagnostic(Severity::Error)
+                .with_message("invalid evaluator function definition")
+                .with_primary_label(span!(l, r), "evaluators cannot have more than one trace segment")
+                .emit();
+            return Err(ParseError::Failed.into());
+        }
 
-        for (segment, (span, bindings)) in trace.into_iter().enumerate() {
+        let mut segments = Vec::with_capacity(trace.len());
```
bobbinth, line 206: Related to the above comment: shouldn't it be just one trace segment instead of a vector of segments?

codegen/ace/src/layout.rs
```diff
@@ -207,12 +208,12 @@ impl Layout {
...
-        // We should only be able to access the main and aux segments.
-        if segment > 1 {
-            return None;
+        let segment_index = match segment {
+            TraceSegmentId::Main => 0,
+            TraceSegmentId::Aux => 1,
         };
         let segments_in_row = self.trace_segments.get(row_offset)?;
-        let segment_region = segments_in_row.get(segment)?;
+        let segment_region = segments_in_row.get(segment_index)?;
```
bobbinth: We can now use `segment::index()` here, right?

mir/src/passes/translate.rs
```diff
@@ -1281,10 +1285,10 @@ impl<'a> MirBuilder<'a> {
     // Check assumptions, probably this assumed that the inlining pass did some work
     fn trace_access(&self, access: &ast::SymbolAccess) -> Option<TraceAccess> {
         let id = access.name.as_ref();
-        for (i, segment) in self.trace_columns.iter().enumerate() {
+        if let Some(segment) = self.trace_columns.first() {
```
bobbinth: Should we add an `assert!` to make sure that `self.trace_columns` has only one segment?

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -86,8 +86,11 @@ fn add_aux_trace_assertions(func_body: &mut codegen::Function, ir: &Air) {
     // TODO: These values are constant across all rows and therefore can be computed only once
     //       before starting the constraint evaluation.
     for access in ir.reduced_public_input_table_accesses() {
-        let boundary_value =
-            air_ir::Value::PublicInputTable(access).to_string(ir, ElemType::Ext, 0);
+        let boundary_value = air_ir::Value::PublicInputTable(access).to_string(
+            ir,
+            ElemType::Ext,
+            TraceSegmentId::Aux,
```
bobbinth, line 88: Was this a bug previously? (we used to pass `0`, but now passing `TraceSegmentId::Aux`)

air/src/passes/translate_from_mir.rs
```diff
@@ -34,33 +34,40 @@ impl Pass for MirToAir<'_> {
 
         let buses = mir.constraint_graph().buses.clone();
 
-        let mut trace_columns = mir.trace_columns.clone();
+        if mir.trace_columns.len() != 1 {
+            panic!("Expected one trace segment, but found multiple: {:?}", mir.trace_columns);
+        }
```
bobbinth: nit: this could be an `assert!` right?

air/src/passes/translate_from_mir.rs
```diff
@@ -34,33 +34,40 @@ impl Pass for MirToAir<'_> {
...
+        let main_trace_segment = mir.trace_columns.first().unwrap();
+        if main_trace_segment.id != TraceSegmentId::Main {
+            panic!(
+                "Expected trace segment to be the main segment, but found: {:?}",
+                main_trace_segment.id
+            );
```
bobbinth: Same nit as above.

air/src/passes/translate_from_mir.rs
```diff
@@ -34,33 +34,40 @@ impl Pass for MirToAir<'_> {
...
+                .map(|k| Span::new(k.span(), (Identifier::new(k.span(), k.name()), 1)))
                 .collect();
 
             // Add buses as `aux` trace columns
             let aux_trace_segment = TraceSegment::new(
                 SourceSpan::default(),
-                AUX_SEGMENT,
-                Identifier::new(SourceSpan::default(), Symbol::new(AUX_SEGMENT as u32)),
+                TraceSegmentId::Aux,
+                Identifier::new(SourceSpan::default(), Symbol::new(TraceSegmentId::Aux as u32)),
```
bobbinth: Could this be `TraceSegmentId::Aux::index()`?

air/src/passes/translate_from_mir.rs
```diff
@@ -34,33 +34,40 @@ impl Pass for MirToAir<'_> {
...
+        }
+        let mut trace_columns = BTreeMap::new();
+        trace_columns.insert(main_trace_segment.id, main_trace_segment.clone());
 
         let mut bus_bindings_map = BTreeMap::new();
         if !buses.is_empty() {
             let bus_raw_bindings: Vec<_> = buses
                 .keys()
-                .map(|k| Span::new(k.span(), (Identifier::new(k.span(), k.name()), AUX_SEGMENT)))
+                .map(|k| Span::new(k.span(), (Identifier::new(k.span(), k.name()), 1)))
```
bobbinth, line 58: Should we replace `1` here with `TraceSegmentId::Aux::index()`?

air/src/passes/translate_from_mir.rs
```diff
@@ -432,11 +442,12 @@ impl AirBuilder<'_> {
...
 
-                if let Some(prev) = self.trace_columns[trace_access.segment].mark_constrained(
-                    lhs_span,
-                    trace_access.column,
-                    boundary.kind,
-                ) {
+                if let Some(prev) = self
+                    .trace_columns
+                    .get_mut(&trace_access.segment)
+                    .unwrap()
```
bobbinth: nit: I would replace `unwrap()` with `expect()`.

parser/src/ast/trace.rs
```diff
@@ -5,7 +5,30 @@ use miden_diagnostics::{SourceSpan, Spanned};
...
+        match self {
+            TraceSegmentId::Main => write!(f, "main"),
+            TraceSegmentId::Aux => write!(f, "aux"),
+        }
+    }
+}
+
+impl TraceSegmentId {
+    /// Returns the index of this segment
+    pub fn index(&self) -> usize {
```
bobbinth: nit: could this be `const fn`?

bobbinth (commented):
Looks good! Thank you! I left some comments inline.

parser/src/ast/trace.rs
Leo-Besancon: It would indeed collide with the struct that describes the different columns a given trace_segment has: https://github.com/massalabs/air-script/blob/add_trace_segment_enum/parser/src/ast/trace.rs#L37

codegen/winterfell/src/air/boundary_constraints.rs
Leo-Besancon: I think so, my reasonning is that as PublicInputTable can only be used for buses, they should only be valid in the aux trace. But because of the `max()` it did not have any impact.

What do you think?

air/src/passes/translate_from_mir.rs
Leo-Besancon: Yes, I still need the `as u32` as index() returns a `usize`, but it makes more sence.

air/src/passes/translate_from_mir.rs
Leo-Besancon: The 1 here refers to the size of the column binding (e.g. when we declare `main: [s[3]]`, not to the Id of the TraceSegment (it was a bug previously but as the value was the same it did not have any impact).

air/src/graph/mod.rs
Leo-Besancon: We can even use .max() provided by the `#[derive(Ord)]`. Not sure if there would be a reason to not use it. (e.g. when we have potentially multiple main traces).

parser/src/parser/grammar.lalrpop
Leo-Besancon: Indeed, currently it's the case (it cannot even parse an empty list such as `ev ev_func() {}`.
I was thinking, would there be any use at all to allow no trace here, as evaluators still have access to buses without passing them as argument? It does not seem very useful, but maybe it can still be valid?

```
ev ev_func() {
  # No main trace usage
  q.insert(1, 3) for 2;
}
```

parser/src/parser/grammar.lalrpop
Leo-Besancon: This is basically related to legacy code that I did not refactor yet, waiting to see if https://github.com/0xMiden/air-script/issues/379 has impact on this. If we still want to do this, I think it will need to be in another PR.

Vec<TraceSegement> are still found in two places where they are not needed for now, both in the AST and in the MIR:
- to keep track of a program's traces before the Air
- for evaluators parameters

parser/src/ast/trace.rs
bobbinth: Makes sense! I think what we currently call `TraceSegment` would probably need to be eventually renamed into something like `TraceTable`, and then we can rename `TraceSegmentId` to just `TraceSegment`. But that's for another PR (e.g., when we do something like I described [here](https://github.com/0xMiden/air-script/issues/422#issuecomment-3146978590)).

parser/src/parser/grammar.lalrpop
bobbinth: Yeah - I can't imagine something like that being useful. My current thinking is that we shouldn't support things like this.

@Al-Kindi-0 @adr1anh, what do you guys think?

parser/src/parser/grammar.lalrpop
bobbinth: Makes sense. As I mentioned in the other comment, when we go this route we'll have multiple "trace tables" but each of these tables would still belong to the `Main` trace segment (i.e., we'll still have only 2 segments: `Main` and `Aux`).

codegen/winterfell/src/air/boundary_constraints.rs
bobbinth: Looking at the underlying code of `to_string()` it seems like it doesn't actually matter what we pass as  `TraceSegmentId` (or `ElemType` for that matter) - so, I guess either way is fine.

Question: what does `PublicInputTableAccess` correspond to in the language?

bobbinth: Looking a bit more into this, it may be possible to get rid of the `trace_segment` parameter all together here. It is only used by `TraceAccess::to_string()` and there maybe `ElemType` could be sufficient to communicate the required info?

Leo-Besancon: > Question: what does `PublicInputTableAccess` correspond to in the language?

It corresponds to public inputs used as bus boundaries. They let us differentiate with `PublicInputAccess` that reference regular public inputs, accessible from the main trace.

```
public_inputs {
    # public input table
    inputs: [[4]],
    # public input
    outputs: [4],
}

boundary_constraints {
    enf p.first = inputs;
}
```

Leo-Besancon: > Looking a bit more into this, it may be possible to get rid of the `trace_segment` parameter all together here. It is only used by `TraceAccess::to_string()` and there maybe `ElemType` could be sufficient to communicate the required info?

I'm not sure it's sufficient, we have to know two things (whether we are describing main / aux constraints for one, and the trace we are accessing). I've not yet make it work.

parser/src/ast/trace.rs
bobbinth: nit: should we explicitly set `Main = 0` and `Aux = 1`?

parser/src/parser/grammar.lalrpop
adr1anh: I'm not up-to date on how evaluators work, but I really don't see any point of allowing bus operations using constant data. It provides no extra knowledge other than the number of rows contained in the table.

codegen/ace/src/tests/quotient.rs
adr1anh: This is something we could modify later on, but I also think we don't necessarily need to make it cleaner. Using `BTreeMaps` seems a bit more confusing, as it would suggest that we allow proofs where we are missing either main or aux trace, which we currently don't. If anything, I would see the following struct being useful 

```rust
struct TraceSegments<T> {
    main: T,
    aux: T,
    quotient: T,
}
```

which could be indexed using a `TraceSegmentId`, ie `trace_segments[TraceSegmentId::Main]`. I haven't fully thought this through though yet.

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -45,18 +47,17 @@ pub(super) fn add_fn_get_aux_assertions(impl_ref: &mut Impl, ir: &Air) {
 /// trace segment
 fn add_main_trace_assertions(func_body: &mut codegen::Function, ir: &Air) {
     let elem_type = ElemType::Base;
```
bobbinth: Do we actually need to define this variable? Seems like we only use it once in the code below and we could easily replace that with `ElemType::Base`.

codegen/winterfell/src/air/boundary_constraints.rs
```diff
@@ -73,7 +74,6 @@ fn add_main_trace_assertions(func_body: &mut codegen::Function, ir: &Air) {
 /// trace segment (used for buses boundary constraints for variable length public inputs)
 fn add_aux_trace_assertions(func_body: &mut codegen::Function, ir: &Air) {
     let elem_type = ElemType::Ext;
```
bobbinth: Similar comment as above - I'm not sure we need to define this variable.

codegen/winterfell/src/air/boundary_constraints.rs
bobbinth: My thinking here was that I'm not sure the current definition of the `Coden` trait is very clean. It may be possible to simplify it to something like
```rust
pub trait Codegen {
    fn to_string(&self, ir: &Air, elem_type: ElemType) -> String;
}
```
Or
```rust
pub trait Codegen {
    fn to_string(&self, ir: &Air, trace_segment: TraceSegmentId) -> String;
}
```
But I also haven't thought through this completely, and given that winterfell codegen may not be something we care about in the longer term, we could probably leave it as is.

bobbinth (approved):
Looks good! Thank you! I left a few more small comments inline.

Also, I think we should probably do a couple of follow-ups on this:
- We should get rid of the `$main` keyword.
- I'd love to figure out if we can get rid of `TraceSegmentId::index()` somehow.

Let's create issues for these.

codegen/winterfell/src/air/boundary_constraints.rs
adr1anh: When this was implemented, I wanted to reuse the existing `to_string` method which takes these additional parameters. We don't use them so they could indeed be arbitrary. A `PublicInputTableAccess` really just corresponds to the virtual boundary value of a bus. During the evaluation we only need to consider the public input table name and the bus type such as
`range_check_bus_logUp` or `kernel_rom_multiset`.

adr1anh:
This looks good overall and seems like a good step forward that doesn't affect too much of the existing logic (which really helped with the review, thanks!). 

I agree about the awkwardness surrounding the `Vec<TraceSegment>` and it would be nice to make things clearer. Given that we're now assuming a fixed proof shape with only two traces, it seems like we could replace the `BTreeMap` with a structure like 
```rust
struct TraceShape<T> {
  main: T,
  aux: T,
}
```
The generic makes it easier to reuse across different modules, we can also make it indexable by a `TraceSegmentId`, and wrap it in a `FullTraceShape<T>` which also includes the quotient. I haven't fully thought it out, but it seems like it would work at least for the ACE module which I'm more familiar with.

codegen/ace/src/tests/quotient.rs
Leo-Besancon: > This is something we could modify later on, but I also think we don't necessarily need to make it cleaner. Using `BTreeMaps` seems a bit more confusing, as it would suggest that we allow proofs where we are missing either main or aux trace, which we currently don't. If anything, I would see the following struct being useful
> 
> ```rust
> struct TraceSegments<T> {
>     main: T,
>     aux: T,
>     quotient: T,
> }
> ```
> 
> which could be indexed using a `TraceSegmentId`, ie `trace_segments[TraceSegmentId::Main]`. I haven't fully thought this through though yet.

Indeed I do prefer your suggestion! I'll add a reference to your comment in the followup issue to remove `::index()` that Bobbin talked about below.

mir/src/passes/translate.rs
Leo-Besancon: Sorry I missed that one before, I've added the assert.

Leo-Besancon:
> Also, I think we should probably do a couple of follow-ups on this:
> 
>     * We should get rid of the `$main` keyword.
> 
>     * I'd love to figure out if we can get rid of `TraceSegmentId::index()` somehow.
> 
> 
> Let's create issues for these.

Thanks! I've added the issues here:
https://github.com/0xMiden/air-script/issues/441
https://github.com/0xMiden/air-script/issues/442

codegen/winterfell/src/air/boundary_constraints.rs
Leo-Besancon: I could be wrong but it does not seem possible to me. From what I gather:

- In main boundaries, we need:
	- Consts to be written as `Felt::new(c)`
- In main transition, we need:
	- Consts to be written as `E::from(Felt::new(c)))`
	- Main trace to be written as `main_current[i]`
- In aux transition, we need:
	- Main trace to be written as `E::from(main_current[i])`

If we call the same `Codegen::to_string()` method on the roots of the constraints, I don't think we can disambiguate which cases we should be in in all cases.

However, maybe we could:
- keep only the `trace_segment: TraceSegmentId` parameter
- always write constants as `E::from(Felt::new(c)))` in the to_string() impl
- Have a helper function to convert the `E::from(Felt::new(c)))` back to `Felt::new(c))` in boundaries, but it seems even less clean IMO (as it's manipulations of Strings)

What do you think?

bobbinth: Let's leave as is for now (and maybe forever) :)

bobbinth (approved):
All looks good! Thank you!

I will merge this PR and we can continue the follow-up work ing #441 and #442.

PR #439, opened 2025-08-06
Add Constant Propagation pass after MIR passes

### LINKED ISSUE #427: Add Constant Propagation pass after MIR passes
As raised in the comments of https://github.com/0xMiden/air-script/pull/424, the MIR pipeline shows a regression of Constant Propagation in some cases (in particular, for Exp operations where powers are indices in a list comprehension).

We should either re-implement a Constant Propagation pass on the MIR, or, if more suitable, better handle the Exp operations (as the scope of the issue is limited to specific cases).

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/427

Tasks:
- [x] Implement folding of the following cases 
  - [x] BinOp(Const, Const) -> Const (BinOp being Add, Sub, Mul and Exp)
  - [x] Add(x, 0) | Add(0, x) -> x
  - [x] Mul(x, 0) | Mul(0, x) -> 0
  - [x] Mul(x, 1) | Mul(1, x) -> x
  - [x] Exp(0, x) -> 0
  - [x] Exp(x, 0) -> 1 (note: this one is redundant with how we expand Exp node during lowering MIR > AIR, I've only included it here to be exhaustive)
  - [x] Sub(x, 0) -> x
- [x] Adapt lowering to Air (e.g. relax the Sub(x,y) requirement for Enf nodes)
- [x] Clean code


mir/src/passes/mod.rs
```diff
@@ -3,12 +3,12 @@ mod translate;
 mod unrolling;
 mod visitor;
 // Note: ConstantPropagation and ValueNumbering are not implemented yet in the MIR
-//mod constant_propagation;
+mod constant_propagation;
 //mod value_numbering;
```
bobbinth: Comment on line 5 should be updated.

Also, what is the `value_numbering` pass? Do we still need it?

air-script/tests/list_comprehension/list_comprehension_nested.rs
```diff
@@ -82,9 +82,9 @@ impl Air for ListComprehensionAir {
     fn evaluate_transition<E: FieldElement<BaseField = Felt>>(&self, frame: &EvaluationFrame<E>, periodic_values: &[E], result: &mut [E]) {
         let main_current = frame.current();
         let main_next = frame.next();
-        result[0] = E::ZERO + main_current[0] * E::ONE + main_current[1] * E::from(Felt::new(2_u64)) - E::from(Felt::new(3_u64));
-        result[1] = E::ZERO + main_current[0] * E::from(Felt::new(2_u64)) + main_current[1] * E::from(Felt::new(3_u64)) - E::from(Felt::new(5_u64));
-        result[2] = E::ZERO + main_current[0] * E::from(Felt::new(3_u64)) + main_current[1] * E::from(Felt::new(4_u64)) - E::from(Felt::new(7_u64));
+        result[0] = main_current[0] * E::ONE + main_current[1] * E::from(Felt::new(2_u64)) - E::from(Felt::new(3_u64));
```
bobbinth: Question: why doesn't `main_current[0] * E::ONE` doesn't get collapsed into `main_current[0]`?

mir/src/ir/node.rs
```diff
@@ -229,7 +229,8 @@ impl Link<Node> {
                 Root::None(span) => Node::None(*span),
             };
         } else {
-            unreachable!();
+            // If the [Node] is stale, we set it to None
+            to_update = Node::None(self.span());
```
bobbinth, line 233: What does "stale" mean in this context?

air/src/passes/translate_from_mir.rs
```diff
@@ -514,6 +514,16 @@ impl AirBuilder<'_> {
...
+            Op::Boundary(boundary) => {
+                // If we have a boundary, this means the ConstantPropagation pass removed the `Sub`
+                // node.
+                let zero_node = mir::ir::Value::create(SpannedMirValue {
+                    value: MirValue::Constant(ConstantValue::Felt(0)),
+                    span: boundary.span(),
+                });
+                let sub = Sub::create(bc.clone(), zero_node, boundary.span());
+                self.build_boundary_constraint(&sub)
+            },
```
bobbinth: This is done to help the witnerfell codegen, right? If so, is this something we can handle in the codegen itself? That is, could we check there if the root of boundary constraint is a single trace access node, and generate `x - 0` expression there?

mir/src/ir/owner.rs
```diff
@@ -208,7 +208,8 @@ impl Link<Owner> {
                 Root::None(span) => Owner::None(*span),
             };
         } else {
-            unreachable!();
+            // If the [Owner] is stale, we set it to None
+            to_update = Owner::None(self.span());
```
bobbinth, line 212: Similar to the question in the other comment: what does "stale" mean here?

bobbinth (commented):
Looks good! Thank you! I left a few questions/comments inline.

mir/src/passes/mod.rs
Leo-Besancon: Updated the comment. If I understand correctly, Value numbering seems redundant with how we implement CSE by evaluating the nodes directly, and so I've removed the file and the comments in https://github.com/0xMiden/air-script/pull/434.

air-script/tests/list_comprehension/list_comprehension_nested.rs
Leo-Besancon: I've noticed this while working on https://github.com/0xMiden/air-script/pull/444 (that is based on this branch), but I added a fix here in the latest commit. Basically, if the constant was wrapped in an Accessor node, it was not considered as constant and not folded.

air/src/passes/translate_from_mir.rs
Leo-Besancon: Indeed, I've better factorized the code in the latest commit, and reused the better handling of boundary constraints in winterfell of PR https://github.com/0xMiden/air-script/pull/419

mir/src/ir/node.rs
Leo-Besancon: `Node` and `Owner` store Backlinks (Weak references) to the underlying `Op` or `Root`.
	
Here, stale nodes refer to nodes that are no longer referenced in the graph, for which the weak reference have been dropped.
	
That would happen if for instance if we:
- clone a given node
- update the clone (with the `set` method of `Link<Op>`)
- try to get and reuse the original node.

mir/src/passes/mod.rs
```diff
@@ -1,14 +1,14 @@
...
 mod inlining;
 mod translate;
 mod unrolling;
 mod visitor;
-// Note: ConstantPropagation and ValueNumbering are not implemented yet in the MIR
-//mod constant_propagation;
+// Note: ValueNumbering is not implemented yet in the MIR
 //mod value_numbering;
-//pub use constant_propagation::ConstantPropagation;
 //pub use value_numbering::ValueNumbering;
```
bobbinth: I thought based on https://github.com/0xMiden/air-script/pull/439#discussion_r2281788485 we no longer need "value number" - but maybe I misunderstood?

air/src/passes/translate_from_mir.rs
```diff
@@ -604,6 +570,79 @@ impl AirBuilder<'_> {
     fn insert_op(&mut self, op: Operation) -> NodeIndex {
         self.air.constraint_graph_mut().insert_node(op)
     }
+
+    /// Extracts the trace access information from a given [Mir] `Boundary``.
+    fn extract_trace_from_boundary(
+        &self,
+        boundary: MirBoundary,
+    ) -> Result<MirTraceAccess, CompileError> {
```
bobbinth: nit: there is an extra backtick at the end of line 574.

Also, could we expand the doc comment a bit more to explain what this does? My current understanding is that given a boundary constraint expression, this figures out which trace column this boundary constraint references - correct?

air/src/passes/translate_from_mir.rs
```diff
@@ -604,6 +570,79 @@ impl AirBuilder<'_> {
...
+
+        Ok(trace_access)
+    }
+
+    /// Marks a boundary as constrained by the given trace access information.
+    fn mark_constrained_boundary(
+        &mut self,
+        trace_access: MirTraceAccess,
+        boundary: &MirBoundary,
+    ) -> Result<(), CompileError> {
```
bobbinth: Could we expand the comment here as well? My understanding is that the reason we do this is to make sure we don't have duplicate boundary constraints. Or are there other reasons?

air/src/passes/translate_from_mir.rs
bobbinth, line 1: Not for this PR, but I would love to make this file a bit more readable - e.g., by splitting out some large `match` branches into helper functions and generally, maybe adding submodules etc. You've already started doing this in this PR, but would be great to continue.

bobbinth (approved):
Looks good! Thank you! I left a few small comments inline. After these are addressed, we should be good to merge.

One general question: what's the main reason to do constant propagation in MIR rather than in the `AlgebraicGraph`? It seems like doing it in the `AlgebraicGraph` could be a bit more straight-forward?

air/src/passes/translate_from_mir.rs
Leo-Besancon: Yes it started pretty simple but we iteratively added quite a few things in this file. Here is the refactor issue: https://github.com/0xMiden/air-script/issues/445

air/src/passes/translate_from_mir.rs
Leo-Besancon: Added some details here and below, please let me know if it's still unclear!

mir/src/passes/mod.rs
Leo-Besancon: True, it is removed in the other PR but may as well remove it here too!

Leo-Besancon:
> Looks good! Thank you! I left a few small comments inline. After these are addressed, we should be good to merge.
> 
> One general question: what's the main reason to do constant propagation in MIR rather than in the `AlgebraicGraph`? It seems like doing it in the `AlgebraicGraph` could be a bit more straight-forward?

Good point! Initially I just put it there as it was a regression introduced by MIR (expressions previously handled in AST's ConstantPropagation were no longer handled), but it would have been easier in the Air.
But now I think the main reason to keep it here is to ensure that computed indices (https://github.com/0xMiden/air-script/pull/444) are folded before the end of MIR.
This ensures that we do not have to handle vectors in the AIR (or during lowering of MIR to AIR).

bobbinth (approved):
Looks good! Thank you!

> Good point! Initially I just put it there as it was a regression introduced by MIR (expressions previously handled in AST's ConstantPropagation were no longer handled), but it would have been easier in the Air.
> But now I think the main reason to keep it here is to ensure that computed indices (#444) are folded before the end of MIR.
> This ensures that we do not have to handle vectors in the AIR (or during lowering of MIR to AIR).

Makes sense! Let's still create an issue to investigate moving constant folding into AIR, but maybe keeping a simpler pass for index folding here (not sure it would make sense - but we can discuss that in the issue).

Leo-Besancon:
> Looks good! Thank you!
> 
> > Good point! Initially I just put it there as it was a regression introduced by MIR (expressions previously handled in AST's ConstantPropagation were no longer handled), but it would have been easier in the Air.
> > But now I think the main reason to keep it here is to ensure that computed indices (#444) are folded before the end of MIR.
> > This ensures that we do not have to handle vectors in the AIR (or during lowering of MIR to AIR).
> 
> Makes sense! Let's still create an issue to investigate moving constant folding into AIR, but maybe keeping a simpler pass for index folding here (not sure it would make sense - but we can discuss that in the issue).

Perfect, here is the issue: https://github.com/0xMiden/air-script/issues/446

PR #440, opened 2025-08-08
Reorder match optimizer

### LINKED ISSUE #431: [MIR] Unroll list comprehensions before match statements to allow more optimizations
This is a follow-up issue related to https://github.com/0xMiden/air-script/pull/413.

Currently, we try and optimize If nodes during the UnrollingFirstPass. This means that For nodes have been transformed into: `Vector<Parameter(0, for_node), ..., Parameter(n, for_node)>`, but it's only later, during UnrollingSecondPass, that we replace each `Parameter(i, for_node)` parameter with the correct value (the body of the corresponding for_node evaluated at index `i`).

The implication is that constraints that rely on list comprehensions would always have different .eval() values, and such constraints in match statements may not be combined optimally.

The updated unrolling workflow would look something like:
1. We keep If nodes the same during UnrollingFirstPass
2. We unroll the list comprehensions in UnrollingSecondPass
3. We handle If nodes in a third pass, now that we know list comprehensions values.

For implementing this, we'll have to watch out for 
- how list comprehensions and match statements can interact
- If moving the unroll of If nodes later has other consequences. For example, if If node is not a root node, we may need to apply another Unrolling pass afterwards to make sure that no Vector remains in the graph.

_Originally posted by @Leo-Besancon in https://github.com/0xMiden/air-script/pull/413#discussion_r2236682797_

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/431

Task list:
- [x] Implement UnrollingThirdPass to handle `If` nodes
- [x] Add E2E test mixing selectors to combine and list comprehension
- [x] Check the produced graph and ensure the match optimizer correctly handled this case.
- [x] Clean and refactor codeto avoid redundancy between UnrollingFirstPass and UnrollingThirdPass.

Note: This branch is currently based on top of PR https://github.com/0xMiden/air-script/pull/434, it will need to be rebased on next when it is merged.


air-script/tests/selectors/selectors_combine_with_list_comprehensions.air
```diff
@@ -0,0 +1,59 @@
...
+
+public_inputs {
+    stack_inputs: [1],
+}
+
+boundary_constraints {
+    enf c.first = 0;
+}
+
+# Note: we want the following constraints values to be equivalent:
```
adr1anh: I don't fully understand what this comment means at first glance. What are we testing/expecting here?

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,468 @@
...
+
+        Ok(updated_enf)
+    }
+
+    fn visit_boundary_bis(
+        &mut self,
+        _graph: &mut Graph,
+        boundary: Link<Op>,
+    ) -> Result<Option<Link<Op>>, CompileError> {
+        let mut updated_boundary = None;
```
adr1anh: Nit(also applicable to the other visit functions): this variable can be omitted by returning the result in the `if` branch.

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,468 @@
...
+            let initial_value = fold_ref.initial_value.clone();
+
+            let iterator_ref = iterator.borrow();
+            let Op::Vector(iterator_vector) = iterator_ref.deref() else {
+                unreachable!("Expected vector iterator in fold, found: {:?}", iterator_ref);
+            };
+            let iterator_nodes = iterator_vector.children().borrow().deref().clone();
+
+            let mut acc_node = initial_value;
+            for iterator_node in iterator_nodes {
```
adr1anh: Nit: we can replace the loop with `fold`

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,468 @@
...
+
+        Ok(Some(Vector::create(new_vec, if_ref.span())))
+    }
+
+    fn visit_vector_bis(
+        &mut self,
+        _graph: &mut Graph,
+        vector: Link<Op>,
+    ) -> Result<Option<Link<Op>>, CompileError> {
+        let mut updated_vector = None;
```
adr1anh: Return the `Option` from the `if` branch

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,468 @@
...
+    ) -> Result<Option<Link<Op>>, CompileError> {
+        let mut updated_enf = None;
+
+        {
+            let enf_ref = enf.as_enf().unwrap();
+            let expr = enf_ref.expr.clone();
+            if let Op::Vector(vec) = expr.borrow().deref() {
+                let ops = vec.children().borrow().deref().clone();
+                let mut new_vec = vec![];
+                for op in ops.iter() {
```
adr1anh: Nit: use `Vec::collect`

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,468 @@
...
+// each visit_*_bis function returns an Option<Link<Op>> instead of Result<(), CompileError>,
+// to mutate the nodes (e.g. modifying a Operation<Vectors> to Vector<Operations>)
+impl UnrollingThirdPass<'_> {
+    fn visit_value_bis(
+        &mut self,
+        _graph: &mut Graph,
+        value: Link<Op>,
+    ) -> Result<Option<Link<Op>>, CompileError> {
+        // safe to unwrap because we just dispatched on it
+        let mut updated_value = None;
```
adr1anh: Nit: return result from `match`

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,468 @@
...
+    Matrix::create(res_m, span)
+}
+
+fn unroll_binary_op(
+    lhs: Link<Op>,
+    rhs: Link<Op>,
+    parent: Link<Op>,
+    span: SourceSpan,
+) -> Result<Option<Link<Op>>, CompileError> {
+    let mut updated_binary_op = None;
```
adr1anh: Return result from `if` branches

adr1anh (commented):
Mostly left some nits, but otherwise looks good.

Leo-Besancon:
> Mostly left some nits, but otherwise looks good.

Thank you, I've refactored the visit functions and improved the test comment.

Leo-Besancon:
Edit: fixed several issues in the latest commits.
- The E2E test was too simple, and the list comprehensions were folded into constant vectors during AST's constant propagation, so it did not effectively test the reorder.
- The lowering of Let statements into MIR was incorrect
- The handling of ListComprehensions in AST's ConstantPropagation also had a bug
- I removed the handling of If nodes in the First pass of Unrolling

With a more suitable test, it seems to fail for now, I'll investigate and fix it!

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,441 @@
...
+                    segment: trace_access_binding.segment,
+                    column: trace_access_binding.offset + index,
+                    row_offset: 0,
+                }),
+            });
+            vec.push(val);
+        }
+        Vector::create(vec, span)
+    }
+}
```
bobbinth: This function seems to be identical to the function with the same name in the `unrolling_first_past.rs` file. Is there a reason not to have it just in the `src/passes/unrolling/mod.rs` file and then used in both the first and the third pass files?

This applies to some other functions in this file as well.

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,441 @@
...
+        _graph: &mut Graph,
+        add: Link<Op>,
+    ) -> Result<Option<Link<Op>>, CompileError> {
+        // safe to unwrap because we just dispatched on it
+        let add_ref = add.as_add().unwrap();
+        let lhs = add_ref.lhs.clone();
+        let rhs = add_ref.rhs.clone();
+
+        unroll_binary_op(lhs, rhs, add.clone(), add_ref.span())
+    }
```
bobbinth: Why does this method need to take `self` or `_graph` as parameters? Seems like they are not really needed here, and then this could be just a pure function. This pure function can then be shared between this and first passes.

This applies to some other methods in this file too.

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,441 @@
...
+    Matrix::create(res_m, span)
+}
+
+fn unroll_binary_op(
+    lhs: Link<Op>,
+    rhs: Link<Op>,
+    parent: Link<Op>,
+    span: SourceSpan,
+) -> Result<Option<Link<Op>>, CompileError> {
```
bobbinth: Could we add doc comments explaining what is supposed to happen in this function? My understanding is that this is specifically for unrolling binary operations where operands are vectors, right? If so, how can this be expressed in AirScript (I may be misremembering something, but I didn't think we could apply binary operations to vectors)?

Also, if either `lhs` or `rhs` is not a vector, the function would just return `None` - but shouldn't it be an error? Or is the situation where this function is called with `lhs` being a vector and `rhs` not being a vector is impossible? (if so, would be good to explain the the doc comments why this is the case).

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,441 @@
...
+        }
+        Ok(None)
+    }
+
+    fn visit_matrix_bis(
+        &mut self,
+        _graph: &mut Graph,
+        _matrix: Link<Op>,
+    ) -> Result<Option<Link<Op>>, CompileError> {
+        Ok(None) // Matrix are already unrolled, we have nothing to do
```
bobbinth: If matrixes are already unrolled, should this be `unreachable`? And do we need this function at all?

bobbinth (commented):
Looks good! Thank you. I reviewed all code except for actually checking if the codegen of the new test is correct and left some comments inline. The comments are about reducing code duplication and also simplifying the code and improving its readability.

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,441 @@
...
+            });
+
+        Ok(Some(resulting_node))
+    }
+
+    /// Visiting an `If` node consists of evaluating all the main trace constraints contained in
+    /// the match arms, and combining them to optimize the resulting vector of constraints if
+    /// possible. We handle bus related constraints separately, as they cannot be combined with
+    /// main trace constraints.
+    fn visit_if_bis(
```
Al-Kindi-0: Could we add a few words about what this method returns ?

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,441 @@
...
+        &mut self,
+        _graph: &mut Graph,
+        vector: Link<Op>,
+    ) -> Result<Option<Link<Op>>, CompileError> {
+        // safe to unwrap because we just dispatched on it
+        let vector_ref = vector.as_vector().unwrap();
+        let children = vector_ref.elements.borrow().deref().clone();
+        let size = vector_ref.size;
+
+        if size == 1 {
```
Al-Kindi-0: A comment here would be helpful

Al-Kindi-0 (commented):
Did an initial pass and it looks good.
Would be good to get some more clarity on the 'unroll_binary_op' function in relation to the point raised by @bobbinth

mir/src/passes/unrolling/unrolling_third_pass.rs
Leo-Besancon: In most cases, matrices are constants and already accessed / unrolled during the AST's constant propagation. However, there are some cases (e.g. https://github.com/0xMiden/air-script/blob/next/air/src/tests/variables.rs#L131, but also later when we'll have computed indices on matrices) where we have Matrix nodes in the graph at that point.

The unrolling pass:
- Keeps scalar as is (or compute them, for instance with `Fold` nodes)
- Converts Vector-shaped operations (that are not yet Vector) into Vector nodes. This applies to e.g. `If`, `For`, or currently an `Add` on two vectors.
- Converts Matrix-shaped operations (that are not yet Matrix) into Matrix nodes. This applies to e.g. Values that are constant matrices.

So what I meant by the comment is that `Matrix` nodes are already in their primitive forms. Do you see how we can clarify this?

In aa7ca120, I've removed the function and just return Ok(None) where it is called for simplicity.

mir/src/passes/unrolling/unrolling_third_pass.rs
Leo-Besancon: This was to replicate the signature for `visit_*(...)` methods in visitor.rs, but I agree it's not needed. I've removed these two arguments, marked them as public in unrolling_first_pass, and imported them in unrolling_third_pass in e1098d04.

I've also removed the unnecessary `to_link_and()` helper function as it can now be achieved with `map_or`.

mir/src/passes/unrolling/unrolling_third_pass.rs
Leo-Besancon: This is no longer needed now that I reuse the same `visit_*_bis` functions from `unrolling_first_pass.rs` in `unrolling_third_pass.rs` below.

mir/src/passes/unrolling/unrolling_third_pass.rs
Leo-Besancon: Thank you for pointing these out!

You are correct, we currently cannot directly add two vectors in the frontend (instead of `v + u`, we would do something like `[v + u for (v, u) in (v, u)]`).
During the design phase for the MIR pipeline, I think we talked about the possibility of supporting in the MIR such operations, but I don't remember if it was clear whever we wanted it or not. It was a long time ago and maybe we'd rather simplify this and remove this possibility.

Note that when testing, I've also encountered one case where we rely on this: https://github.com/0xMiden/air-script/blob/next/mir/src/tests/list_comprehension.rs#L31
`enf expected = result;` with two vectors will enforce that the vector `Sub(expected - result)` equals to [0, 0, 0, 0].

Our choices are:
- Either we remove this support, make sure we catch this during semantic analysis, and rewrite the test as:
`enf expected = result for (expected, result) in (expected, result);` I'm leaning towards this one)
- Or we allow them and document them accordingly (note that `Mul` and `Exp` between two vectors are also handled element by element in the MIR)

Leo-Besancon: > Also, if either lhs or rhs is not a vector, the function would just return None - but shouldn't it be an error? Or is the situation where this function is called with lhs being a vector and rhs not being a vector is impossible? (if so, would be good to explain the the doc comments why this is the case).

The `Ok(None)` case is to handle the case where we have both lhs and rhs scalars: `Add(1, 2)` is already unrolled so we have nothing to do.
		
If one of them is a vector and not the other, I agree it should be unreachable and it's probably a good idea to catch it there.

Although:
- It depends on whever we keep these functions or not, I'll wait for you input on this
- It may be better handled with direct typing checks (depending on the Typing PR)

mir/src/passes/unrolling/match_optimizer.rs
```diff
@@ -120,12 +120,8 @@ impl<'a> MatchOptimizer<'a> {
                 Some(index) => {
                     // If the eval is already in the list, we use its index in the
                     // node_evals vec
-                    let Some(constraints_vec) = self.constraints_evaluation_indices.get_mut(&index)
-                    else {
-                        unreachable!(
-                            "Error: Reference to an evaluation index not found in constraints_evaluation_indices"
-                        );
-                    };
+                    let constraints_vec = self.constraints_evaluation_indices.get_mut(&index).expect("Error: An existing evaluation index was not found in constraints_evaluation_indices");
```
bobbinth: nit: formatting here looks a bit off.

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,148 @@
...
+            work_stack: vec![],
+            random_inputs: RandomInputs::default(),
+        }
+    }
+}
+
+// For the first pass of Unrolling, we use a tweaked version of the Visitor trait,
+// each visit_*_bis function returns an Option<Link<Op>> instead of Result<(), CompileError>,
+// to mutate the nodes (e.g. modifying a Operation<Vectors> to Vector<Operations>)
+impl UnrollingThirdPass<'_> {
```
bobbinth: The comment refers to the "first pass" - is this correct?

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,148 @@
...
+
+impl Visitor for UnrollingThirdPass<'_> {
+    fn work_stack(&mut self) -> &mut Vec<Link<Node>> {
+        &mut self.work_stack
+    }
+
+    // We visit all boundary constraints and all integrity constraints
+    // No need to visit the functions or evaluators, as they should have been inlined before this
+    // pass
+    fn root_nodes_to_visit(&self, graph: &Graph) -> Vec<Link<Node>> {
```
bobbinth, line 91: Should we add some `assert!` statements here to make sure that functions and evaluators have already been inlined? Or is this already handled in the `visit_node()` below?

mir/src/passes/unrolling/unrolling_first_pass.rs
```diff
@@ -660,21 +386,155 @@ impl Visitor for UnrollingFirstPass<'_> {
...
-    f: F,
-) -> Result<Option<Link<Op>>, CompileError>
-where
-    F: FnOnce(&mut Graph, Link<Op>) -> Result<Option<Link<Op>>, CompileError>,
-{
-    if let Some(op) = back.to_link() {
-        f(graph, op)
-    } else {
-        Ok(None)
+pub fn visit_enf_bis(enf: Link<Op>) -> Result<Option<Link<Op>>, CompileError> {
```
bobbinth: Since this (and other similar functions) are used both here and in the third pass, it probably makes sense to move them to the parent module (i.e., to `src/passes/unrolling/mod.rs`.

mir/src/passes/unrolling/unrolling_third_pass.rs
```diff
@@ -0,0 +1,148 @@
...
+            | Node::Evaluator(_)
+            | Node::Call(_)
+            | Node::For(_)
+            | Node::Parameter(_) => {
+                unreachable!(
+                    "Unexpected node during Unrolling: Function, Evaluators, Calls, For nodes and Parameters should have been inlined before this pass. Found: {:?}",
+                    node
+                );
+            },
+        };
```
bobbinth: A few nits here (could be addresses in a follow-up):

- I would change the type of `updated_op` to just `Option<Link<Op>>`. This does mean adding a `?` operators at the end of most match arms, but I think it looks a bit cleaner.
- What is the difference here between arms that return `None` vs. the ones that are unreachable? For example, why is `Node::Matrix` gets converted to `None`, even though the comment says that matrixes are already unrolled. From what I can tell, making `Node::Matrix` into `unreachable` doesn't cause any issues (or at least all the tests are passing).
- Instead of spelling out everything that is unreachable, should we just use a wildcard match?

Summarizing the above, this could look something like:
```rust
let updated_op: Option<Link<Op>> = match node.borrow().deref() {
    Node::Enf(e) => e.to_link().map_or(Ok(None), visit_enf_bis)?,
    Node::Boundary(b) => b.to_link().map_or(Ok(None), visit_boundary_bis)?,
    Node::Add(a) => a.to_link().map_or(Ok(None), visit_add_bis)?,
    Node::Sub(s) => s.to_link().map_or(Ok(None), visit_sub_bis)?,
    Node::Mul(m) => m.to_link().map_or(Ok(None), visit_mul_bis)?,
    Node::Exp(e) => e.to_link().map_or(Ok(None), visit_exp_bis)?,
    Node::Fold(f) => f.to_link().map_or(Ok(None), visit_fold_bis)?,
    Node::Vector(v) => v.to_link().map_or(Ok(None), visit_vector_bis)?,
    Node::Value(v) => v.to_link().map_or(Ok(None), visit_value_bis)?,
    Node::If(i) => i.to_link().map_or(Ok(None), |el| self.visit_if_bis(el))?,
    Node::BusOp(_b) => None,
    Node::Accessor(_a) => None,
    _ => {
        unreachable!(
            "Unexpected node during Unrolling: Function, Evaluators, Calls, For nodes and Parameters should have been inlined before this pass. Found: {:?}",
            node
        );
    },
};
```

And all the tests are still passing. Though, it is not really clear to me why `Node::BusOp` and `Node::Accessor` as unreachable causes errors.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - but the ones that require non-trivial amount of work, we can address in a follow-up.

mir/src/passes/unrolling/unrolling_third_pass.rs
bobbinth: > we remove this support, make sure we catch this during semantic analysis, and rewrite the test as:
>   `enf expected = result for (expected, result) in (expected, result);` I'm leaning towards this one)

Yes, let's do this - though, I'd do this in a follow-up PR.

mir/src/passes/unrolling/unrolling_third_pass.rs
Leo-Besancon: I've fixed your first and third points in the latest commit.
For your second point:
- `Ok(None)` would mean that we can have this node type in the graph, and we do nothing if we find them (they are already in the state we want them in).
- `unreachable` would mean that we do not expect to encounter this node type in the graph.

The comment `matrixes are already unrolled` refers to the fact that if we encounter them, then we should do nothing (similarly we do nothing with vectors of size greater than 1).

For instance, if we replace the `Ok(None)` by an `unreachable` for matrices in the `UnrollingFirstPass`, this following test fails:
https://github.com/0xMiden/air-script/blob/reorder_match_optimizer_upstream/air/src/tests/variables.rs#L46

It does not fail during the third pass because the second pass already handles them, but I think we could still encounter them.

mir/src/passes/unrolling/unrolling_third_pass.rs
Leo-Besancon: Yes this is already done in the visit_node() function.

Instead of relying on panics inside a pass whenever its assumptions are broken, maybe we could have a way to check them earlier. Here, it could look like:
- a function `Inlining::assert_pass_results()` that ensures all functions and evaluators have been inlined,
- or a function `Unrolling::assert_pre_pass_assumptions()` that does the same.

I'm not sure if it's really useful (we don't have too much to check for now I think), but I could make an issue for this if you want.

mir/src/passes/unrolling/unrolling_third_pass.rs
Leo-Besancon: I've updated the comment to refer more to what the 3rd pass does.

mir/src/passes/unrolling/unrolling_third_pass.rs
Leo-Besancon: Ok, here is the corresponding issue: https://github.com/0xMiden/air-script/issues/458

bobbinth (approved):
Looks good! Thank you! We can probably refactor things a bit more, but let's leave it for follow-up PRs.

PR #443, opened 2025-08-11
CI: upgrade rust-toolchain installs in CI

### LINKED ISSUE #394: mirror CI from `miden-vm`
`air-script` and [miden-vm](https://github.com/0xMiden/miden-vm) have different `cargo fmt` and `cargo clippy` as part of their CI. Also, air-script currently doesn't use a makefile either.

To help with developer experience and cohesion, it'd be great to use the same workflow here as is used in `miden-vm`

Original message by @bobbinth: https://github.com/0xMiden/air-script/pull/376#discussion_r2071122538

Task list:
- [ ] add a makefile with the same parameters for `cargo fmt/clippy` as in `miden-vm` 
- [ ] update CI to use the makefile

huitseeker:
Migrates to https://github.com/dtolnay/rust-toolchain instead of deprecated actions-rs.

~~Contributes to~~ Closes #394

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bugreport.yml
+++ b/.github/ISSUE_TEMPLATE/1-bugreport.yml
@@ -1,0 +1,38 @@
+name: "Bug Report"
+description: "File a bug report"
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for taking the time to fill out this bug report!
+  - type: textarea
+    id: version
+    attributes:
+      label: "Packages versions"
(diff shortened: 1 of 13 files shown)


bobbinth (commented):
Looks good! Thank you!

I think one other place to cross-reference is [.github](https://github.com/0xMiden/.github/tree/main/.github). I know @mmagician recently put a bunch of reference workflows there (not all of them relevant here).

One thing that I think would be great to add is the changelog check. In `miden-vm` we run it as a part of CI and show an error unless a PR has `no changelog` label.

bobbinth (approved):
Looks good! Thank you!

PR #444, opened 2025-08-14
Allow support for computed indices

### LINKED ISSUE #245: Add support for computed indexes
Add support for computed indexes, i.e. `s[i + 1]` which can be used in constraint comprehension.

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/245

Task list:
- [x] Allow computed indices in AST
  - [x] Replace e.g. `AccessType::Index(usize)` by `AccessType::Index(Box<ScalarExpr>)` in parser
  - [x] Adapt AST's SemanticAnalysis (many thing we cannot resolve early)
  - [x] Adapt AST's ConstantPropagation 
- [x] Allow computed indices in MIR
  - [x] Update Accessor types to reference MIR nodes in AccessType, and update lowering from AST
  - [x] Update MIR's Unrolling pass
  - [x] Update MIR's ConstantPropagation pass  
- [x] Add E2E tests, containing for example this list comprehension:
```
let y = [j * vec[j + 1] for j in 0..4];
```
- [x] Add parser tests
- [x] Add MIR tests 
- [x] Potentially more complex E2E tests
- [x] Check that there are no regression in semantic analysis assumptions
- [ ] We should integrate this PR with the new typing system, to ensure that indices are integers and not felts

Note: This PR contains a fix for Mir's Inlining that has been raised by the `computed_indices_complex.rs` E2E test. Because of this fix, we need to call the Inlining multiple times (until there are no more `Call` nodes in the graph). Maybe there is a better way to handle this (for instance, keeping track when we duplicate a Call node and add it to the `working_stack`), but I haven't thought too much about it for now.

Question: @adr1anh @bobbinth Should we also handle computed slice bounds? For now, slices are assumed constant (either an integer or a reference to a constant symbol) and are resolved early. I still handle them in the MIR, but the are constant by that point, so I should refactor the code to either ignore slices in the MIR, or handle them fully as computed expressions.
It may look something like this (though it may not be the best example to showcase usage..):
```
    let x = [ sum([2^i for i in j..j+3]) for j in 0..3 ]
    # x = [ sum(1, 2, 4), sum(2, 4, 8), sum(4, 8, 16) ]
    # x = [ 7, 14, 28]
```


air-script/tests/computed_indices/computed_indices_simple.air
```diff
@@ -0,0 +1,36 @@
+def ComputedIndicesAir
+
+const VEC = [0, 1, 2, 3, 4];
```
adr1anh: This isn't used?

air/src/passes/translate_from_mir.rs
```diff
@@ -126,7 +129,11 @@ struct AirBuilder<'a> {
 /// so we need to ensure these cases are properly indexed.
 fn indexed_accessor(mir_node: &Link<Op>) -> Link<Op> {
     if let Some(accessor) = mir_node.as_accessor() {
-        if let AccessType::Index(index) = accessor.access_type {
+        if let MirAccessType::Index(index) = accessor.access_type.clone() {
+            let index = match get_inner_const(&index) {
```
adr1anh: Just unwrap?

parser/src/sema/semantic_analysis.rs
```diff
@@ -184,6 +185,7 @@ impl VisitMut<SemanticAnalysisError> for SemanticAnalysis<'_> {
                                 offset: binding.offset,
                                 size: binding.size,
                                 ty: binding.ty,
+                                access: binding.access.clone(),
```
adr1anh: no need to clone

parser/src/sema/semantic_analysis.rs
```diff
@@ -1152,54 +1163,59 @@ impl SemanticAnalysis<'_> {
...
+                                    .with_secondary_label(arg.span(), format!("callee expects {} trace columns here, but this binding provides {}", param.size, tr_size))
+                                    .emit();
                         }
                     },
                     Ok(BindingType::Vector(ref elems)) => {
                         let mut size = 0;
                         for elem in elems.iter() {
                             match elem {
                                 BindingType::TraceColumn(tr) | BindingType::TraceParam(tr) => {
+                                    let tr_size = match tr.ty() {
```
adr1anh: can we make this snippet a method of `Type`?

parser/src/ast/trace.rs
```diff
@@ -248,48 +250,72 @@ impl TraceBinding {
...
+                return Err(InvalidAccessError::IndexIntoScalar);
             },
-            AccessType::Matrix(..) => Err(InvalidAccessError::IndexIntoScalar),
+            (Some(AccessType::Matrix(..)), AccessType::Matrix(..)) => {
+                return Err(InvalidAccessError::SliceOfMatrix);
+            },
+            (Some(AccessType::Matrix(..)), AccessType::Slice(..)) => {
+                return Err(InvalidAccessError::SliceOfMatrix);
+            },
+            _ => todo!(),
```
adr1anh: What other patterns are we expecting?

mir/src/passes/translate.rs
```diff
@@ -1259,31 +1267,53 @@ impl<'a> MirBuilder<'a> {
...
+                        Some(
+                            Value::builder()
+                                .value(SpannedMirValue {
+                                    span: access.span(),
+                                    value: MirValue::TraceAccessBinding(tab),
+                                })
+                                .build(),
+                        )
+                    },
+                    //FIXME? Convert to MirAccessType and add Accessor wrapper?
```
adr1anh: Do we need an issue for this ?

mir/src/passes/translate.rs
```diff
@@ -1294,8 +1324,18 @@ impl<'a> MirBuilder<'a> {
 
         if segment.name == id {
             // We access $main[i]
-            if let AccessType::Index(column) = access.access_type {
-                Some(TraceAccess::new(TraceSegmentId::Main, column, access.offset))
+            // FIXME: Check is_constant
```
adr1anh: Issue?

mir/src/passes/translate.rs
```diff
@@ -1307,20 +1347,43 @@ impl<'a> MirBuilder<'a> {
...
+                        let mir_binding_access = self.translate_access_type(binding_access)?;
+                        let accessor =
+                            Accessor::create(value, mir_binding_access, 0, access.span());
+                        Ok(Some(accessor))
+                    } else {
+                        Ok(Some(value))
+                    }
+                },
+                AccessType::Index(_extra_offset) if binding.size > 1 => {
+                    // FIXME?
```
adr1anh: Highlighting the `FIXME`

mir/src/passes/mod.rs
```diff
@@ -382,10 +393,20 @@ pub fn duplicate_node_or_replace(
...
         Op::Accessor(accessor) => {
             let indexable = accessor.indexable.clone();
-            let access_type = accessor.access_type.clone();
+            let new_access_type = match accessor.access_type.clone() {
+                MirAccessType::Default => MirAccessType::Default,
+                MirAccessType::Index(index) => MirAccessType::Index(
+                    current_replace_map.get(&index.get_ptr()).unwrap().1.clone(),
+                ),
+                MirAccessType::Matrix(row, col) => MirAccessType::Matrix(
+                    current_replace_map.get(&row.get_ptr()).unwrap().1.clone(),
```
adr1anh: Why not use `current_replace_map[&row.get_ptr()]`

mir/src/passes/inlining.rs
```diff
@@ -29,6 +29,59 @@ impl<'a> Inlining<'a> {
...
+            first_pass.func_eval_nodes_where_called.clone(),
+        );
+        Visitor::run(&mut second_pass, ir.constraint_graph_mut())?;
+
+        Ok(second_pass.had_calls)
+    }
+}
+
+// If we have to run the inlining algorithm `INLINING_LIMIT`, we return an error
+const INLINING_LIMIT: usize = 10;
```
adr1anh, line 61: Where does this come from?

mir/src/passes/constant_propagation.rs
```diff
@@ -121,6 +120,185 @@ impl ConstantPropagation<'_> {
...
+                                )
+                                .with_primary_label(index.span(), "index is not constant")
+                                .emit();
+                            return Err(CompileError::Failed);
+                        },
+                    };
+                    // Check that the child node is a vector, raise diag otherwise
+                    // Replace the current node by the index-th element of the vector
+                    // Raise diag if index is out of bounds
+                    if let Op::Vector(indexable_vector) = indexable.borrow().deref() {
```
adr1anh: My editor complains that this 67 line fragment is duplicated mir/src/passes/unrolling.rs:708

mir/src/passes/constant_propagation.rs
```diff
@@ -121,6 +120,185 @@ impl ConstantPropagation<'_> {
...
+                                    value: MirValue::TraceAccess(TraceAccess {
+                                        segment: trace_access.segment,
+                                        column: trace_access.column + index_usize,
+                                        row_offset: trace_access.row_offset,
+                                    }),
+                                });
+                                updated_accessor = Some(new_node);
+                            },
+                            _ => {
+                                unreachable!("indexable is {:?}", indexable); // raise diag
```
adr1anh: Can we do this in this PR?

mir/src/passes/constant_propagation.rs
```diff
@@ -121,6 +120,185 @@ impl ConstantPropagation<'_> {
...
+                                )
+                                .emit();
+                            return Err(CompileError::Failed);
+                        },
+                    };
+                    // Check that the child node is a matrix, raise diag otherwise
+                    // Replace the current node by the index-th element of the vector
+                    // Raise diag if index is out of bounds
+
+                    if let Op::Vector(indexable_vector) = indexable.borrow().deref() {
```
adr1anh: Also duplicate

adr1anh (commented):
Not a full review as a lot of the language theory goes over my head. Left some code comments in line though

parser/src/sema/semantic_analysis.rs
Leo-Besancon: I've made it a method of `TraceBinding`

parser/src/ast/trace.rs
Leo-Besancon: I've removed the todo!() and handled all cases

mir/src/passes/translate.rs
Leo-Besancon: Actually we no longer need this as we keep the assumption that slices are constant! I've removed the comment.

mir/src/passes/translate.rs
Leo-Besancon: This is a small improvement, I've handled it in b27ec359964043fc9a9452886e6cbdf0b66aeccb

mir/src/passes/translate.rs
Leo-Besancon: Same as above, handled in b27ec359964043fc9a9452886e6cbdf0b66aeccb

mir/src/passes/inlining.rs
Leo-Besancon: I've explained this in the description of the PR above:

> Note: This PR contains a fix for Mir's Inlining that has been raised by the computed_indices_complex.rs E2E test. Because of this fix, we need to call the Inlining multiple times (until there are no more Call nodes in the graph). Maybe there is a better way to handle this (for instance, keeping track when we duplicate a Call node and add it to the working_stack), but I haven't thought too much about it for now.

It was an issue that were never catched by the existing tests.

If you prefer, I can create a separate PR for this instead!

mir/src/passes/constant_propagation.rs
Leo-Besancon: I've cleaned up the error handling in this function (also deduplicated it):
- Emit diagnostics when we have out-of-bound indices with Matrix access
- But keeping the unreachable!() there as it would be a direct computer bug and not a user-facing issue. I've improved the error message.

Leo-Besancon:
Hello @adr1anh, thank you for the review!

I think I've addressed all your comments for now.

I've also created this issue for one of the FIXME that should probably not be handled in this PR: https://github.com/0xMiden/air-script/issues/452

Al-Kindi-0:
@Leo-Besancon , is this ready? Or would you like an extra round of reviews?
It would be great to get this merged asap as it will simplify some of the work during the migration of VM constraints to the updated syntax.

Leo-Besancon:
> @Leo-Besancon , is this ready? Or would you like an extra round of reviews? It would be great to get this merged asap as it will simplify some of the work during the migration of VM constraints to the updated syntax.

Hi! This PR is ready on my side, but it needs additional reviews, especially as it touches various parts of the code (parser, AST and MIR).
The best order to go forward would be:
- Review and merge this PR into `next`
- Merge `next` into https://github.com/0xMiden/air-script/pull/423
- Add typing checks to computed indices in https://github.com/0xMiden/air-script/pull/423
- Review and merge https://github.com/0xMiden/air-script/pull/423 into `next`

Then, the v0.5 Milestone would be done and the syntax isn't likely to change much (aside from the multi trace issue).

mir/src/passes/mod.rs
```diff
@@ -450,3 +473,262 @@ pub fn duplicate_node_or_replace(
...
+        Op::Value(Value {
+            value:
+                SpannedMirValue {
+                    value: MirValue::Constant(ConstantValue::Felt(c)),
+                    ..
+                },
+            ..
+        }) => Some(*c),
+        _ => None,
+    }
```
bobbinth, line 482: I think it would be cleaner to add a method on `Value` for this - something like `Value::get_inner_constant()`. Then here it would be:
```rust
match value.borrow().deref() {
    Op::Value(value) => value.get_inner_constant(),
    _ => None,
}
```

We can also then re-use this code in the other `get_inner_const()` function (in constant propagation). In fact, I wonder if it may make sense to combine these two methods.

air-script/tests/computed_indices/computed_indices_complex.air
```diff
@@ -0,0 +1,52 @@
...
+
+    # vec_1[2] = 7
+    let z = vec_1[y];
+
+    enf a[1] = double(z);
+}
+
+fn apply_mds(state: felt[2]) -> felt[3] {
+    return [sum([s * m for (s, m) in (state, mds_row)]) for mds_row in MDS];
+}
```
bobbinth, line 61: This is the part that was not possible before, right?

air-script/tests/computed_indices/computed_indices_complex.air
```diff
@@ -0,0 +1,52 @@
...
+    # 2 * 2 - 4
+    let x = double(2) - vec_2[1];
+
+    # vec_2[0] = 2
+    let y = vec_2[x];
+
+    # vec_1[2] = 7
+    let z = vec_1[y];
+
+    enf a[1] = double(z);
```
bobbinth: I would maybe simplify this test and at the same time make it a bit more robust. Specifically:

- Let's use trace columns as the `state`.
- Maybe we can apply MDS once and then in Rust code we'd see the actual formula of how it is getting applied (I think).

mir/src/passes/mod.rs
```diff
@@ -450,3 +473,262 @@ pub fn duplicate_node_or_replace(
...
+}
+
+/// Handle the visit of an accessor node, used for both Unrolling and ConstantPropagation passes
+/// The `compute_indices` bool indicates whether we the indices need to be known constant at this
+/// stage.
+pub fn handle_accessor_visit(
+    accessor: Link<Op>,
+    compute_indices: bool,
```
bobbinth: nit (feel free to ignore): I would maybe invert `compute_indices` into something like `expect_constant_indices`. May make the meaning a bit more clear (at least that's how we describe it in the doc comments).

parser/src/ast/trace.rs
```diff
@@ -230,6 +230,8 @@ pub struct TraceBinding {
     pub size: usize,
     /// The effective type of this binding
     pub ty: Type,
+    /// The access type associated to this TraceBinding
+    pub access: Option<AccessType>,
```
bobbinth: What this filed mean here? That is, in what situations would a trace binding have `access` vs. when it could be `None`?

parser/src/ast/trace.rs
```diff
@@ -248,48 +250,77 @@ impl TraceBinding {
...
 
     /// Returns a [Type] that describes what type of value this binding represents
     #[inline]
     pub fn ty(&self) -> Type {
-        self.ty
+        match self.access.clone() {
+            Some(AccessType::Default) => self.ty,
+            Some(AccessType::Slice(range_expr)) => Type::Vector(range_expr.to_slice_range().len()),
+            Some(AccessType::Index(_)) => Type::Felt,
+            Some(AccessType::Matrix(..)) => Type::Felt,
```
bobbinth: Under what circumstances could `TraceBinding` have type `Matrix`? Should this be `unreachable!`?

bobbinth (commented):
Looks good! Thank you! This is a very light review from me - I didn't really review the parser changes, and even the MIR part, I reviewed sporadically. I did leave a few questions/comments inline though.

mir/src/passes/inlining.rs
adr1anh: Ah I see! I think my question was about why we chose 10 as a constant, though intuitively I imagine this is a good estimate.

Would it make sense to make this a variable parameter? Maybe through an environment variable?

air-script/tests/computed_indices/computed_indices_complex.air
adr1anh: This may not be the right place, but I imagine this is actually the _only_ place where we'd ever use matrices. For the rare cases where we want to evaluate a matrix (only hash functions I think), we could just manually implement it. If so it might be worth removing matrices entirely. WDYT?

mir/src/passes/mod.rs
```diff
@@ -450,3 +473,262 @@ pub fn duplicate_node_or_replace(
...
+                    ..
+                },
+            ..
+        }) => Some(*c),
+        _ => None,
+    }
+}
+
+/// Handle the visit of an accessor node, used for both Unrolling and ConstantPropagation passes
+/// The `compute_indices` bool indicates whether we the indices need to be known constant at this
```
Al-Kindi-0: this needs rewording

Al-Kindi-0 (approved):
LGTM (given my understanding of this part of the codebase)

mir/src/passes/mod.rs
Soulthym: > In fact, I wonder if it may make sense to combine these two methods.
I tried it but it seemed to break the added test (https://github.com/0xMiden/air-script/pull/444/files/88bb70ce0b281ab3593e9675ba7e07e6215f969b#diff-0ac8abbe17a8a361edfcb0a3a06c7108d5c316885eebc406d0669a3b5754a821)

For now Value's impl is reused in both constant propagation and translation from ast to Mir  in 6e734dc4

air-script/tests/computed_indices/computed_indices_complex.air
Soulthym: > This is the part that was not possible before, right?

Yes when coupled with a function. I think it was ok without a function before because it was computed before semantic analysis.

> If so it might be worth removing matrices entirely. WDYT?

I agree, its already in the todo list for the frontend refactor. Ill gather those in a separate issue.

mir/src/passes/mod.rs
Soulthym: Renamed in 90b5bdc1 and everywhere else in c8adbf58

parser/src/ast/trace.rs
Soulthym: Im not sure why it is this way or if it could be without the Option, so I'm only interpreting here based on the code.

It seems it allows setting it after creation, potentially handling the difference between an unset access_type and AccessType::Default, distinguishing between an error (a mismatch) and a missing access_type.

Noting so we can revisit with @Leo-Besancon based on his feedback

parser/src/ast/trace.rs
Soulthym: Made it unreachable in abbf5253

mir/src/passes/mod.rs
Soulthym: Reworded in c8adbf5, let me know if it needs more adjustments.

air-script/tests/computed_indices/computed_indices_complex.air
Soulthym: I've adapted the air-script of the test in 82cd993 but it does not compile as is, I'm investigating why that is.
It seems we extract the constant from the wrong value, extracting the `vec_1 instead of 1` or report the error incorrectly here: https://github.com/0xMiden/air-script/blob/82cd9931172ae6dec9a9f8d1077d93a671e51fec/air-script/tests/computed_indices/computed_indices_complex.air#L39

air-script/tests/computed_indices/computed_indices_complex.air
Leo-Besancon: > This is the part that was not possible before, right?

Not exactly, the apply_mds function was already tested in https://github.com/0xMiden/air-script/blob/allow-computed-indices-upstream/air-script/tests/list_comprehension/list_comprehension_nested.air
In the list_comprehension_nested test, it is indeed applied to the state as you suggested below, so we can check the rust code.

What is new is to be able to index into the `vec_1` and `vec_2` list comprehensions with indices whose can only be computed late into the compilation process (for instance, because they rely on list comprehensions)

In this test (computed_indices_complex), I needed a vector that would not be folded as a constant during AST's constant propagation, otherwise I would not be able to test the computed indices targetting list comprehensions properly.

I'll add some comments to make what this test does clearer.

air-script/tests/computed_indices/computed_indices_complex.air
Leo-Besancon: > I would maybe simplify this test and at the same time make it a bit more robust. Specifically:
> 
>     * Let's use trace columns as the `state`.
> 
>     * Maybe we can apply MDS once and then in Rust code we'd see the actual formula of how it is getting applied (I think).

In the latest commit, I've made it so that one MDS is kept on a constant vector to be able to produce constant indices, and use it to index into the second MDS that does rely on the state.
I've also added comments on what is expected / happening in the test, please tell me if it's better!

parser/src/ast/trace.rs
Leo-Besancon: > What this filed mean here? That is, in what situations would a trace binding have `access` vs. when it could be `None`?

Indeed the Option is unnecessary as the None case can just be handled the same as the Default case. I've removed it in the latest commit

Al-Kindi-0:
@Leo-Besancon , is this good to go?

Leo-Besancon:
> @Leo-Besancon , is this good to go?

Hi! Yes, if you guys are happy with our latest replies, it should be good to merge, I think we've addressed everything!

Al-Kindi-0:
Thank you!

PR #447, opened 2025-08-22
Refactor trace segment handling to remove `TraceSegmentId::index()` and introduce `TraceShape<T>`

### LINKED ISSUE #442: Remove `TraceSegmentId::index()` method and improve references to trace segments in the codebase
In https://github.com/0xMiden/air-script/pull/438, we've introduced an enum to avoid referencing a trace segment by a number:

```rust
pub enum TraceSegmentId {
   Main = 0,
   Aux = 1,
}
```

In different parts of the codebase, we want to store data indexed by the trace segment, for instance:
- In the AST and in the MIR, we want to describe the trace bindings defined by the user (that can only be the main trace at this stage)
- In the AIR, we want to describe boudary and integrity constraints for the main and the aux trace
- In the ACE codegen, the layout contains the main and the aux traces, but also the quotient that is considered the third segment. We use the method `TraceSegmentId::index()` to convert the TraceSegmentId back to an usize and index by it, which is not optimal.

Handling `Vec<TraceSegment>` or `BTreeMap<TraceSegmentId, TraceSegment>` is suboptimal as we need to either use `TraceSegmentId::index()`, or ensure that the BTreeMap does contain all the information we expect at each stage of the compilation.

An improvement and a way to unify the different parts of the codebase is described by @adr1anh in: https://github.com/0xMiden/air-script/pull/438#issuecomment-3173659705
```rust
struct TraceShape<T> {
  main: T,
  aux: T,
}
```

Having for example `FrontEndTraceShape<T>`, `TraceShape<T>`, `FullTraceShape<T>` should ensure that the needed data is always there, and remove the need of the `index` method.

_Originally posted by @bobbinth in https://github.com/0xMiden/air-script/pull/438#pullrequestreview-3104259358_

surajk-m:
Closes #442 

This PR addresses the issue raised in #438 by removing the `TraceSegmentId::index()` method and introducing a type safe approach to handling trace segments across the codebase. The goal is to eliminate the reliance on numeric indices for referencing trace segments (main, aux, and quotient) and unify the representation of trace data using the proposed TraceShape<T> and FullTraceShape<T> structs, as suggested by @adr1anh and @bobbinth.

@adr1anh @bobbinth , I would appreciate your feedback on whether the `TraceShape<T>` and `FullTraceShape<T>` implementations fully cover all use cases in the AIR and ACE modules.

### DIFF
--- a/air/src/ir/mod.rs
+++ b/air/src/ir/mod.rs
@@ -24,6 +24,94 @@ pub use self::{
     value::{PeriodicColumnAccess, PublicInputAccess, Value},
 };
 
+/// A fixed two segment trace shape containing values for the main and aux segments.
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+pub struct TraceShape<T> {
+    main: T,
+    aux: T,
+}
+
+impl<T> TraceShape<T> {
+    pub fn new(main: T, aux: T) -> Self {
+        Self { main, aux }
+    }
+
+    #[inline]
+    pub fn get(&self, id: TraceSegmentId) -> &T {
+        match id {
+            TraceSegmentId::Main => &self.main,
+            TraceSegmentId::Aux => &self.aux,
+        }
+    }
+
+    #[inline]
+    pub fn get_mut(&mut self, id: TraceSegmentId) -> &mut T {
+        match id {
+            TraceSegmentId::Main => &mut self.main,
+            TraceSegmentId::Aux => &mut self.aux,
+        }
+    }
+
+    pub fn map<U, F: FnMut(&T) -> U>(&self, mut f: F) -> TraceShape<U> {
+        TraceShape { main: f(&self.main), aux: f(&self.aux) }
+    }
+}
+
+impl<T> core::ops::Index<TraceSegmentId> for TraceShape<T> {
+    type Output = T;
+    fn index(&self, index: TraceSegmentId) -> &Self::Output {
+        self.get(index)
+    }
+}
+
+impl<T> core::ops::IndexMut<TraceSegmentId> for TraceShape<T> {
+    fn index_mut(&mut self, index: TraceSegmentId) -> &mut Self::Output {
+        self.get_mut(index)
+    }
+}
+
+/// A fixed three segment trace shape containing values for the main, aux, and quotient segments.
+///
+/// This wraps a two segment `TraceShape<T>` for the witness traces, and adds a separate
+/// `quotient` segment which is not addressable via `TraceSegmentId`.
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+pub struct FullTraceShape<T> {
+    pub segments: TraceShape<T>,
+    pub quotient: T,
+}
+
+impl<T> FullTraceShape<T> {
+    pub fn new(main: T, aux: T, quotient: T) -> Self {
+        Self {
+            segments: TraceShape::new(main, aux),
+            quotient,
+        }
+    }
+
+    #[inline]
+    pub fn segments(&self) -> &TraceShape<T> {
+        &self.segments
+    }
+
+    #[inline]
+    pub fn segments_mut(&mut self) -> &mut TraceShape<T> {
+        &mut self.segments
+    }
+}
+
+impl<T> core::ops::Index<TraceSegmentId> for FullTraceShape<T> {
+    type Output = T;
+    fn index(&self, index: TraceSegmentId) -> &Self::Output {
+        &self.segments[index]
+    }
+}
+
+impl<T> core::ops::IndexMut<TraceSegmentId> for FullTraceShape<T> {
+    fn index_mut(&mut self, index: TraceSegmentId) -> &mut Self::Output {
+        &mut self.segments[index]
+    }
+}
+
 /// The offset of the "current" row during constraint evaluation.
 pub const CURRENT_ROW: usize = 0;
 /// The minimum cycle length of a periodic column
--- a/air/src/ir/mod.rs
+++ b/air/src/ir/mod.rs
@@ -41,15 +41,9 @@ impl<T> TraceShape<T> {
     }
 
     /// Returns an iterator over mutable references to `(TraceSegmentId, T)` in segment order.
-    pub fn iter_mut(
-        &mut self,
-    ) -> impl Iterator<Item = (TraceSegmentId, &mut T)> {
+    pub fn iter_mut(&mut self) -> impl Iterator<Item = (TraceSegmentId, &mut T)> {
         let (main, aux) = (&mut self.main, &mut self.aux);
-        [
-            (TraceSegmentId::Main, main),
-            (TraceSegmentId::Aux, aux),
-        ]
-        .into_iter()
+        [(TraceSegmentId::Main, main), (TraceSegmentId::Aux, aux)].into_iter()
     }
 }
 
--- a/air/src/ir/mod.rs
+++ b/air/src/ir/mod.rs
@@ -43,7 +43,7 @@ impl<T> TraceShape<T> {
     /// Returns an iterator over mutable references to `(TraceSegmentId, T)` in segment order.
     pub fn iter_mut(
         &mut self,
-    ) -> core::array::IntoIter<(TraceSegmentId, &mut T), 2> {
+    ) -> impl Iterator<Item = (TraceSegmentId, &mut T)> {
         let (main, aux) = (&mut self.main, &mut self.aux);
         [
             (TraceSegmentId::Main, main),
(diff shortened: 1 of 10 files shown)


air/src/ir/mod.rs
```diff
@@ -22,6 +22,94 @@ pub use self::{
     value::{PeriodicColumnAccess, PublicInputAccess, Value},
 };
 
+/// A fixed two segment trace shape containing values for the main and aux segments.
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+pub struct TraceShape<T> {
+    main: T,
+    aux: T,
```
adr1anh: These could be `pub`, avoiding the need for the `get` and `get_mut` methods. The `Index` and `IndexMut` implementation would be used for indexing with a `TraceSegmentID`.

air/src/passes/translate_from_mir.rs
```diff
@@ -634,12 +642,11 @@ impl AirBuilder<'_> {
...
-            .trace_columns
-            .get_mut(&trace_access.segment)
-            .expect("Boundary constraint on an unknown trace segment")
-            .mark_constrained(boundary.span(), trace_access.column, boundary.kind)
-        {
+        if let Some(prev) = self.trace_columns.get_mut(trace_access.segment).mark_constrained(
+            boundary.span(),
+            trace_access.column,
+            boundary.kind,
+        ) {
```
adr1anh: Adding the following trait implementations could simplify this call. Up to you if you think it is worth it.

```
impl<T> core::ops::Index<usize> for TraceShape<T> {
    type Output = &T;

    fn index(&self, index: usize) -> &Self::Output {
        match index {
            0 => &self.main,
            1 => &self.aux,
            _ => panic!("invalid segment index"),
    }
```

adr1anh (commented):
Thanks for tackling this, it looks really good. 

I think we could go a bit further here with the following suggestions. I might be missing some details that make these points redundant, so up to you whether they are worth implementing. 
- Adding `impl Index{Mut}<usize> for {Full}TraceShape` would simplify some calls that still rely on indices. Panicking when encountering an out-of-range index would be okay as that matches indexing behavior of other types.
- Replace `trace_segments: [[InputRegion; 3]; 2]` in the ACE Layout with `[FullTraceShape; 2]`
- We can remove `get` and `get_mut` as these are handled by the `Index` trait.
- Make the fields  of `TraceShape` public.
- I think keeping `TraceSegmentId::index` could still be useful, though it could be replaced by `From<TraceSegmentId> for usize` since the conversion never fails.

bobbinth:
@surajk-m - are you still planning to finish this PR?

air/src/passes/translate_from_mir.rs
surajk-m: The proposed `Index<usize>` trait implementation for `TraceShape<T>` looks promising and could indeed simplify indexing calls. I'll integrate it with a matching `IndexMut<usize>` for mutable access.

surajk-m:
Hey @bobbinth @adr1anh,

Thanks for the feedback! Sorry for the delayed reply. Yes, I’m still planning to finish this up soon. I’ve been tied up with other tasks, but I’m aiming to push the changes by tomorrow.

impl `Index{Mut}<usize> for {Full}TraceShape` will clean up the index based calls nicely. I’ll add impl `Index<usize>` and `IndexMut<usize>` for both `TraceShape` and `FullTraceShape`, with panics on out of range indices to match standard behavior. The example impl you provided for Index is spot on, I’ll use something similar and extend it for mutability.

I’ll keep index() for now since it’s explicit and useful in some contexts, but I’ll also add the From `impl` as a convenience (it’s infallible, so no issues there).

Regarding `trace_segments: [[InputRegion; 3]; 2]` in ACE Layout, switching to `[FullTraceShape; 2]` makes total sense for consistency and better type safety. I’ll make that change.

adr1anh:
This looks great! One final comment: how easy is it to use this in the `AceInputs`?

bobbinth:
@surajk-m - seems like the build is failing here? would you be able to fix this?

surajk-m:
Hey @bitwalker @bobbinth, thanks for flagging the build issue,

> Super glad you like the changes! On the `AceInputs` no worries at all. Since we kept the `AirInputs` interface as is (still exposing `main`, `aux`, and `quotient` separately), all the updates to `AceVars.segments` and the indexing logic are nicely tucked away. We don’t need to tweak anything in existing `AceInputs` code. 

> The switch to `[FullTraceShape<Vec<QuadFelt>>; 2]` makes segment handling smoother, and `into_ace_vars` now uses `FullTraceShape::new(main_*, aux_*, quotient_*)` for more consistent row building. 

> The memory write loop’s also cleaner with the new `Index/IndexMut<usize>` setup, using `layout.trace_segments[row][i]` and 
`self.segments[row][i]`. It all plays nicely with `AceInputs` out of the box.

air/src/ir/mod.rs
```diff
@@ -24,6 +24,134 @@ pub use self::{
...
+impl<T> core::ops::IndexMut<TraceSegmentId> for TraceShape<T> {
+    fn index_mut(&mut self, index: TraceSegmentId) -> &mut Self::Output {
+        match index {
+            TraceSegmentId::Main => &mut self.main,
+            TraceSegmentId::Aux => &mut self.aux,
+        }
+    }
+}
+
+impl<T> core::ops::Index<usize> for TraceShape<T> {
```
Leo-Besancon, line 69: It seems that these two implementations are never used in the codebase (it's only used for `FullTraceShape` I believe)

```rust
impl<T> core::ops::Index<usize> for TraceShape<T>
impl<T> core::ops::IndexMut<usize> for TraceShape<T>
```

I think we could remove them as we have `Index<TraceSegmentId>` and `IndexMut<TraceSegmentId>`, but we could also keep them to stay consistent with `FullTraceShape` (that does rely on them).

What do you think @adr1anh?

air/src/passes/translate_from_mir.rs
```diff
@@ -65,19 +71,21 @@ impl Pass for MirToAir<'_> {
...
-                    Symbol::new(TraceSegmentId::Aux.index() as u32),
-                ),
+                Identifier::new(SourceSpan::default(), Symbol::intern("$aux")),
                 bus_raw_bindings,
             );
             for binding in aux_trace_segment.bindings.iter() {
                 bus_bindings_map.insert(binding.name.unwrap(), binding.offset);
             }
-            trace_columns.insert(aux_trace_segment.id, aux_trace_segment);
+            trace_columns_aux = aux_trace_segment;
```
Leo-Besancon: If there are buses, you currently create two TraceSegment for the aux trace (one empty to initialize it, that is replaced with one with the bus bindings).

I think it would be cleaner to only allocate it once:
```rust
let trace_columns_aux = if buses.empty() {
  // return empty aux TraceSegment
} else {
  // return TraceSegment with bus_raw_bindings  
}
```

codegen/ace/src/tests/quotient.rs
```diff
@@ -66,7 +66,10 @@ pub fn eval_quotient(air: &Air, ace_vars: &AceVars, log_trace_len: u32) -> QuadF
             Operation::Value(v) => match v {
                 Value::Constant(c) => QuadFelt::from(Felt::new(c)),
                 Value::TraceAccess(access) => {
-                    let segment_index = access.segment.index();
+                    let segment_index = match access.segment {
+                        TraceSegmentId::Main => 0,
+                        TraceSegmentId::Aux => 1,
+                    };
                     ace_vars.segments[access.row_offset][segment_index][access.column]
```
Leo-Besancon: You can use directly 
```rust
ace_vars.segments[access.row_offset][access.segment][access.column]
```

Leo-Besancon (commented):
Good job, it's a lot cleaner with this PR!

I've left a few (non-blocking on my side) comments, but otherwise it looks good to me.

adr1anh:
I think it would still be good to use `FullTraceShape` in [`AirInputs`](https://github.com/surajk-m/air-script/blob/57e00abab91be07b047e34f5faad6d5d243236a4/codegen/ace/src/inputs.rs#L24) to replace `main, aux, quotient`.

air/src/ir/mod.rs
adr1anh: I'd favor keeping them for consistency :)

air/src/passes/translate_from_mir.rs
surajk-m: Got it, I’ve updated it so the aux trace is allocated only once.

surajk-m:
Hey @adr1anh, `AirInputs` now uses `FullTraceShape`. Thanks for the feedback above @adr1anh @Leo-Besancon @bobbinth, we can merge the PR.

Leo-Besancon (approved):
LGTM!

bobbinth:
@surajk-m - could you please merge the latest `next` into this branch? The test should pass after this.

adr1anh (approved):

bobbinth (approved):
Looks good! Thank you!

PR #449, opened 2025-08-25
Fix regressions on MIR and list_comprehensions

Leo-Besancon:
I have noticed two issues when implementing https://github.com/0xMiden/air-script/pull/440. I've fixed them on the branch of the PR, but it is cleaner to do a specific PR for these regressions.

1. When lowering the bodies of functions or evaluators, the lowering of Let statements in MIR did not correctly insert their body into the root bodies.
For instance:
```
ev are_squares_in_order([d[3]]) {
    let base = [i for i in 0..3];
    let squares = square_all(base);
    enf d[0] = squares[0];
    enf d[1] = squares[1];
    enf d[2] = squares[2];
}
```

Only the `enf d[2] = squares[2];` constraint was taken into account, the two other enfs were ignored.

2. During ConstantPropagation of list_comprehensions, some early returns caused issues with bindings and lexical scopes.
For instance:
```
    let alpha = [double(i) for i in 0..3];
    let beta = [double(i) for i in 0..4];
```
Here, when visiting the beta list comprehension, the binding `i` was considered still in scope (and with a value of zero), so beta was reduced to `let beta = [0, 0, 0, 0];` instead of its correct value.

I've improved the E2E tests to test these fixes.


mir/src/passes/translate.rs
```diff
@@ -392,8 +390,16 @@ impl<'a> MirBuilder<'a> {
...
+            let new_stmt = self.translate_statement(stmt)?;
+            if i < let_stmt.body.len() - 1 {
+                match self.root.borrow().deref() {
+                    Root::Function(f) => f.body.borrow_mut().push(new_stmt.clone()),
+                    Root::Evaluator(e) => e.body.borrow_mut().push(new_stmt.clone()),
+                    Root::None(_span) => {},
+                }
+            }
+            ret_value = new_stmt;
         }
```
bobbinth: Could we add some comments explaining what this does? In general, would be great to add more comments to this module (especially doc comments for each function explaining what it is supposed to do) - but that's probably for another PR.

Things that are not clear to me:

- What is `self.root` here? I'm assuming it represents the context of innermost function/evaluator that we are processing, but it would be great to explain this next to filed definition.
- What does it mean when `self.root` is `Root::None`?
- Why are we treating the last statement differently? i.e., it seems like all other statements are added to the body of w/e the root represents (except for the `Root::None` case, where they seem to be swallowed?), but the last statement is not added to the body and instead returned. A doc comment explaining what this function is doing and what it is supposed to return would help here.

mir/src/passes/translate.rs
```diff
@@ -351,7 +351,6 @@ impl<'a> MirBuilder<'a> {
         body: &'a Vec<ast::Statement>,
     ) -> Result<Link<Root>, CompileError> {
         self.root = func.clone();
-        self.bindings.enter();
         let func = func;
         for stmt in body {
             let op = self.translate_statement(stmt)?;
```
bobbinth: Could we add doc comments (and maybe inline comments) to this function explaining what it is supposed to do?

Things that are not clear to me:
- What is `func` here?
- Why do we take it as a parameter and also return it? From what I can tell, the returned value is never used at callsites.
- Why do we need `let func = func;` statement? I'm not sure it does anything, but maybe I'm missing something?
- Why do we set `self.root` to `func` after every iteration of the loop?

air-script/tests/evaluators/test_air.rs
```diff
@@ -24,8 +24,8 @@ impl AirTester for EvaluatorsAirTester {
                 state[1] = start;
                 state[3] = start;
                 state[4] = start;
-                state[5] = start;
-                state[6] = start;
+                state[5] = Felt::new(1);
+                state[6] = Felt::new(4);
```
bobbinth: Should we note set `state[4]` to `Felt::ZERO` here?

Also, why are we skipping `state[2]` here?

bobbinth (commented):
Looks good! Thank you for catching these! I left some comments inline mostly because the code (and the effect of the changes) wasn't clear to me.

mir/src/passes/translate.rs
Leo-Besancon: Hi! I've added some comments in commit 19a68bc5da07b1caf242d4ed04428ab28c94b137.

I agree the code is a bit confusing, because it was written to only rely on a single node, which is not the case for Let's, so we need to workaround this initial design.
An alternative would be to add all statements (including the last one), and to return an Op::None so that is is not added twice.

But I think the cleanest way would be to make the following changes to avoid any weird logic, what to you think?

```rust
/// Translates a statement into a vector of operations. We return a vector as let statements contain all the statements in its scope.
fn translate_statement(&mut self, stmt: &'a ast::Statement) -> Result<Vec<Link<Op>>, CompileError> {
    match stmt {
        ast::Statement::Let(let_stmt) => self.translate_let(let_stmt),
        ast::Statement::Expr(expr) => Ok(vec![self.translate_expr(expr)?],
		[...]
    }
}

/// Translates a let statement, binding its value to its name for the scope of its body, and returning all the inner statements of its body.
fn translate_let(&mut self, let_stmt: &'a ast::Let) -> Result<Vec<Link<Op>>, CompileError> {
	[...]
	
	// Translating all the statements of the let's body, and adding them to our returned vectors, flattened
    let ret_value = let_stmt.body.iter()
        .map(|stmt| self.translate_statement(stmt))
        .collect::<Result<Vec<_>, _>>()?
        .into_iter()
        .flatten()
        .collect();
	Ok(ret_value)
}
```

Then, when we translate a statement, we handle all the inner statements like we do now. So if we are translating a fucntion / evaluator, we can directly insert all the statements into its body

air-script/tests/evaluators/test_air.rs
Leo-Besancon: Fixed both issues in 41c4e561416041422504c6b01fac53ebb8fbdbe3, thank you

mir/src/passes/translate.rs
Leo-Besancon: I've refactored this part in 42332b4d, most of it was not needed. I still think it would be cleaner to have `translate_statement` returning a `Vec<Link<Op>>` (as proposed above) to remove the need to use `self.root` to rely on setting the statements inside Let statements

bobbinth (approved):
Looks good! Thank you!

mir/src/passes/translate.rs
bobbinth: Yes, I think this would be cleaner - but we can do this as a follow-up. Let's create an issue for this.

Leo-Besancon: Here is the issue: https://github.com/0xMiden/air-script/issues/454

PR #453, opened 2025-09-03
fix: remove_integrity_constraints_root to target integrity list

radik878:
- Correct removal to operate on integrity_constraints_roots instead of boundary_constraints_roots.
- Preserves existing behavior elsewhere; no other files changed.


adr1anh (approved):
Thanks for catching this!

PR #460, opened 2025-09-17
[MIR] Remove binary operation unrolling

### LINKED ISSUE #458: [MIR] Remove binary operation unrolling
In the MIR, we implemented unrolling on binary operations. For instance, a node `Add(Vec(1, 2), Vec(3, 4))` would be unrolled into `Vec(Add(1, 3), Add(2, 4))`.

As it is currently not supported by the language, and for simplicity, we want to remove this feature.

Note that when testing, I've encountered one case where we rely on this: https://github.com/0xMiden/air-script/blob/next/mir/src/tests/list_comprehension.rs#L31
`enf expected = result;` with two vectors will enforce that the vector `Sub(expected - result)` equals to [0, 0, 0, 0].

So we need to:
- [x] Remove this support
- [x] Make sure we catch this during semantic analysis
- [x] Rewrite the test as `enf expected = result for (expected, result) in (expected, result);`

_Originally posted by @Leo-Besancon and @bobbinth in https://github.com/0xMiden/air-script/pull/440#discussion_r2329940091_

Leo-Besancon:
This PR closes https://github.com/0xMiden/air-script/issues/458.

It consists of:
- checking the types of binary expressions in sema
- doing nothing if we encounter `Add`, `Sub`, `Mul`, `Exp`, `Boundary` nodes in unrolling 1st and 3rd pass
- updating the relevant tests

I also needed to fix a behaviour that I'm not 100% sure why emerged (the need to do a little more work during the lowering into AIR by applying `add_row_offset_if_trace_access` to `Accessor`).


bobbinth:
I tried to resolve the merge conflict - but seems like this broke something. Sorry!

Leo-Besancon:
> I tried to resolve the merge conflict - but seems like this broke something. Sorry!

I think it was just the GitHub runner that went down, a rerun of the CI job solved the issue. Thank you!

bobbinth (approved):
Looks good! Thank you!

air/src/passes/translate_from_mir.rs
```diff
@@ -197,8 +223,9 @@ impl AirBuilder<'_> {
     /// Will panic when encountering an unexpected operation
     /// (i.e. that is not a binary operation, a value, enf node or an accessor)
     fn insert_mir_operation(&mut self, mir_node: &Link<Op>) -> Result<NodeIndex, CompileError> {
-        let mir_node = indexed_accessor(mir_node);
+        let mir_node = accessor_to_scalar(mir_node);
         let mir_node = vec_to_scalar(&mir_node);
+        let mir_node = accessor_to_scalar(&mir_node);
         let mir_node_ref = mir_node.borrow();
```
Al-Kindi-0, line 244: I would add some comments to explain what is happening in this sequence

Al-Kindi-0 (approved):
LGTM

adr1anh (approved):

air/src/passes/translate_from_mir.rs
bobbinth: @Soulthym - would you be able to address this comment? Or we can merge as is and then @Leo-Besancon could do it in a separate PR (if needed).

Leo-Besancon: Hi @Al-Kindi-0, I've added comments in the lastest commit, please let me know if what is happening is still not very clear!

PR #461, opened 2025-09-17
Add plonky3 backend

### LINKED ISSUE #450: Implement Plonky3 `AirBuilder` main trace constraints code generation
In light of upcoming support for Plonk3, we should start implementing support for generating constraints using the [`AirBuilder`](https://github.com/Plonky3/Plonky3/blob/main/air/src/air.rs) interface. There are several examples in the plonky3 repository on how to implement it, such as [`keccak`](https://github.com/Plonky3/Plonky3/blob/main/keccak-air/src/air.rs).

We should focus on outputting the main trace constraints before handling bus constraints in the auxiliary trace, though these should be easy to support once the main ones are handled.

Leo-Besancon:
## Describe your changes

This PR aims to close https://github.com/0xMiden/air-script/issues/450.

It became the tracking branch for Plonky3 Codegen related features.

Tasklist:
- [x] Setup main codegen pipeline
- [x] Add E2E test pipeline (for binary test)
- [x] Test the generated code (by proving / verifying the program with a provided trace)
- [x] Handle all AirScript features relative to main trace
  - [x] Main Boundary constraints 
  - [x] Main Transition constraints 
  - [x] Public inputs (fixed-lengths)
  - [x] Periodic columns
    - [x] Traits `AirBuilderWithPeriodicColumns` and `BaseAirWithPeriodicColumns`
    - [x] implement it on a wrapper struct `DebugConstraintBuilderWithPeriodicColumns`
    - [x] implement helper `check_constraints_with_periodic_columns` to check the constraints are valid for a given trace
    - [ ] ~implement the traits for upstream structs needed for `prove` and `verify`~ moved to a followup: https://github.com/0xMiden/air-script/pull/523
- [x] Add all other E2E tests (note: for E2E relying on buses, I generated the main constraints)
- [x] Aux trace handling


codegen/plonky3/src/air/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+fn add_air_trait(scope: &mut Scope, ir: &Air, name: &str) {
+    // add the implementation block for the Air trait.
+    let air_impl = scope
+        .new_impl(name)
+        .generic("AB: AirBuilderWithPublicValues")
+        .impl_trait("Air<AB>");
+
+    let eval_func = air_impl.new_fn("eval").arg_ref_self().arg("builder", "&mut AB");
+    eval_func.line("let main = builder.main();");
+    eval_func.line("let public_values = builder.public_values().to_vec();");
```
adr1anh: I think we can avoid the vector allocation here.

codegen/plonky3/src/air/transition_constraints.rs
```diff
@@ -0,0 +1,18 @@
+use air_ir::{Air, TraceSegmentId};
+use codegen::Function;
+
+use super::Codegen;
+
+/// Adds the main transition constraints to the generated code.
+pub(super) fn add_main_transition_constraints(eval_func: &mut Function, ir: &Air) {
+    // add the main integrity constraints
+    for constraint in ir.integrity_constraints(TraceSegmentId::Main) {
```
adr1anh: I wonder if we could support both transition and integrity constraints (those which are active in all rows, including the last).

adr1anh (commented):
Had a quick sneak peek, but this is great already! Left two small comments about the codegen.

codegen/plonky3/src/air/transition_constraints.rs
Leo-Besancon: Oh I see, indeed I think it's possible to call:

- `builder.when_transition().assert_zero::<_>(...)` if the constraint has a dependancy on the next row,
- `builder.assert_zero::<_>(...)` otherwise

codegen/plonky3/src/air/mod.rs
Leo-Besancon: As `public_values()` borrows `builder` and the constraints need a mutable borrow (and to reference the public_values) on it I'm not sure it's possible?

adr1anh: Could we use the fact that we know the length of the public inputs and use it as a constant array size?

air-script/tests/helpers/plonky3_periodic_columns.rs
```diff
@@ -0,0 +1,146 @@
...
+pub trait BaseAirWithPeriodicColumns<F>: BaseAir<F> {
+    fn get_periodic_columns(&self) -> Vec<Vec<F>> {
+        vec![]
+    }
+}
+
+pub trait AirBuilderWithPeriodicColumns: AirBuilder {
+    type PeriodicColumnsVar: Field + Into<Self::Expr>;
+
+    fn periodic_columns(&self) -> Vec<Self::PeriodicColumnsVar> {
```
adr1anh: We should also return a slice here which we can cast to an array.

adr1anh: I think this is more complicated issue though, since I see below that we're using this to make testing easier.

air-script/tests/helpers/plonky3_periodic_columns.rs
```diff
@@ -0,0 +1,146 @@
...
+    periodic_columns: Vec<Vec<F>>,
+}
+
+impl<'a, F> AirBuilderWithPeriodicColumns for DebugConstraintBuilderWithPeriodicColumns<'a, F>
+where
+    F: Field + Into<Self::Expr>,
+{
+    type PeriodicColumnsVar = F;
+
+    fn periodic_columns(&self) -> Vec<Self::PeriodicColumnsVar> {
```
adr1anh: While maybe not for this PR, I think we can represent `periodic_columns` in the `DebugConstraintBuilderWithPeriodicColumns` struct a bit differently
- Pad all periodic columns to the maximum length by periodic repetition
- Transpose the columns so that we have a vector of rows.
- We just return `&self.periodic_columns[self.row_index % col.len()]`

air-script/tests/constants/constants_plonky3.rs
```diff
@@ -0,0 +1,52 @@
...
+impl<AB: AirBuilderWithPublicValues + AirBuilderWithPeriodicColumns> Air<AB> for ConstantsAir {
+    fn eval(&self, builder: &mut AB) {
+        let main = builder.main();
+        let public_values: [_; NUM_PUBLIC_VALUES] = builder.public_values().try_into().expect("Wrong number of public values");
+        let periodic_values = builder.periodic_columns();
+        let (main_current, main_next) = (
+            main.row_slice(0).unwrap(),
+            main.row_slice(1).unwrap(),
+        );
+        builder.when_first_row().assert_zero::<_>(main_current[0] - AB::Expr::from(AB::F::from_u64(1)));
```
adr1anh: We should be able to call `AB::Expr::from_u64()`.

adr1anh: And as a followup, we should be able to replace this with `AB::Expr::ONE`

air-script/tests/functions/functions_complex_plonky3.rs
```diff
@@ -0,0 +1,44 @@
...
+            main.row_slice(0).unwrap(),
+            main.row_slice(1).unwrap(),
+        );
+        builder.when_first_row().assert_zero::<_>(main_current[3]);
+        builder.when_transition().assert_zero::<_>(main_next[16] - main_current[16] * ((main_current[3] * main_current[3] * main_current[3] * main_current[3] * main_current[3] * main_current[3] * main_current[3] * main_current[1] * main_current[2] + main_current[3] * main_current[3] * (AB::Expr::from(AB::F::from_u64(1)) - main_current[1]) * main_current[2] + main_current[3] * main_current[1] * (AB::Expr::from(AB::F::from_u64(1)) - main_current[2]) + (AB::Expr::from(AB::F::from_u64(1)) - main_current[1]) * (AB::Expr::from(AB::F::from_u64(1)) - main_current[2])) * main_current[0] - main_current[0] + AB::Expr::from(AB::F::from_u64(1))));
+        builder.when_transition().assert_zero::<_>(main_next[3] - (main_current[4] + main_current[5] + main_current[6] + main_current[7] + main_current[8] + main_current[9] + main_current[10] + main_current[11] + main_current[12] + main_current[13] + main_current[14] + main_current[15] + AB::Expr::from(AB::F::from_u64(1))) * AB::Expr::from(AB::F::from_u64(2)));
```
adr1anh: Eventually we'd want to replace ` * AB::Expr::from(AB::F::from_u64(2)))` with `.double()`

adr1anh (approved):
Looks really good! Just left some minor comments which can be addressed in a follow-up PR.

air-script/tests/constants/constants_plonky3.rs
Leo-Besancon: Done in the latest commit with `AB::Expr::ZERO`, `AB::Expr::ONE` and `AB::Expr::from_u64()`

air-script/tests/functions/functions_complex_plonky3.rs
Leo-Besancon: Done in the latest commit, the only change I needed to introduce was to convert e.g. `main_next[16]` (an `AB::Var`) to an `AB::Expr` with `.into()` to have the `double` method available, which seems ok to me.

air-script/tests/helpers/plonky3_periodic_columns.rs
Leo-Besancon: Hi! I've created an issue to do this in a followup: https://github.com/0xMiden/air-script/issues/492

adr1anh (approved):
Looks good to me!

air-script/tests/selectors/test_air_plonky3.rs
```diff
@@ -0,0 +1,61 @@
+use std::marker::PhantomData;
+
+use p3_challenger::{HashChallenger, SerializingChallenger32};
+use p3_circle::CirclePcs;
+use p3_commit::ExtensionMmcs;
+use p3_field::{PrimeField64, extension::BinomialExtensionField};
+use p3_fri::create_benchmark_fri_params;
+use p3_matrix::dense::RowMajorMatrix;
+use p3_merkle_tree::MerkleTreeMmcs;
+use p3_mersenne_31::Mersenne31;
```
Al-Kindi-0: nit (feel free to ignore): we can probably use just Goldilocks to stay consistent throughout

Al-Kindi-0 (approved):
Looks great!
Support and testing for periodic columns and buses will come in a follow up PR once we agree on how our version of P3 backend will look like.

air-script/tests/selectors/test_air_plonky3.rs
Leo-Besancon: Done in the latest commit!

codegen/plonky3/src/air/graph.rs
```diff
@@ -0,0 +1,222 @@
...
+            Value::PeriodicColumn(pc) => {
+                let index =
+                    ir.periodic_columns.iter().position(|(qid, _)| qid == &pc.name).unwrap();
+                match elem_type {
+                    ElemType::Base => format!("AB::Expr::from(periodic_values[{index}].clone())"),
+                    ElemType::Ext => {
+                        format!("AB::ExprEF::from(periodic_values[{index}].clone().into())")
+                    },
+                    ElemType::ExtFieldElem => {
+                        format!("AB::EF::from(periodic_values[{index}].clone())")
```
huitseeker: This branch uses `ElemType::ExtFieldElem` both inside `eval()` and inside the generated bus helpers. In `eval()` the `AB::EF` name exists, but in `buses_initial_values()` and `buses_transitions()` there is no `AB` in scope, so any bus expression that references a periodic column will generate Rust that does not compile. I think this arm needs to emit `EF::from(periodic_values[idx].clone())` instead.

air-script/src/miden_vm_aux_trace_generator.rs
```diff
@@ -0,0 +1,117 @@
...
+    // of `F``
+    let main_transposed = main.transpose();
+    let mut felt_columns_vec = Vec::new();
+    for row in main_transposed.rows() {
+        let col_felt = row.map(|x| Felt::new(x.as_canonical_u64())).collect::<Vec<_>>();
+        felt_columns_vec.push(col_felt);
+    }
+    let col_matrix = ColMatrix::new(felt_columns_vec);
+    let last_program_row = main.height().into();
+    let main_trace = MainTrace::new(col_matrix, last_program_row);
```
huitseeker: `MainTrace::new` expects the last valid program row index, but `main.height()` is the row count, so helpers that later read `main_trace.last_program_row()` can step one past the trace. In the MassaLabs fork, `BlockHashTableRow::table_init()` reads `decoder_hasher_state_first_half(main_trace.last_program_row())`, so this can go out of bounds on padded traces. I think this needs `num_rows - 1` at minimum, or the actual halt row if that is available.

huitseeker (commented):
Thanks for pushing this forward. The Plonky3 backend work looks like the right direction.

One architectural concern before this is something we would want to land on `next`: we should avoid baking MassaLabs forks into the long-term `air-script` dependency story. If  this backend is the forward path, I think we should aim for it to integrate against the official 0xMiden stack rather than rely on forks.

  Concretely, I think there are two follow-ups we should plan for:

  1. Move the Plonky3 integration off fork-specific VM/crypto dependencies and onto the official Miden stack.
  2. Isolate the Plonky3 backend dependency surface from default `air-script` consumers.
     Right now these forked dependencies are unconditional normal dependencies of the `air-script` crate. It would be better to depend on miden-crypto Plonky3 re-exports into a separate crate/module so the core language/compiler stays clean.

  That would give us a cleaner path from “working branch” to “something we can comfortably maintain upstream.”
/cc @Al-Kindi-0

codegen/plonky3/src/air/graph.rs
Leo-Besancon: Indeed current tests don't use this codepath so I did not catch this, addressed in d363738b2a11ce6d98ab0a577f111a25a3afdf45

air-script/src/miden_vm_aux_trace_generator.rs
Leo-Besancon: Nice catch! I've addressed this in a6f53cb853046a4f3346f7b2ed3ed1b962a56709 in the same way I saw it handled in miden-vm

Leo-Besancon:
> Thanks for pushing this forward. The Plonky3 backend work looks like the right direction.
> 
> One architectural concern before this is something we would want to land on `next`: we should avoid baking MassaLabs forks into the long-term `air-script` dependency story. If this backend is the forward path, I think we should aim for it to integrate against the official 0xMiden stack rather than rely on forks.
> 
> Concretely, I think there are two follow-ups we should plan for:
> 
>     1. Move the Plonky3 integration off fork-specific VM/crypto dependencies and onto the official Miden stack.
> 
>     2. Isolate the Plonky3 backend dependency surface from default `air-script` consumers.
>        Right now these forked dependencies are unconditional normal dependencies of the `air-script` crate. It would be better to depend on miden-crypto Plonky3 re-exports into a separate crate/module so the core language/compiler stays clean.
> 
> 
> That would give us a cleaner path from “working branch” to “something we can comfortably maintain upstream.” /cc @Al-Kindi-0

Hi! Indeed, we should be able to improve the dependencies (both avoid forks and depend on released versions instead of git revisions).

Note that currently the dependency to a massalabs maintained miden-vm is only used for aux trace generation, but the current implementation for `miden_vm_aux_trace_generator` could probably be deprecated.
I think multiple followups are possible, depending on how you want to use this work:
- The branch `batch_inverse_aux_trace` improves the generation of traces (but is preliminary work, it would need testing / benchmarking to use)
- Or we could keep the manual miden-vm implementations (that depend on the trace) without relying on this helper, and keeping the boilerplate in the miden-vm side

Issue #463, opened 2025-09-19
Initial structure with minimal example


Al-Kindi-0:
Introduce an initial project structure for writing the Miden VM constraints. This should contain a minimal running example to start from.

PR #464, opened 2025-09-19
Initial structure for Miden VM constraints

Al-Kindi-0:
## Describe your changes
Addresses #377 and #463

This also includes a few changes to the backend in order to fix some issues related to argument counting as well as export of functions from modules.

- Updated `CHANGELOG.md`


constraints/bitwise.air
```diff
@@ -0,0 +1,69 @@
...
+# BITWISE CHIPLET TRANSITION CONSTRAINTS
+##########################################################################################
+#
+# The bitwise chiplet processes 32-bit integers by decomposing them into individual
+# bits, applying bitwise operations, and recomposing the results. This approach
+# enables verification of bitwise logic within the field arithmetic constraint system.
+#
+# PLANNED CONSTRAINT EVALUATORS:
+#
+# ev bitwise_decomposition([bitwise[16]]) {
```
adr1anh, line 38: Should this be 13 columns with two hierarchical selectors? Or are we including the padding?

constraints/chiplets.air
```diff
@@ -0,0 +1,93 @@
...
+    enf s[4]' = s[4] when s[0] & s[1] & s[2] & s[3] & s[4];
+}
+
+##########################################################################################
+# CHIPLET FLAG COMPUTING
+##########################################################################################
+
+# Memory chiplet identification function
+#
+# CONSTRAINT DEGREE: 3 (cubic due to three-factor multiplication)
```
adr1anh: The parenthesis may be a bit verbose

constraints/chiplets.air
```diff
@@ -0,0 +1,93 @@
...
+#     let hash_active = ... ;
+#     let bitwise_active= ... ;
+#     let memory_active = ... ;
+#     let ace_active = ... ;
+#     let ker_rom_active = ... ;
+#     match enf {
+#         case hash_active: hash_chiplet_constraints([chiplets[5..20]]),
+#         case bitwise_active: bitwise_chiplet_constraints([chiplets[5..20]]),
+#         case memory_active: memory_chiplet_constraints([chiplets[5..20]]),
+#         case ace_active: ace_chiplet_constraints([chiplets[5..20]]),
```
adr1anh: We use some pretty funky selectors in the ACE chiplet such as
- `next_is_ace_first`
- `is_ace_transition`
- `is_ace_last` 

We'll tackle this when we get there but I wonder if there's a way where we could either derive them all here and pass them to the individual chiplet procedures. 
Even if they're mostly redundant, the optimization should remove any selector which is not actually used.

constraints/range_checker.air
```diff
@@ -0,0 +1,86 @@
...
+    let value = range_checker[1];        # v: Current 16-bit value [0, 65535]
+    let multiplicity = range_checker[0]; # m: Usage count (how many times v is range-checked)
+
+    ####################################################################################
+    # OPERATION FLAGS - Determine when range checks are needed
+    ####################################################################################
+
+    # U32 range check operation flag
+    # Identifies stack operations that require 16-bit range checking
+    let not_4 = binary_not(decoder[5]);
```
adr1anh, line 57: I'm not sure I get what this variable represents

constraints/range_checker.air
```diff
@@ -0,0 +1,86 @@
...
+    let not_5 = binary_not(decoder[6]);
+    let u32_rc_op = decoder[7] * not_4 * not_5;
```
adr1anh, line 59: Or this one either

constraints/utils.air
```diff
@@ -0,0 +1,129 @@
...
+#
+# CONSTRAINT DEGREE: 2
+#
+# USAGE PATTERN:
+# ```air
+# let flag = column[i];
+# enf is_binary([flag]);  // Ensures flag ∈ {0, 1}
+# ```
+ev is_binary([a]) {
+    enf a^2 = a;
```
adr1anh, line 27: I think this will end up being an instruction, which is good because we can use the optimized `assert_bool` in plonky3

constraints/utils.air
```diff
@@ -0,0 +1,129 @@
...
+# This is a multiplexer function that selects between two values based on condition
+#
+# CONSTRAINT DEGREE: 2
+#
+# USAGE EXAMPLES:
+# ```air
+# let result = select(condition, true_value, false_value);
+# ```
+fn select(condition: felt, if_true: felt, if_false: felt) -> felt {
+    return condition * if_true + (1 - condition) * if_false;
```
adr1anh: Maybe implement as follows to explicitly avoid one multiplication
```suggestion
    return condition * (if_true - if_false) + if_false;
```

mir/src/passes/inlining.rs
```diff
@@ -524,27 +525,50 @@ fn check_evaluator_argument_sizes(
                 };
                 trace_segments_arg_vector_len += size;
             } else if let Some(accessor) = child.as_accessor() {
```
adr1anh, line 527: I'll let others comment on this part as I'm less familiar with it.

constraints/utils.air
```diff
@@ -0,0 +1,129 @@
...
+# Logical XOR: returns 1 if exactly one of a or b is 1, else 0
+#
+# CONSTRAINT DEGREE: 2
+#
+# USAGE EXAMPLES:
+# ```air
+# let result = binary_xor(flag_a, flag_b);
+# ```
+fn binary_xor(a: felt, b: felt) -> felt {
+    return a + b - 2 * a * b;
```
adr1anh: Ideally we would have 
```suggestion
    let ab = a*b;
    return a + b - (ab + ab);
```

adr1anh (commented):
Looks amazing! Left a couple of comments about things that weren't entirely clear on the first pass. There's also comments with potential optimizations that I think would be nice to be performed automatically by the compiler.

constraints/bitwise.air
Al-Kindi-0: Fixed in the follow up PR on Bitwise chiplet

constraints/chiplets.air
Al-Kindi-0: Removed

constraints/chiplets.air
Al-Kindi-0: Would be good to be able to add some sort of virtual columns and then treat them in airscript as we do for the usual (committed) columns. Otherwise, we can pass the selector columns flags and then compute the flags wherever they are need, although this can result in code duplication and dispersion of some of the logic. (See ACE chiplet PR)

constraints/range_checker.air
Al-Kindi-0: This is the negation of the fourth op bit (out of the 7 op bits)

constraints/range_checker.air
Al-Kindi-0: Here, we are computing a flag indicating all ops that require u16 range checks. Should we elaborate more?

constraints/utils.air
Al-Kindi-0: Yeah, that is the intention, namely try to mimic the AirBuilder interface

constraints/utils.air
Al-Kindi-0: Good idea, done

constraints/utils.air
Al-Kindi-0: Though I would probably leave the multiplication by 2 since, depending on the target, we could use a double method which is sometimes faster than a generic addition

constraints/range_checker.air
adr1anh: On second pass it makes sense, up to you.

constraints/utils.air
adr1anh: Ah yes agreed. We should probably optimize multiplications by two like this (cc @Leo-Besancon )

adr1anh (approved):
Looks good to me! Just letting someone more familiar with the inlining code to give a quick look.

adr1anh (approved):
LGTM, once the lint check passes

constraints/tests/miden_vm.rs
```diff
@@ -0,0 +1,172 @@
...
+/// Loads the MidenVM AIR example
+pub fn load_miden_vm_air() -> std::io::Result<String> {
+    let crate_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+    let path = format!("{}/miden_vm.air", crate_dir);
+    let content = std::fs::read_to_string(path)?;
+
+    Ok(content)
+}
+
+#[test]
```
huitseeker, line 34: The hardcoded EXPECTED_CIRCUIT array with 112 values is hard to maintain. Consider using https://insta.rs/ like in https://github.com/0xMiden/miden-vm/pull/2086 for easier updates and better test maintainability.

constraints/utils.air
```diff
@@ -0,0 +1,130 @@
+##########################################################################################
```
huitseeker, line 1: I notice that some modules have STATUS lines while the utils.air file is missing one.

constraints/miden_vm.air
```diff
@@ -0,0 +1,179 @@
...
+}
+
+##########################################################################################
+# PUBLIC INPUTS
+##########################################################################################
+#
+# Public inputs define the interface between the prover and verifier, establishing
+# the computation's inputs and expected outputs that must be verified.
+#
+# INPUT STRUCTURE:
```
huitseeker, line 42: The current mixed approach (implemented selectors + missing chiplet constraints) creates confusion about what functionality actually exists. I'd love more explicit markings for the next developer to beware, or explicitly disabling the chiplets.

huitseeker (approved):
Low confidence on the constraints review — still catching up on context about the VM design. Left come comments inline for the code organization, which overall makes sense to me. Shall we open issues for the missing pieces which we would link to the stacked PRs coming after this one?

constraints/tests/miden_vm.rs
Al-Kindi-0: Removed entirely as this test seizes to make sense when we introduce further constraints than the ones we have in this PR

constraints/utils.air
Al-Kindi-0: My thinking about the STATUS line is that it is in relation to the constraints. The utils module is there to hold helper evaluators and functions and hence a notion of STATUS does not apply, in my opinion.

constraints/miden_vm.air
Al-Kindi-0: I think the evaluator of the chiplets constraints is commented out, or are you thinking of something else?

Al-Kindi-0:
@huitseeker , I am merging this now but let me know if I haven't addressed your comments and I can fix them in a separate PR

PR #466, opened 2025-09-23
Update and include Bitwise chiplet constraints

### LINKED ISSUE #465: Update Bitwise chiplet constraints

Al-Kindi-0:
## Describe your changes

Closes #465 

- Updated `CHANGELOG.md`


constraints/bitwise.air
adr1anh, line 58: This is defined by the selector `s` right?

constraints/bitwise.air
```diff
@@ -66,4 +57,92 @@ periodic_columns {
...
\ No newline at end of file
+##########################################################################################
+
+##########################################################################################
+# HELPERS
+##########################################################################################
+
+### Helper evaluators #############################################################################
+
+# Enforces that the bitwise selector is valid.
```
adr1anh: The description of what the selector does would go here. I feel like `op_flag` might be more descriptive than selector which would dictate whether a row is active or not.

constraints/bitwise.air
```diff
@@ -32,27 +34,16 @@ periodic_columns {
 # The bitwise chiplet processes 32-bit integers by decomposing them into individual
 # bits, applying bitwise operations, and recomposing the results. This approach
 # enables verification of bitwise logic within the field arithmetic constraint system.
```
adr1anh, line 38: I'd mention that each operation spans 8 rows.

constraints/bitwise.air
```diff
@@ -66,4 +57,92 @@ periodic_columns {
...
+
+# Enforces that the bitwise selector is valid.
+#
+# Max constraint degree: 2
+ev bitwise_selector([s]) {
+    # Enforce selector is binary.
+    # Constraint degree: 2
+    enf is_binary([s]);
+
+    # Enforce selector should stay the same throughout the cycle.
```
adr1anh:
```suggestion
    # Enforce operation flag should stay the same throughout the cycle.
```

constraints/bitwise.air
adr1anh: Something like `k_first` and `k_transition` would be more descriptive

constraints/bitwise.air
```diff
@@ -66,4 +57,92 @@ periodic_columns {
...
+    # Enforce that for all rows in an 8-row cycle, except for the last one, the values in a and b
+    # columns are increased by the values contained in the individual bit columns a_limb and 
+    # b_limb.
+    # Constraints degree: 2
+    enf a' = a * 16 + a_aggr when k1;
+    enf b' = b * 16 + b_aggr when k1;
+}
+
+
+# Enforces that the output of the bitwise operation is aggregated correctly from the decomposed
```
adr1anh, line 114: What would a more descriptive name be for `zp` and `z`? Something like `acc_prev` and `acc`?

constraints/bitwise.air
```diff
@@ -66,4 +57,92 @@ periodic_columns {
...
+    # Enforce that for all rows the value in the z column is computed by multiplying the previous
+    # output value (from the zp column in the current row) by 16 and then adding it to the bitwise
+    # operation applied to the row's set of bits of a_limb and b_limb. The entire constraint must 
+    # also be multiplied by the operation selector flag to ensure it is only applied for the 
+    # appropriate operation. The constraint for AND is enforced when s = 0 and the constraint for 
+    # XOR is enforced when s = 1. Because the selectors for the AND and XOR operations are mutually
+    # exclusive, the constraints for different operations can be aggregated into the same result
+    # indices.
+    # Constraints degree: 3
+    let a_and_b = sum([2^i * a * b for (i, a, b) in (0..4, a_limb, b_limb)]);
```
adr1anh: Ideally, we would compute this using a fold or reverse operation here that would be slightly more efficient since we replace the multiplication by an addition
```rust
a_and_b_iter.rev().reduce(|(acc, a_and_b)| acc + acc + a_and_b)
```

constraints/bitwise.air
```diff
@@ -66,4 +57,92 @@ periodic_columns {
...
+        case s: z = zp * 16 + a_xor_b,
+        case !s: z = zp * 16 + a_and_b,
+    };
+}
+
+### Helper functions ##############################################################################
+
+# Returns value aggregated from limbs.
+fn aggregate(limbs: felt[4]) -> felt {
+    return sum([2^i * a for (i, a) in (0..4, limbs)]);
```
adr1anh: Again, an addition chain here might be more efficient. While this description is much cleaner, it might be possible to optimize this in the IR?

constraints/bitwise.air
```diff
@@ -66,4 +57,92 @@ periodic_columns {
...
+}
+
+
+# Enforces that the input to the bitwise chiplet is decomposed into limbs correctly.
+#
+# Max constraint degree: 2
+ev input_decomposition([a, b, a_limb[4], b_limb[4]]) {
+    # Enforce that the input is decomposed into valid bits.
+    # Constraints degree: 2
+    enf is_binary([a]) for a in a_limb;
```
adr1anh, line 90: This will be a good candidate for [`assert_bools`](https://github.com/Plonky3/Plonky3/blob/bc7f8d41b08f3aa3b86092aadd75a2ccb0cf3cef/air/src/air.rs#L148) in Plonky3

constraints/bitwise.air
```diff
@@ -66,4 +57,92 @@ periodic_columns {
...
+    # operation applied to the row's set of bits of a_limb and b_limb. The entire constraint must 
+    # also be multiplied by the operation selector flag to ensure it is only applied for the 
+    # appropriate operation. The constraint for AND is enforced when s = 0 and the constraint for 
+    # XOR is enforced when s = 1. Because the selectors for the AND and XOR operations are mutually
+    # exclusive, the constraints for different operations can be aggregated into the same result
+    # indices.
+    # Constraints degree: 3
+    let a_and_b = sum([2^i * a * b for (i, a, b) in (0..4, a_limb, b_limb)]);
+    let a_xor_b = sum([2^i * (a + b - 2 * a * b) for (i, a, b) in (0..4, a_limb, b_limb)]);
+    enf match {
```
adr1anh: I'd be curious if we can get this to compile down to `z = zp * 16 + s * (a_xor_b - a_and_b)  + a_and_b` which saves one multiplication

constraints/bitwise.air
```diff
@@ -66,4 +57,92 @@ periodic_columns {
...
+        case s: z = zp * 16 + a_xor_b,
+        case !s: z = zp * 16 + a_and_b,
+    };
+}
+
+### Helper functions ##############################################################################
+
+# Returns value aggregated from limbs.
```
adr1anh: Is it worth mentioning that `limbs` is little-endian?

adr1anh (commented):
This all looks really good, and without having studies the constraints in the docs before hand, I'm quite confident about their soundness. 

As a general comment that would apply to other chiplets and traces, I think it would be useful to move away from single letter variables which were practical when formatting as LaTeX but which can now be a lot more descriptive in code. 

If I'm being pedantic, it would be useful to have a quick overview of the chiplet at the top of the file, but on the other hand the docs are already quite descriptive. If we do chose to go for more descriptive names then I think it'll be worth mentioning the mapping.

I also left a few comments around optimizing some constraints. The description here is clear, and this may be hard to actually implement, but we should be able to save some multiplications in some places.

constraints/bitwise.air
adr1anh: I think we can use the `binary_and` and `binary_or` from the parent file

constraints/bitwise.air
Al-Kindi-0: Yes, I have changed its name following your suggestion

constraints/bitwise.air
Al-Kindi-0: done

constraints/bitwise.air
Al-Kindi-0: done

constraints/bitwise.air
Al-Kindi-0: done

constraints/bitwise.air
Al-Kindi-0: agreed, changed

constraints/bitwise.air
Al-Kindi-0: not sure, my thinking was to keep close to the docs but we could change if we find a better name

constraints/bitwise.air
Al-Kindi-0: Yes, I think having a fold and a reverse in the syntax is a good idea. Should we create an issue for this?

constraints/bitwise.air
Al-Kindi-0: I think we can target ease of reading for now and then do it more optimally later, especially if we get the fold + reverse pattern supported

constraints/bitwise.air
Al-Kindi-0: Should be easy to detect it also

constraints/bitwise.air
Al-Kindi-0: Yeah, agreed but I think we can have a convention that enf-match should not be used with two matching arms. and instead what you suggested should be used instead.
I have went on and changed to the above proposal

constraints/bitwise.air
Al-Kindi-0: done

Al-Kindi-0:
> As a general comment that would apply to other chiplets and traces, I think it would be useful to move away from single letter variables which were practical when formatting as LaTeX but which can now be a lot more descriptive in code.
> 

I agree, but on the other hand the equations defining the constraints start becoming harder to parse the longer variable names get. I think having a header comment, like a module doc, where we explain the purpose and functional name of each variable is a good compromise, but I am open to other suggestions as well

> If I'm being pedantic, it would be useful to have a quick overview of the chiplet at the top of the file, but on the other hand the docs are already quite descriptive. If we do chose to go for more descriptive names then I think it'll be worth mentioning the mapping.
> 

I agree, we can do this if we have agreement on this and the previous point

> I also left a few comments around optimizing some constraints. The description here is clear, and this may be hard to actually implement, but we should be able to save some multiplications in some places.

Agreed, and left some comments where relevant

constraints/bitwise.air
adr1anh: Agreed!

adr1anh:
Looks good! 

We should create an issue for all the potential plonky3 optimizations we could take advantage of, and the kind of expressions we're expecting to be simplified. (cc @Leo-Besancon et al.)

adr1anh:
> I agree, but on the other hand the equations defining the constraints start becoming harder to parse the longer variable names get. I think having a header comment, like a module doc, where we explain the purpose and functional name of each variable is a good compromise, but I am open to other suggestions as well

There's definitely a middle ground here, and I agree that longer variables make things harder to parse in the expression itself. We can always refactor this later with feedback from some fresh eyes.

adr1anh (approved):

constraints/bitwise.air
Al-Kindi-0: done

constraints/bitwise.air
Al-Kindi-0: done

constraints/bitwise.air
```diff
@@ -66,4 +59,100 @@ periodic_columns {
...
+    # output value (from the zp column in the current row) by 16 and then adding it to the bitwise
+    # operation applied to the row's set of bits of a_limb and b_limb. The entire constraint must
+    # also be multiplied by the operation selector flag to ensure it is only applied for the
+    # appropriate operation. The constraint for AND is enforced when op_flag = 0 and the constraint for
+    # XOR is enforced when op_flag = 1. Because the selectors for the AND and XOR operations are mutually
+    # exclusive, the constraints for different operations can be aggregated into the same result
+    # indices.
+    # Constraints degree: 3
+    let a_and_b = sum([2^i * a * b for (i, a, b) in (0..4, a_limb, b_limb)]);
+    let a_xor_b = sum([2^i * (a + b - 2 * a * b) for (i, a, b) in (0..4, a_limb, b_limb)]);
```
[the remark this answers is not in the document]
Al-Kindi-0: good catch, fixed now

adr1anh (approved):

adr1anh (approved):
Looks good!

PR #468, opened 2025-09-23
Update and move hasher chiplet constraints

### LINKED ISSUE #467: Update Hasher chiplet constraints

Al-Kindi-0:
## Describe your changes
Closes #467
Also includes #257

- Updated `CHANGELOG.md`


constraints/hasher.air
```diff
@@ -10,49 +10,214 @@
...
+
+# Enforce node index constraints
+ev node_index([s[3], i]) {
+    let f_out = get_f_out(s);
+    let f_mp = get_f_mp(s);
+    let f_mv = get_f_mv(s);
+    let f_mu = get_f_mu(s);
+    let f_mpa = get_f_mpa(s);
+    let f_mva = get_f_mva(s);
+    let f_mua = get_f_mua(s);
```
adr1anh: Could we explain what the subscripts refer to?

Al-Kindi-0: Good idea, added a section towards the end to explain each op flag, this way it will be easier to understand the logic of the op flags and avoid having to explain them at every occurrence.

constraints/hasher.air
```diff
@@ -10,49 +10,242 @@
...
+
+    # Flag that is true when the performed operation is one of the operations represented by flags
+    # f_abp, f_mpa, f_mva or f_mua
+    let f_comp = f_abp + f_mpa + f_mva + f_mua;
+
+    # Enforce that selector columns are binary.
+    enf is_binary([selector]) for selector in s;
```
adr1anh: Small nit: I'd put this before the flag computation

constraints/hasher.air
```diff
@@ -10,49 +10,242 @@
...
+
+    # Enforce that unless f_out = 1 or f_out' = 1, the values in columns s[1] and s[2] are copied 
+    # over to the next row.
+    enf is_unchanged([s[1]]) when !f_out & !f_out_next;
+    enf is_unchanged([s[2]]) when !f_out & !f_out_next;
```
adr1anh, line 62: From the docs, I don't fully understand why we avoid copying when `!f_out_next`

adr1anh: Ah, because if we did this in row 6, then we can't set the selectors for `h_out` or `s_out`

constraints/hasher.air
```diff
@@ -10,49 +10,242 @@
 # 4. Merkle path verification.
 # 5. Merkle root update.
 #
-# STATUS: Not implemented
+# STATUS: Partially implemented (missing bus interactions)
```
adr1anh, line 13: I think we're missing the constraint for `addr`, unless it uses the system clock column?

constraints/hasher.air
```diff
@@ -10,49 +10,242 @@
...
+
+    # Flag indicating to enforce the constraint that b is binary only when a new node is absorbed into
+    # the hasher state (when the hash operation is either one of Merkle path verification or 
+    # Merkle root update)
+    let f_an = f_mp + f_mv + f_mu + f_mpa + f_mva + f_mua;
+
+    # b is the value of the bit which is discarded during shift by one bit to the right.
+    let b = i - 2 * i';
+
+    # Enforce that b is binary only when a new node is absorbed into the hasher state.
```
adr1anh, line 99: The remaining rows contain the next index.

constraints/hasher.air
```diff
@@ -10,49 +10,242 @@
...
+    enf is_binary([selector]) for selector in s;
+
+    # Enforce that unless f_out = 1 or f_out' = 1, the values in columns s[1] and s[2] are copied 
+    # over to the next row.
+    enf is_unchanged([s[1]]) when !f_out & !f_out_next;
+    enf is_unchanged([s[2]]) when !f_out & !f_out_next;
+
+    # Enforce that if any of f_abp, f_mpa, f_mva, f_mua flags is set to 1, the next value of s[0] 
+    # is 0.
+    enf s[0]' * f_comp = 0;
```
adr1anh: If I understand correctly, if we're in the last row of an operation which carries over to the next block, then the next row (which is the first of the next block), we must perform an output?

I think I see now though. We're indicating with `s[0]` that this is a block in which we will output the value when we reach the end?

adr1anh (commented):
I gave it another pass focusing on the actual constraints. I'm still confused about the overall constraints, even with the docs. 

While I haven't fully thought things through, I think it would help if we state the constraints we apply at 
- first row
- transition rows
- penultimate row
- last row

constraints/hasher.air
Al-Kindi-0: done

constraints/hasher.air
Al-Kindi-0: yes, the selectors can change their values within a cycle only at the end

constraints/hasher.air
Al-Kindi-0: I think we use the `clk` column from the system columns and this is used AFAIK only in the bus constraints which are still to be implemented

constraints/hasher.air
Al-Kindi-0: not sure I follow the comment

constraints/hasher.air
Al-Kindi-0: Basically, if we are outputting a result in the current row then the next row should be the start of a new operation, which implies that `s[0]` must be `1` (note that `s[0]` partitions the hasher ops into two groups with the `1` group consisting of output ops).
If, on the other hand, the op of the next cycle is one of the absorption ops (which happen on rows 1 less than multiples of 8) then the `s[0]` in the next row (which is a multiple of 8 row) is set to `0` so that we prohibit any op which starts a (new) computation.

constraints/hasher.air
adr1anh: Sorry it was a note for me. All good

adr1anh (approved):
Looks good. We'll definitely need another more thorough audit of the actual contraints but for now it seems like they match the docs.

PR #471, opened 2025-09-24
fix: avoid state mutation on insert conflicts; use Entry API in Graph

MozirDmitriy:
Previously, mir/src/ir/graph.rs used BTreeMap::insert in insert_function, insert_evaluator, and insert_bus. On key conflicts, insert replaces the existing value before returning the old one. As a result, these methods could return Err(CompileError::Failed) while the map had already been mutated, leaving the graph in an inconsistent state.
This change switches to BTreeMap::entry:
functions/evaluators: 
- insert on Vacant; on Occupied, replace only if the existing value is a Root::None placeholder, otherwise return Err without changing the map.

buses: 
- insert on Vacant; on Occupied, return Err without changing the map.

### DIFF
--- a/mir/src/ir/graph.rs
+++ b/mir/src/ir/graph.rs
@@ -34,10 +34,14 @@ impl Graph {
         if node.as_function().is_none() {
             return Err(CompileError::Failed);
         }
-        match self.functions.insert(ident, node) {
-            None => Ok(()),
-            Some(link) => {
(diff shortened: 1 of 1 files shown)


Leo-Besancon:
This PR currently does not build (double borrow issues), and the underlying code seems to work fine (although it could be a bit cleaner). Closing to focus on higher-priority PRs.

PR #473, opened 2025-09-26
Enforce errors for missing items in Import::Partial

GarmashAlex:
- Switched partial import resolution to iterate requested items and resolve each via Module::get(&Identifier).
- If an item is not exported by the source module, return ImportUndefined(from) instead of silently ignoring it.
- This prevents no-op imports, aligns behavior with the ImportUndefined error semantics and user expectations for explicit imports, and improves diagnostics by surfacing incorrect use statements early.


Soulthym:
Hey!
I didnt find a case where the current implementation causes issues here.

Do you mind adding a test case, or at least sharing a set of air-script files where the issue is present, so we can compare this implementation and the existing one?

PR #477, opened 2025-09-27
feat(parser): fix selector check to use Span<u64>.item instead of deref

VolodymyrBg:
The selector in list comprehensions is ScalarExpr::Const(Span<u64>). Dereferencing Span<u64> (*selected == 0) is incorrect and inconsistent with the rest of the codebase, where Span<T> values are accessed via .item. This change uses selected.item == 0, ensuring correct semantics.


huitseeker (changes requested):
Deref impl matches item field access in https://github.com/0xMiden/miden-diagnostics/blob/f99efe05b8873723fcbb1ba22c31d15c9e4b5afe/diagnostics/src/span.rs#L221-L228, nothing wrong here except aesthetics.

PR #484, opened 2025-09-30
Kernel ROM constraints

Al-Kindi-0:
- Updated `CHANGELOG.md`

### DIFF
--- a/constraints/kernel_rom.air
+++ b/constraints/kernel_rom.air
@@ -1,0 +1,77 @@
+##########################################################################################
+# KERNEL ROM CHIPLET
+##########################################################################################
+#
+# The Kernel ROM chiplet is responsible for tracking execution of kernel (system) calls.
+# It maintains a record of all kernel procedure digests and ensures proper initialization
+# of kernel procedures while preserving privacy about call frequency.
+#
+# The chiplet contains digest values for all kernel procedures and enforces contiguity
+# constraints to ensure proper execution tracking.
+#
+# STATUS: Core constraints implemented, bus constraints not implemented
+#
+# REFERENCES:
+# - Kernel ROM Chiplet: https://0xmiden.github.io/miden-vm/design/chiplets/kernel_rom.html
+##########################################################################################
+
+mod kernel_rom
+
+use utils::*;
+
+##########################################################################################
+# KERNEL ROM CHIPLET TRANSITION CONSTRAINTS
+##########################################################################################
+
+# Enforces the constraints on the kernel ROM chiplet given its columns.
+#
+# Parameters:
+# - s4: Chiplet selector flag
+# - sfirst: Section first flag (1 = start of new digest block, 0 = continuation of a given block)
+# - r[4]: Kernel procedure root/digest (4 field elements)
+#
+# Max constraint degree: 3
+ev ker_rom_chiplet_constraints([s4, sfirst, r[4]]) {
+    enf kernel_rom_selector([sfirst]);
+    enf kernel_rom_digest_contiguity([s4, sfirst, r]);
+}
+
+# Enforces initialization constraints for the first row of kernel ROM chiplet execution.
+#
+# According to the official specification, the first row must have sfirst = 1 to ensure
+# proper digest matching and initialization of kernel procedure tracking.
+#
+# Max constraint degree: 1
+ev kernel_rom_chiplet_constraints_first_row([s4, sfirst, r[4]]) {
+    # FIRST ROW INITIALIZATION CONSTRAINT:
+    # The first row of any kernel ROM chiplet execution must have sfirst' = 1
+    # This ensures proper initialization of kernel procedure digest tracking.
+    # Without this, the chiplet could start in an invalid intermediate state.
+    enf sfirst' = 1;
+}
+
+##########################################################################################
+# HELPERS
+##########################################################################################
+
+# Enforces that the kernel ROM selector is valid.
+#
+# Max constraint degree: 2
+ev kernel_rom_selector([sfirst]) {
+    # Enforce that sfirst is binary (0 or 1)
+    enf is_binary([sfirst]);
+}
+
+# Enforces digest contiguity constraints for the kernel ROM.
+#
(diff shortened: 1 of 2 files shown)


constraints/kernel_rom.air
```diff
@@ -0,0 +1,68 @@
...
+    # Constraint degree: 2
+    enf is_binary([sfirst]);
+
+    # First row initialization constraint
+    # This is part of the ACE chiplet constraints
+}
+
+# Enforces digest contiguity constraints for the kernel ROM.
+#
+# When sfirst = 0, consecutive rows must contain identical digest values to ensure
```
adr1anh:
```suggestion
# When s_first' = 0, consecutive rows must contain identical digest values to ensure
```

We stop copying if the next row is the start of a new block

constraints/kernel_rom.air
```diff
@@ -0,0 +1,68 @@
...
+# proper tracking of kernel procedure executions.
+#
+# Max constraint degree: 3
+ev kernel_rom_digest_contiguity([s4, sfirst, r[4]]) {
+    # Enforce that when sfirst = 0 (not the start of a new digest block),
+    # the digest values in the current row must be identical to the next row.
+    # This ensures contiguous blocks of identical digest values.
+    #
+    # Note: This constraint is disabled when the next row is the last row of the kernel ROM chiplet
```
adr1anh:
```suggestion
    # Note: This constraint is disabled when the current row is the last row of the kernel ROM chiplet
```

constraints/kernel_rom.air
Al-Kindi-0: good catch, fixed

constraints/kernel_rom.air
Al-Kindi-0: fixed

constraints/kernel_rom.air
```diff
@@ -0,0 +1,82 @@
...
+# Max constraint degree: 3
+ev kernel_rom_digest_contiguity([s4, sfirst, r[4]]) {
+    # Enforce that when sfirst = 0 (not the start of a new digest block),
+    # the digest values in the current row must be identical to the next row.
+    # This ensures contiguous blocks of identical digest values.
+    #
+    # Note: This constraint is disabled when the current row is the last row of the kernel ROM chiplet
+    # which is indicated by s4' being set to 1
+    # 
+    # TODO: Check if the following inlines to what is expected
```
plafer: Is this still relevant? It's unlikely to be done if left as a TODO in the code - should we instead create an issue?

constraints/kernel_rom.air
```diff
@@ -0,0 +1,82 @@
...
+    enf is_unchanged([r]) for r in r when binary_and(binary_not(s4'), binary_not(sfirst'));
```
plafer:
```suggestion
    enf is_unchanged([r_i]) for r_i in r when binary_and(binary_not(s4'), binary_not(sfirst'));
```
nit to make this more readable

constraints/kernel_rom.air
```diff
@@ -0,0 +1,82 @@
...
+# KERNEL ROM CHIPLET TRANSITION CONSTRAINTS
+##########################################################################################
+
+# Enforces the constraints on the kernel ROM chiplet given its columns.
+#
+# Max constraint degree: 3
+ev ker_rom_chiplet_constraints([s4, sfirst, r[4]]) {
+    enf kernel_rom_selector([sfirst]);
+    enf kernel_rom_digest_contiguity([s4, sfirst, r]);
+}
```
plafer, line 37: Aren't we missing a constraint (or 4) to make sure that `sfirst` is set to 1 *only* for the first row of a set of equal roots? I think that currently it's always possible to set `sfirst = 1` for all rows.

This would also be missing in the docs.

constraints/kernel_rom.air
```diff
@@ -0,0 +1,82 @@
...
+use utils::*;
+
+##########################################################################################
+# KERNEL ROM CHIPLET TRANSITION CONSTRAINTS
+##########################################################################################
+
+# Enforces the constraints on the kernel ROM chiplet given its columns.
+#
+# Max constraint degree: 3
+ev ker_rom_chiplet_constraints([s4, sfirst, r[4]]) {
```
plafer, line 34:
```suggestion
ev ker_rom_chiplet_constraints([chiplet_selector_4, sfirst, root[4]]) {
```

I personally prefer descriptive names 🤷‍♂️

constraints/kernel_rom.air
Al-Kindi-0: The way the verifier uses variable-length public inputs is what prevents `sfirst = 1` from appearing multiple times for the same digest:

  - `sfirst = 1` sends a `vinit` message for that digest
  - The verifier initializes the bus with exactly one `vinit` per unique digest
  - If the prover sets `sfirst = 1` multiple times for the same digest, the bus won't balance

@adr1anh

constraints/kernel_rom.air
Al-Kindi-0: Agreed, though for consistency's sake I think it is better to keep as is. Added a comment to the main evaluator to introduce each variable

constraints/kernel_rom.air
Al-Kindi-0: agreed, changed

constraints/kernel_rom.air
Al-Kindi-0: Agreed, checking in-lining should be part of the sanity-checking work we will do to ensure the "soundness" of the passes  
Removed now

constraints/kernel_rom.air
adr1anh: To add to @Al-Kindi-0's explanation, it's technically possible to have the digest appear more than once with `sfirst=1`, though that would require the verifier to include that digest twice in the VLPI. The uniqueness check is therefore enforced by the verifier.

adr1anh (approved):
LGTM

constraints/kernel_rom.air
plafer: Ah yes - makes sense!

plafer (approved):
LGTM

Issue #486, opened 2025-10-02
Vector Slice Expansion in Evaluator Calls


Al-Kindi-0:
### Problem Description

The compiler fails to properly expand vector slice arguments when evaluators call other evaluators, leading to argument count mismatches.

### Minimal Example

```air
ev chiplets_constraints([chiplets[20]]) {
    enf chiplet_selectors([chiplets[0..5]]);  // Should expand slice to 5 args
}

ev chiplet_selectors([s[5]]) {              // Expects 5 individual parameters
    enf is_binary([s[0]]);
    enf is_binary([s[1]]) when s[0];
    enf is_binary([s[2]]) when s[0] & s[1];
    enf is_binary([s[3]]) when s[0] & s[1] & s[2];
    enf is_binary([s[4]]) when s[0] & s[1] & s[2] & s[3];
}
```

PR #495, opened 2025-10-16
Fix evaluator argument vector slice expansion

Soulthym:
This PR solves https://github.com/0xMiden/air-script/issues/486

It adds special handling for remaining slices when translating to mir, and generalizes evaluator argument checking and unpacking when inlining to unwrap Accessors and support Vectors.

- [x] Fix: translate remaining slices to `Vector<Accessor(MirAccessType::Index(_), )>`
- [x] Fix: support newly inserted Vectors in inlining
- [x] Fix: inlining through accessor
- [x] Tests: add new test for the slice evaluator argument edgecase.
- [x] Document api changes & changed invariants.


Soulthym:
This should be ready for review!

Al-Kindi-0 (commented):
Did an initial pass and it looks good, thanks!
Could we make a test for slicing of slices?

mir/src/passes/inlining.rs
```diff
@@ -538,6 +539,10 @@ fn check_evaluator_argument_sizes(
         let children = trace_segments_arg_vector.children();
         let mut trace_segments_arg_vector_len = 0;
         for child in children.borrow().deref() {
+            // We need to extract the outermost accessors, if any.
+            // This can happen when slices are used to rebind
+            // trace bindings in nested evaluator calls
+            let child = extract_accessor(child.clone());
```
adr1anh: nit: no need to clone i think

mir/src/passes/inlining.rs
```diff
@@ -555,31 +560,37 @@ fn check_evaluator_argument_sizes(
...
-                    unreachable!("expected value or parameter, got {:?}", child);
-                }
+            } else if let Some(vector) = child.as_vector() {
+                // When slices are used to rebind trace bindings in nested evaluator calls,
+                // translation from Ast to Mir inserts a Vector of
+                // Accessor(MirAccessType::Index), representing the unwrapped
+                // slice.
+                // We need to make sure that this vector is a vector of Felts
+                // and to add its length to the total length of arguments
+                assert!(
```
adr1anh: nit: I'd replace this with a for loop that calls assert

mir/src/passes/inlining.rs
```diff
@@ -555,31 +560,37 @@ fn check_evaluator_argument_sizes(
...
-                    let size = match ty {
-                        MirType::Felt => 1,
-                        MirType::Vector(len) => *len,
-                        _ => unreachable!("expected felt or vector, got {:?}", ty),
-                    };
-                    trace_segments_arg_vector_len += size;
-                } else {
-                    unreachable!("expected value or parameter, got {:?}", child);
-                }
+            } else if let Some(vector) = child.as_vector() {
```
adr1anh: I might be wrong as I don't have full context. Would it make sense to consider `value` as a vector with a single element? We would merge the first and last branch it seems?

adr1anh (approved):
Looks good, it seems like this addresses the issue and solves the original issue,.

Soulthym:
> Did an initial pass and it looks good, thanks! Could we make a test for slicing of slices?

Added 2 tests:
- nested slices alone (a[0..10][0..5])
- nested slices in evaluator args

Nested slices alone seem to break the parser.

mir/src/passes/inlining.rs
```diff
@@ -611,6 +622,45 @@ fn check_evaluator_argument_sizes(
...
+            match indexable.borrow().deref() {
+                Op::Vector(v) => v.children().borrow()[idx].clone(),
+                Op::Matrix(m) => m.children().borrow()[idx].clone(),
+                _ => unreachable!("expected vector or matrix, got {:#?}", indexable),
+            }
+        },
+        MirAccessType::Matrix(row, col) => {
+            let row = get_inner_const(row).expect("expected constant row, got {row:#?}") as usize;
+            let col =
+                get_inner_const(col).expect("expected constant column, got {col:#?}") as usize;
```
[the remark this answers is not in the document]
Soulthym: fixed both in b17be10

mir/src/passes/inlining.rs
Soulthym: dereferencing doesnt seem to be possible here. I think I'll keep the clone for now

mir/src/passes/inlining.rs
Soulthym: fixed in 42e9b9d

mir/src/passes/inlining.rs
Soulthym: In this if branch, `child` will be a vector of N elements, each element being an Accessor(.., AccessType::Index). It is the translated version of a slice, since a slice is not currently expressible in MIR.

So no this vector can have N children, one per index the slice range enumerates.

Other vectors should not happen in this case, as its a result of the translation from Ast to Mir for slices only. Considering this case didn't trigger before this PR, and that we only added support for slicing in evaluator arguments, this should be the only kind of vector we find here.

Soulthym:
@adr1anh @Al-Kindi-0 I have applied fixes based on your reviews, should be ready on my side!

mir/src/passes/translate.rs
```diff
@@ -1053,6 +1053,39 @@ impl<'a> MirBuilder<'a> {
...
+    // This is used to completely eliminate slice accesses in MIR
+    fn translate_potential_slice(
+        &mut self,
+        access_expr: &Link<Op>,
+        access: &'a ast::SymbolAccess,
+    ) -> Option<Link<Op>> {
+        // If it's a slice access, we need to create a vector of MirAccessType::Index
+        if let AccessType::Slice(ast::RangeExpr { start, end, .. }) = &access.access_type {
+            let ast::RangeBound::Const(Span { item: start, .. }) = start else {
+                unreachable!("expected constant start in slice access, got: {:#?}", start);
```
huitseeker: The current slice implementation could provide clearer error messages for users. 

1. Instead of the `unreachable!()` with a technical message, could we do stg like "Slice expressions must use constant integer bounds (e.g., `arr[0..5]`). Dynamic bounds like variables or expressions are not yet supported."

2. The error for intermediate variable slicing is unclear. Is it an internal error, signaling a contract violation of contract violation, or could we be more clear this is as-yet-unsupported?

3. Could give better guidance on empty slices: "Slice range is empty (start >= end). Check your slice bounds."

mir/src/passes/inlining.rs
```diff
@@ -555,31 +560,41 @@ fn check_evaluator_argument_sizes(
...
-                        MirType::Felt => 1,
-                        MirType::Vector(len) => *len,
-                        _ => unreachable!("expected felt or vector, got {:?}", ty),
-                    };
-                    trace_segments_arg_vector_len += size;
-                } else {
-                    unreachable!("expected value or parameter, got {:?}", child);
+            } else if let Some(vector) = child.as_vector() {
+                // When slices are used to rebind trace bindings in nested evaluator calls,
+                // translation from Ast to Mir inserts a Vector of
```
huitseeker: Nit: the argument processing logic in `check_evaluator_argument_sizes()` and `unpack_evaluator_arguments()` has some duplication - both functions iterate through vector children and call `extract_accessor()` with similar validation patterns. 

I'd consider extracting a shared helper like `process_slice_vector_children()` that returns both the count and the unpacked arguments to reduce this duplication.

air-script/tests/evaluators/evaluators_slice.air
```diff
@@ -0,0 +1,33 @@
+def EvaluatorsSliceAir
```
huitseeker, line 1: This PR comes with both strong points (simple range expressions) and some user-visible limitations (nesting, expressions in ranges).

I noticed the documentation could be enhanced to help users understand the current slice capabilities and limitations. Here are some suggestions:

**For \`docs/src/description/syntax.md\`** - I'd add a new "Slice Expressions" section: where I'd highlight 
the Requirements:
- **Constant bounds**: Both \`start\` and \`end\` must be constant integers (e.g., \`a[0..5]\`)
- **Direct access only**: Slicing intermediate variables is not yet supported (e.g., \`let sub = a[0..5]; sub[1..3]\`) (optional, see my other comment)
- **Forward ranges**: \`end\` must be greater than \`start\` (no empty ranges)

**For \`docs/src/description/evaluators.md\`** - Add to the "Using evaluators" section:

```
#### Slice expressions in evaluator calls

Slice expressions can be used in evaluator arguments to automatically expand ranges into individual parameters:

\`\`\`air
ev chiplet_selectors([s[5]]) { /* expects 5 individual params */ }

ev constraints([cols[20]]) {
    // chiplets[0..5] automatically expands to 5 individual arguments
    enf chiplet_selectors([chiplets[0..5]]); 
}
\`\`\`
```

huitseeker (commented):
This looks globally good, but I've added a few ways in which the PR could make the current status of the work a bit more clear.

mir/src/passes/translate.rs
Soulthym: 1 & 3 were implemented in 26be737f

> 2. The error for intermediate variable slicing is unclear. Is it an internal error, signaling a contract violation of contract violation, or could we be more clear this is as-yet-unsupported?

This currently seems to break the parser, I haven't yet found a better way of reporting that error in the general case. Leaving it as a followup issue as it might require a larger refactor of the front-end than the scope of this PR. The issue is: https://github.com/0xMiden/air-script/issues/502

Currently, you can bypass that limitation by nesting evaluator calls:

```air-script
trace_columns {
    main: [a[10]],
}
integrity_constraints {
    enf f(a[0..5]);
}
ev f([a[5]) {
    enf g([a[2..5]);
}
ev g([a[3]) {
    enf x^2 = x for x in a;
}
```

mir/src/passes/inlining.rs
Soulthym: Fixed in 18dddb4

air-script/tests/evaluators/evaluators_slice.air
Soulthym: Added in ca2c2d1

Soulthym:
Thanks for the reviews @huitseeker!

All points have been addressed.

@Al-Kindi-0 this should be ready!

PR #496, opened 2025-10-16
fix: unwrap Enf safely when combining constraints in MatchOptimizer

radik878:
## Describe your changes

The combined main constraints were constructed by multiplying the selector with the representative constraint node, which could be an Enf, creating Mul(selector, Enf(expr)) under Sub(..., 0). This violates the intended shape where main constraints are pure arithmetic expressions enforced to zero, and makes downstream assumptions brittle. The change unwraps Enf to its inner expression before multiplication so selectors are applied to the arithmetic expr itself, aligning with the documented contract and the handling of bus constraints.

### DIFF
--- a/mir/src/passes/unrolling/match_optimizer.rs
+++ b/mir/src/passes/unrolling/match_optimizer.rs
@@ -210,11 +210,15 @@ impl<'a> MatchOptimizer<'a> {
                             Some(Add::create(cur_condition.unwrap(), condition.clone(), span));
                     }
                 }
-                let new_node = Mul::create(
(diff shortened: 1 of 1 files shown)


Leo-Besancon:
Nice catch, thank you! I think the Enf node handling indeed can be improved and made more robust. However, I think this PR in isolation does not solve the issue (we could potentially remove a needed Enf without re-wrapping it).

Instead, I think that improving the checks on the graph assumptions for each pass is the better way to go (see item 6 in https://github.com/0xMiden/air-script/issues/504).

PR #503, opened 2025-10-25
docs: clarify bus constraint variable meanings

crStiv:
## Describe your changes

Added explanations for p/q and p'/q' variables in bus constraint equations.


Leo-Besancon:
Thank you, but the current doc seems sufficient to me as the next row operator is already explained in previous pages.

PR #507, opened 2025-11-04
Support importing hierarchical modules

### LINKED ISSUE #488: Air-Script Import Syntax Insufficient for Complex Module Hierarchies
Air-script's current import syntax lacks support for hierarchical module structures, limiting code organization capabilities for complex constraint systems. The language cannot express nested module relationships or import from deeply nested module paths.

Leo-Besancon:
## Describe your changes

This PR aims to close https://github.com/0xMiden/air-script/issues/488.
The main change is to handle `ModuleId` as a `Vec<Identifier>` instead of just `Identifier`. It allows to represent nested modules through there namespaced path (e.g. `utils::hash::my_hash_module` for the module that is in the file `utils/hash/my_hash_module.air`).

Main limitations:
1. Currently, imports are only allowed with full paths (relative to the importing module). As a result (with the example above), we can do `use utils::hash::my_hash_module::*` but the following are not supported:
  - `use utils::hash::*;`
  - `use utils::*;`
  - `use utils::hash;`
2. It is not possible to declare a module that is the *direct* parent of another. This means that we cannot declare the files `utils.air` or `utils/hash.air` in the example above. The hierarchy is only defined by the folder structure.

@Al-Kindi-0 would this be enough, or would you need one or both of these limitations lifted? I think the first one could be handled if needed, but I had some issues with import resolutions. For the second one, I'm not currently sure of the difficulty.

### DIFF
--- a/parser/src/ast/mod.rs
+++ b/parser/src/ast/mod.rs
@@ -496,22 +496,23 @@ impl Library {
         // to modules in the library. If the module is already in the library, we proceed,
         // if it isn't, then we must parse the desired module from disk, and add it to the
         // library, visiting any of its imports as well.
-        while let Some((module, mut imports)) = worklist.pop_front() {
+        while let Some((_module, mut imports)) = worklist.pop_front() {
             // We attempt to resolve imports on disk relative to the file path of the
             // importing module, if it was parsed from disk. If no path is available,
             // we default to the current working directory.
 
-            let source_dir = match codemap.name(imports.first().unwrap().span().source_id()) {
-                // If we have no source span, default to the current working directory
-                Err(_) => cwd.clone(),
-                // If the file is virtual, then we've either already parsed imports for this module,
-                // or we have to fall back to the current working directory, but we have no relative
-                // path from which to base our search.
-                Ok(FileName::Virtual(_)) => cwd.clone(),
-                Ok(FileName::Real(path)) => {
-                    path.parent().unwrap_or_else(|| Path::new(".")).to_path_buf()
-                },
-            };
+            let (real_path, source_dir) =
+                match codemap.name(imports.first().unwrap().span().source_id()) {
+                    // If we have no source span, default to the current working directory
+                    Err(_) => (false, cwd.clone()),
+                    // If the file is virtual, then we've either already parsed imports for this module,
+                    // or we have to fall back to the current working directory, but we have no relative
+                    // path from which to base our search.
+                    Ok(FileName::Virtual(_)) => (false, cwd.clone()),
+                    Ok(FileName::Real(path)) => {
+                        (true, path.parent().unwrap_or_else(|| Path::new(".")).to_path_buf())
+                    },
+                };
 
             // For each module imported, try to load the module from the library, if it is
             // unavailable we must do extra work to load it into the library, as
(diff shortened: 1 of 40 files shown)


Al-Kindi-0:
One question @Leo-Besancon , there is nothing against defining `utils/hash/mod.air` in the above, right?

Leo-Besancon:
> One question @Leo-Besancon , there is nothing against defining `utils/hash/mod.air` in the above, right?

I think the only thing preventing it is that mod is a reserved keyword to declare modules, so in my test the following fails:
```
15 │ use utils::nested::mod::*;
   │                    ^^^ expected one of: '"*"', 'identifier'
```
But maybe we could say that `use utils::nested;` resolves to the module in `utils/hash/mod.air` automatically.

Al-Kindi-0:
> > One question @Leo-Besancon , there is nothing against defining `utils/hash/mod.air` in the above, right?
> 
> I think the only thing preventing it is that mod is a reserved keyword to declare modules, so in my test the following fails:
> 
> ```
> 15 │ use utils::nested::mod::*;
>    │                    ^^^ expected one of: '"*"', 'identifier'
> ```
> 
> But maybe we could say that `use utils::nested;` resolves to the module in `utils/hash/mod.air` automatically.

That could work too, I believe.

parser/src/ast/module.rs
```diff
@@ -1,11 +1,59 @@
...
+use std::{
+    collections::{BTreeMap, HashSet},
+    ops::Index,
+};
 
 use miden_diagnostics::{DiagnosticsHandler, Severity, SourceSpan, Span, Spanned};
 
 use crate::{ast::*, sema::SemanticAnalysisError};
 
 /// This is a type alias used to clarify that an identifier refers to a module
```
Al-Kindi-0: probably needs updating?

parser/src/ast/mod.rs
```diff
@@ -181,28 +197,40 @@ impl Program {
...
-            let analysis =
-                sema::SemanticAnalysis::new(diagnostics, &program, &library, &mut deps, imported);
+            let analysis = sema::SemanticAnalysis::new(
+                diagnostics,
+                &program,
+                &library,
+                &mut deps_graph,
+                &mut deps_nodes,
+                imported,
+            );
```
Al-Kindi-0, line 236: would it make sense to update the comment and explain briefly what is done here? (at least given the changes in this PR)

Al-Kindi-0 (approved):
LGTM
The features introduced in this PR seems sufficient to me, but I will let the others also comment on this point

parser/src/ast/module.rs
Leo-Besancon: updated

parser/src/ast/mod.rs
Leo-Besancon: > would it make sense to update the comment and explain briefly what is done here? (at least given the changes in this PR)

Thanks, I've updated it in the latest commit, don't hesitate to ping me if it can still be improved

Leo-Besancon:
> > > One question @Leo-Besancon , there is nothing against defining `utils/hash/mod.air` in the above, right?
> > 
> > 
> > I think the only thing preventing it is that mod is a reserved keyword to declare modules, so in my test the following fails:
> > ```
> > 15 │ use utils::nested::mod::*;
> >    │                    ^^^ expected one of: '"*"', 'identifier'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But maybe we could say that `use utils::nested;` resolves to the module in `utils/hash/mod.air` automatically.
> 
> That could work too, I believe.

I've added that part to, and it seems to work on my example.
You can define the module in the file `utils/hash/mod.air` with:
```
mod hash

[...]
```
And in e.g. the root:
```
use utils::hash::*;
```

adr1anh (approved):
LGTM

PR #508, opened 2025-11-05
Target 0xMiden Plonky3 repo and use AirScriptAir and AirScriptBuilder traits

Leo-Besancon:
## Describe your changes

The PR introduces:
- the `AirScriptAir` and `AirScriptBuilder` traits proposed in https://github.com/0xMiden/Plonky3/issues/5 (for now, in the AirScript repo, but will latter be imported from 0xMiden/Plonky3 once our testing is done)
- Aux trace constraints for Plonky3 Codegen using the PermutationBuilder trait
- An adaptation of the DebugConstraintBuilder struct that generates an aux trace (hardcoded for the only Air that is tested against the aux trace, buses_complex_plonky3::BusesAir)


Leo-Besancon:
@adr1anh here I assumed that the randomness was given by:
```rust
            Value::RandomValue(idx) => {
                if *idx == 0 {
                    format!("beta.into()")
                } else {
                    format!("alpha_challenges[{}].into()", idx - 1)
                }
            },
```

But maybe it should only be alpha_challenges for the constraints, beta being used internally in the prover/verifier?

air-script/tests/helpers/mod.rs
```diff
@@ -1,6 +1,11 @@
-mod plonky3_periodic_columns;
+// Deprectated: use updated AirScriptAir and AirScriptBuilder traits instead
```
adr1anh: I think we should more this somewhere else because it's difficult to see when the helpers change vs just the tests

air-script/tests/helpers/plonky3_traits.rs
```diff
@@ -0,0 +1,416 @@
...
+        0
+    }
+
+    /// Number of alpha challenges used in the AIR.
+    fn num_alpha_challenges(&self) -> usize {
+        0
+    }
+
+    /// Periodic constants (base-field) backing periodic_evals().
+    // fn periodic_table(&self) -> &'static [&'static [F]];
```
adr1anh: remove?

codegen/plonky3/src/air/integrity_constraints.rs
```diff
@@ -1,22 +1,30 @@
...
-
-        // If the constraint is a transition constraint (depends on the next row), we do not
-        // evaluate it in the last row, with the `when_transition` method.
-        let assertion = if let ConstraintDomain::EveryFrame(_) = constraint.domain() {
-            format!("builder.when_transition().assert_zero::<_>({expr_root_string});")
-        } else {
-            format!("builder.assert_zero::<_>({expr_root_string});")
-        };
+    let constraints = ir.integrity_constraints(TraceSegmentId::Main);
+    if !constraints.is_empty() {
```
adr1anh: Maybe we can remove this and always print it

codegen/plonky3/src/air/integrity_constraints.rs
```diff
@@ -1,22 +1,30 @@
...
+    for constraint in constraints {
+        let assertion = constraint_to_string(ir, constraint, false);
+        eval_func.line(assertion);
+    }
+}
 
+/// Adds the aux integrity constraints to the generated code.
+pub(super) fn add_aux_integrity_constraints(eval_func: &mut Function, ir: &Air) {
+    let constraints = ir.integrity_constraints(TraceSegmentId::Aux);
+    if !constraints.is_empty() {
```
adr1anh: same as above, we can always print the comment

adr1anh (approved):
looks great just left some small comments

air-script/tests/helpers/plonky3_traits.rs
```diff
@@ -0,0 +1,416 @@
...
+
+            let trace = generate_trace_rows::<Val>(inputs);
+
+            check_constraints_with_airscript_traits::<Goldilocks, Challenge, $air_name>(&$air_name {}, &trace, &inputs_goldilocks);
+
+            /*let prove_with_periodic_columns = prove_with_periodic_columns(&config, &BitwiseAir {}, trace, &inputs_goldilocks);
+            verify_with_periodic_columns(&config, &BitwiseAir {}, &prove_with_periodic_columns, &inputs_goldilocks).expect("Verification failed");*/
+
+            /*let proof = prove(&config, &BitwiseAir {}, trace, &inputs_goldilocks);
+            verify(&config, &BitwiseAir {}, &proof, &inputs_goldilocks).expect("Verification failed");*/
```
Al-Kindi-0: should these be removed and/or replaced with an issue?

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -0,0 +1,416 @@
...
+    /// A flag indicating whether this is a transition row (not the last row).
+    is_transition: F,
+    /// The periodic columns provided for constraint validation.
+    periodic_columns: Vec<EF>,
+    /// The alpha challenge in the extension field.
+    alpha: EF,
+    /// The alpha powers in the extension field.
+    alpha_powers: Vec<EF>,
+    /// The beta challenge in the extension field.
+    beta: EF,
```
Al-Kindi-0: I think the roles of alpha and beta should be reversed namely the randomness vector should be `[alpha, beta^0, beta^1, ...]`
I believe the reason for this comes from the ACE codegen but the reversion there is only done for optimization purposes. Apologizes if the comments there were not sufficient to clarify this.

air-script/tests/helpers/plonky3_traits.rs
```diff
@@ -0,0 +1,416 @@
...
+    /// The alpha challenge in the extension field.
+    alpha: EF,
+    /// The alpha powers in the extension field.
+    alpha_powers: Vec<EF>,
+    /// The beta challenge in the extension field.
+    beta: EF,
+    /// The aux bus boundary values in the extension field.
+    aux_bus_boundary_values: Vec<EF>,
+    /// The aux trace as a vertical pair.
+    permutation: VerticalPair<RowMajorMatrixView<'a, EF>, RowMajorMatrixView<'a, EF>>,
```
Al-Kindi-0: nit: should this be moved next to `main` and renamed to `aux`?

Al-Kindi-0 (approved):
LGTM
Would be great to think of a way to compare the outputs of Winterfell against P3. Haven't spent time thinking about this so not sure how realistic and straightforward this is.

adr1anh:
From the P3 side, we should be able to define an air builder which evaluates the constraints at a set of random points (there's already something super close), but not sure how we can do with with Winterfell

Al-Kindi-0:
For Winterfell, we might be able to use the verifier constraints evaluator (or a small modification of it) in order to evaluate the constraints on an OOD evaluation frame.

air-script/tests/helpers/plonky3_traits.rs
Leo-Besancon: Created the following issue: https://github.com/0xMiden/air-script/issues/512

air-script/tests/helpers/mod.rs
Leo-Besancon: reworked a bit the stucture. The diff for this might be a bit hard to read, but basically:
- moved all the tests into `air-script/src/tests`
- moved all the test helpers (codegen, macros, traits) into `air-script/src/test_utils`
- cleaned up a bit (some files were not checked by cargo / not included in the E2E pipeline, added `scripts/generate_all_e2e_tests`)

Please let me know if you think this structure is better and if it can still be improved

adr1anh (approved):

Leo-Besancon:
> LGTM Would be great to think of a way to compare the outputs of Winterfell against P3. Haven't spent time thinking about this so not sure how realistic and straightforward this is.

Hi, I've tested a few things out, and was able to adapt [Winterfell's `evaluate_constraints`](https://github.com/facebook/winterfell/blob/main/verifier/src/evaluator.rs#L15) on one side and [Plonky3's `VerifierConstraintFolder`](https://github.com/0xMiden/Plonky3/blob/main/uni-stark/src/verifier.rs#L196) one the other and evaluate a random frame for both with a simple Air.

It seems to work with a few caveats (for now, the code is specific to the air and a given row, it's a bit messy, and it's on a simple case with no public inputs, no periodic columns, no aux trace.

I think I'll put it in another PR if I find a way to clean it up.

air-script/src/test_utils/air_tester_macros.rs
```diff
@@ -36,12 +36,12 @@ macro_rules! generate_air_winterfell_test {
...
 /// * `air_name` - The identifier for the AIR struct (e.g., `BinaryAir`)
 #[macro_export]
-macro_rules! generate_air_plonky3_test {
+macro_rules! generate_air_plonky3_test_with_airscript_traits {
     ($test_name:ident, $air_name:ident) => {
         #[test]
         fn $test_name() {
             type Val = Goldilocks;
-            type Challenge = BinomialExtensionField<Val, 3>;
+            type Challenge = BinomialExtensionField<Val, 5>;
```
Al-Kindi-0: nit: it is probably enough to use just degree 2 here making things lighter, though it shouldn't matter really

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -0,0 +1,367 @@
...
+pub trait AirScriptAir<F: Field, AB: AirScriptBuilder<F = F>> {
+    /// Auxiliary width of the AIR.
+    fn aux_width(&self) -> usize {
+        0
+    }
+
+    /// Number of beta challenges used in the AIR.
+    fn num_beta_challenges(&self) -> usize {
+        0
+    }
```
Al-Kindi-0: nit: I would add a comment, and change the name of this method though not sure how easy it is, explaining that the number of beta challenge powers is equal to the width of the largest/widest message sent through all buses in an AIR instance

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -0,0 +1,367 @@
...
+
+/// Target trait for AirScript codegen. Implemented by the prover.
+pub trait AirScriptBuilder:
+    AirBuilder + AirBuilderWithPublicValues + ExtensionBuilder + PermutationAirBuilder
+where
+    <Self as AirBuilder>::F: Field,
+{
+    /// EF evaluations of periodic columns at the AIR’s random point (z). Order defined by
+    /// AirScript.
+    fn periodic_evals(&self) -> &[<Self as ExtensionBuilder>::VarEF];
```
Al-Kindi-0: Q: shouldn't this take `z` as input?

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -0,0 +1,367 @@
...
+
+    /// Global challenges in EF. (We can provide defaults; details not important here.)
```
Al-Kindi-0: I would remove the parenthesis and what is between them

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -0,0 +1,367 @@
...
+    fn alpha(&self) -> <Self as ExtensionBuilder>::VarEF;
+    fn beta(&self) -> <Self as ExtensionBuilder>::VarEF;
+    fn beta_powers(&self) -> &[<Self as ExtensionBuilder>::VarEF];
+
+    /// Aux bus boundary values: EF finals, one per aux/bus column, carried in the proof.
```
Al-Kindi-0: I would write this as:
```rust
Auxiliary buses boundary values, one per bus column, attached to the proof.
```

or something better

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -0,0 +1,367 @@
...
+    fn aux_bus_boundary_values(&self) -> &[<Self as ExtensionBuilder>::VarEF];
+}
+
+/// A builder that runs constraint assertions during testing.
+///
+/// Used in conjunction with [`check_constraints`] to simulate
+/// an execution trace and verify that the AIR logic enforces all constraints.
```
Al-Kindi-0: nit: alignment of the first line

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -0,0 +1,367 @@
...
+        + EF::from_u64(3) * beta_challenges[0]
+        + EF::from(main_current[0].clone()) * beta_challenges[1])
+        * (alpha
+            + EF::from_u64(3) * beta_challenges[0]
+            + EF::from(main_current[0].clone()) * beta_challenges[1])
+        * EF::from(main_current[6].clone());
+    let logup_current = EF::from(aux_current[1].clone());
+    let logup_next = (a * logup_current + b + c - e) * d.inverse();
+
+    // Dummy implementation for illustration purposes.
```
Al-Kindi-0: Q: Should we create a TODO to remove this dummy implementation with something more solid?

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -0,0 +1,367 @@
...
+) where
+    F: Field,
+    EF: ExtensionField<F>,
+    A: for<'a> Air<DebugConstraintBuilderWithAirScriptTraits<'a, F, EF>>,
+    A: for<'a> AirScriptAir<F, DebugConstraintBuilderWithAirScriptTraits<'a, F, EF>>,
+{
+    let height = main.height();
+
+    let aux_bus_boundary_values: Vec<_> = (0..air.aux_width()).map(|_| EF::GENERATOR).collect();
+    let alpha_f: Vec<F> = (0..5).map(|i| F::from_u64(123456789 * i)).collect(); // Dummy alpha in F
```
Al-Kindi-0: nit: can't we just use EF directly and hence avoid leaking the degree extension (i.e., 5)?

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -0,0 +1,367 @@
...
+    let alpha = EF::from_basis_coefficients_iter(alpha_f.iter().cloned()).unwrap();
+    let beta = EF::from_u64(987654321);
```
Al-Kindi-0, line 135: like this

air-script/src/test_utils/plonky3_traits.rs
Al-Kindi-0: I think this will be removed anyway once we have full support of auxiliary traces, right?

air-script/src/tests/computed_indices/test_air_plonky3.rs
```diff
@@ -0,0 +1,73 @@
+use std::marker::PhantomData;
+
+use p3_challenger::{HashChallenger, SerializingChallenger64};
+use p3_circle::CirclePcs;
```
Al-Kindi-0: Q: why do we need this?

Al-Kindi-0 (approved):
LGTM, great stuff!
I haven't checked every line of the codegen-ed code but I am assuming it should not be different from the Winterfell one. Also, I believe there is some work going to test consistency between the P3 and Winterfell backends

air-script/src/test_utils/air_tester_macros.rs
Soulthym: fixed in 5f2f3518

air-script/src/test_utils/plonky3_traits.rs
Soulthym: renamed to max_beta_challenge_power and added a comment on 871058c3

air-script/src/test_utils/plonky3_traits.rs
Soulthym: The comment does seem to insinuate that, but I dont think periodic evals depend on the random point, they're always 1 or 0 for the same row, no matter the input.

Im leaving it as is for now, Ill sync with @Leo-Besancon about this to confirm.

air-script/src/test_utils/plonky3_traits.rs
Soulthym: Added a TODO

> I think this will be removed anyway once we have full support of auxiliary traces, right?

I think so but not certain, Ill sync with @Leo-Besancon on this

air-script/src/test_utils/plonky3_traits.rs
Soulthym: fixed in 1d144757

air-script/src/test_utils/plonky3_traits.rs
Soulthym: is this about https://github.com/0xMiden/air-script/pull/508#discussion_r2560033915 ?

air-script/src/tests/computed_indices/test_air_plonky3.rs
Soulthym: I think this has to do with the macro generate_air_plonky3_test_with_airscript_traits that seems to use it.

Not sure why we're using this one tho, Ill see with Leo so we can use the proper imports.

In any case, the macro should use full path names so we do not need to import those dependencies explicitly.

air-script/src/test_utils/plonky3_traits.rs
Soulthym: done

air-script/src/test_utils/plonky3_traits.rs
adr1anh: The caller which creates the Builder will have already evaluated the columns at z, to avoid doing it at every row.

air-script/src/test_utils/plonky3_traits.rs
Leo-Besancon: Hello! @Al-Kindi-0 This was used to build the aux trace for the bus complex air (that only worked with this specific air), but in https://github.com/0xMiden/air-script/pull/515 it has already been replaced by a dynamic building of the aux trace that works with any Air (in particular the `buses_transitions` function added in the Codegen here: https://github.com/0xMiden/air-script/pull/515/files#diff-ed5104fe9ad8f138c0530924ab07f8c45e2ed62332f7475757ec7bf657d8dd77R227-R262)

This way, we now have two ways to build the Aux trace, either with the codegen, or with the manual implementations given by the MidenVM repo.

Just so it is clear, the Plonky3 PRs are:
- https://github.com/0xMiden/air-script/pull/461 that outlines the base for the main trace codegen
- https://github.com/0xMiden/air-script/pull/508 (this one), based on #461, that adds manual aux trace handling and periodic columns
- https://github.com/0xMiden/air-script/pull/515, based on #508, that uses the super traits and adds automatic aux trace building.

It would be great to merge these PRs in a Plonky3 codegen tracking branch once they are ready :)

air-script/src/tests/computed_indices/test_air_plonky3.rs
Leo-Besancon: In #515 (https://github.com/0xMiden/air-script/pull/515/commits/53eb542018aaff659f33531ae9e0f2a6a5eb28ce), I refactored the macro to import the dependencies explicitly and to use `TwoAdicFriPcs` instead of `CirclePcs`

PR #509, opened 2025-11-09
fix(mir): constant propagation folds 0^0 to 1

Galoretka:
## Describe your changes

The issue was in the MIR constant propagation pass where exponentiation folding handled the base-equals-zero case before checking for a zero exponent, causing 0^0 to fold to 0. This contradicted both the AST constant folding, which uses checked_pow and yields 1 for 0^0, and the MIR evaluator, which evaluates x^0 as 1 in the field. The fix changes the folding order in visit_exp_bis so that we fold x^0 to 1 first and only fold 0^k to 0 when k is a known, nonzero constant, delegating to generic constant folding otherwise. A MIR-level unit test was added to assert that 0^0 folds to a constant 1, guarding against regressions even if AST behavior changes.


mir/src/tests/constant_propagation_exp.rs
```diff
@@ -0,0 +1,57 @@
+use air_pass::Pass;
+use miden_diagnostics::{CodeMap, DiagnosticsConfig, DiagnosticsHandler, Verbosity};
+use miden_diagnostics::term::termcolor::ColorChoice;
+
+use crate::passes::ConstantPropagation;
+use crate::ir::{Enf, Exp, Link, Mir, MirValue, Op, SpannedMirValue, Value, ConstantValue};
+
+#[test]
+fn constant_propagation_folds_zero_pow_zero_to_one() {
```
Leo-Besancon: Could you remove the test (it does not build as is, is not included when running `cargo test` and it does not seem needed to me)

mir/src/passes/constant_propagation.rs
```diff
@@ -91,16 +91,23 @@ impl ConstantPropagation<'_> {
...
+        // x^0 = 1 (must take precedence, including 0^0)
+        if let Some(0) = get_inner_const(&rhs) {
             Ok(Some(Value::create(SpannedMirValue {
                 value: MirValue::Constant(ConstantValue::Felt(1)),
                 span: exp_ref.span,
             })))
+        } else if let Some(0) = get_inner_const(&lhs) {
+            // 0^k = 0, but only when k is known and non-zero
+            if let Some(k) = get_inner_const(&rhs)
+                && k != 0
```
Leo-Besancon: The case "k == 0" is already handled in the check "rhs = 0" above, you do not need it here.

mir/src/passes/constant_propagation.rs
```diff
@@ -91,16 +91,23 @@ impl ConstantPropagation<'_> {
...
+            {
+                return Ok(Some(Value::create(SpannedMirValue {
+                    value: MirValue::Constant(ConstantValue::Felt(0)),
+                    span: exp_ref.span,
+                })));
+            }
+            try_fold_const_binary_op(lhs, rhs, exp.clone(), exp_ref.span())
```
Leo-Besancon: If `k` is not constant, this function call will always return `Ok(None)`, so we should return it directly.

Leo-Besancon (changes requested):
Thank you for the PR, good catch! I've left a few comments inline.

Leo-Besancon:
Hi @Galoretka, just wondering if you planned on updating this PR or should I take over this fix? Thank you!

Galoretka:
> Hi @Galoretka, just wondering if you planned on updating this PR or should I take over this fix? Thank you!

Oh, I'm sorry, I will update it today, sorry for the delay

Galoretka:
> Hi @Galoretka, just wondering if you planned on updating this PR or should I take over this fix? Thank you!

Updated, sorry for the delay again

Leo-Besancon (approved):
LGTM!
I think you can just add a line in the `CHANGELOG.md` file
(e.g. `Fixed MIR's constant propagation to fold 0^0 to 1 (#509)` or similar) for the CI to pass and we can merge! Thank you for the changes.

Galoretka:
> LGTM! I think you can just add a line in the `CHANGELOG.md` file (e.g. `Fixed MIR's constant propagation to fold 0^0 to 1 (#509)` or similar) for the CI to pass and we can merge! Thank you for the changes.

Done

Issue #506, opened 2025-11-01
Avoid Matrix Construct in MDS Multiplication


Al-Kindi-0:
The RPO permutation round currently uses a 2D matrix construct for MDS  matrix multiplication.
 This can be simplified by defining individual row constants and computing explicit dot products, avoiding nested matrix iteration.

  ### Current Implementation
```
  const MDS = [
      [7, 23, 8, 26, 13, 10, 9, 7, 6, 22, 21, 8],
      [8, 7, 23, 8, 26, 13, 10, 9, 7, 6, 22, 21],
      // ... 10 more rows
  ];

  fn apply_mds(state: felt[12]) -> felt[12]{
      return [sum([s * m for (s, m) in (state, mds_row)]) for mds_row in MDS];
  }
```

  This approach uses:
  - A 2D matrix constant MDS
  - Nested list comprehension iterating over matrix rows
  - Matrix indexing to access individual rows

  ### Proposed Implementation

  Define MDS rows as individual constants and compute dot products explicitly:
```
  # MDS matrix rows used for computing the linear layer in a RPO round
  const MDSROWA = [7, 23, 8, 26, 13, 10, 9, 7, 6, 22, 21, 8];
  const MDSROWB = [8, 7, 23, 8, 26, 13, 10, 9, 7, 6, 22, 21];
  const MDSROWC = [21, 8, 7, 23, 8, 26, 13, 10, 9, 7, 6, 22];
  // ... 9 more rows (D through L)

  fn apply_mds(state: felt[12]) -> felt[12]{
      # Compute dot product of state vector with each MDS row
      let result0 = sum([s * m for (s, m) in (state, MDSROWA)]);
      let result1 = sum([s * m for (s, m) in (state, MDSROWB)]);
      let result2 = sum([s * m for (s, m) in (state, MDSROWC)]);
      // ... 9 more dot products

      return [result0, result1, result2, result3, result4, result5,
              result6, result7, result8, result9, result10, result11];
  }
```

#505 shows the viability of this change.

bobbinth:
Thank you for putting this together! Looking at the code, it is indeed pretty simple and I think it is safe to say that the extra compactness of the current approach does not justify the extra complexity introduced by the need to support matrixes.

So, if we go with the proposed implementation, I think we can probably:

- Remove matrix data types.
- Remove support for nested list comprehension (i'm assuming this is only useful in the context of matrixes, but please correct me if this is wrong).

[...]

PR #510, opened 2025-11-10
feat: replace nested Vec<Vec<u64>> with strided Matrix

### LINKED ISSUE #274: Replace `Vec<Vec<T>>` with a native Matrix representation
Matrixes are currently represented with:

https://github.com/0xPolygonMiden/air-script/blob/0ed25aad488140b1aecbc7434c634a87ac269012/air-script-core/src/constant.rs#L41

The representation above has two issues:

- there is one allocation per row
- there is no guarantee the matrix size is correctly, i.e. the following can be created `[[1], [2,3]]`

One sketch proposal is:

```rust
struct Matrix<T, const N: usize, const M: usize> {
    data: Vec<T>,
}

impl<T, const N: usize, const M: usize> Matrix<T, N, M> {
    fn new() -> Matrix<T, N, M> {
        let data = Vec::with_capacity(N * M);
        Matrix { data }
    }
}
```

huitseeker:
Replace the nested Vec<Vec<u64>> representation with a strided dense matrix:
- Single flat Vec<u64> for better memory efficiency and cache locality
- Enables double-indexing via ref-cast: matrix[row][col]
- Validation moved to constructor (parse time) instead of IR validation
- Added from_flat(), slice_rows(), and other utility methods

Updated usages across parser, MIR, and AIR:
- ConstantExpr::Matrix now uses new Matrix type
- ConstantValue::Matrix updated in MIR
- Removed old validation in module.rs (moved to Matrix::new)

Closes #274

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Fix regressions on MIR and list_comprehensions (#449).
 - Fixed a vector unrolling issue in nested match evaluations (#491).
 - Fix evaluator argument vector slice expansion (#495).
+- replace Vec<Vec<u64>> with dense strided Matrix representation (#510).
 
 ## 0.4.0 (2025-06-20)
 
--- a/air/src/tests/constant.rs
+++ b/air/src/tests/constant.rs
@@ -1,4 +1,4 @@
-use super::{compile_from_source, expect_diagnostic};
+use super::compile_from_source;
 
 #[test]
 fn boundary_constraint_with_constants() {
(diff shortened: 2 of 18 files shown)


bobbinth:
Without having looked at the code, is this related to the matrix data type in AirScript? If so, this is something we are considering removing in the near future: https://github.com/0xMiden/air-script/issues/506#issuecomment-3478633415.

huitseeker:
Closing to save on review bandwidth.

PR #511, opened 2025-11-10
feat(air): add memoization to RandomInputs::eval

Forostovec:
## Describe your changes

This change introduces an early return from RandomInputs::eval when a node’s evaluation is already present in evals_map, which prevents repeated traversal of already-computed subgraphs during common subexpression elimination; the behavior remains deterministic because leaf randoms are already cached, and the graph is a DAG, so caching internal node results is semantically safe while considerably improving performance in the CSE evaluation loop.


adr1anh:
Thanks, but we don't accept these small change PRs.

PR #476, opened 2025-09-26
ACE chiplet constraints

Al-Kindi-0:
## Checklist before requesting a review
- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](./CONTRIBUTING.md).
- Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.
- Updated `CHANGELOG.md`

PR #513, opened 2025-11-16
Fix cross-module function calls

Al-Kindi-0:
When functions in one module called functions from another module (e.g., `bitwise::compute_limb_and` calling `utils::binary_and`), these dependencies weren't tracked in the dependency graph.
  Root Cause

  Two issues were preventing proper handling of cross-module function calls:

  1. Missing dependency tracking: The semantic analysis only tracked root → function dependencies, not function → function dependencies across module boundaries.
  2. Incomplete accessor extraction: The MIR inlining pass's extract_accessor function didn't recursively handle Default accessors, causing failures when processing list comprehensions with function calls.

  The fix in this PR

  1. Semantic Analysis (parser/src/sema/semantic_analysis.rs)
  - Added current_function field to track which function/evaluator is being analyzed
  - Set this context when entering functions/evaluators and restore on exit
  - Added dependency edges in visit_mut_call to track function-to-function calls across modules

  2. Inlining Pass (mir/src/passes/inlining.rs)
  - Modified extract_accessor to recursively extract nested Default accessors instead of returning them directly

### DIFF
--- a/parser/src/sema/semantic_analysis.rs
+++ b/parser/src/sema/semantic_analysis.rs
@@ -76,6 +76,9 @@ pub struct SemanticAnalysis<'a> {
     has_undefined_variables: bool,
     has_type_errors: bool,
     in_constraint_comprehension: bool,
+    /// Tracks the qualified identifier of the function or evaluator currently being analyzed.
+    /// This is used to build the dependency graph for function-to-function calls.
+    current_function: Option<QualifiedIdentifier>,
 }
 impl<'a> SemanticAnalysis<'a> {
     /// Create a new instance of the semantic analyzer
@@ -103,6 +106,7 @@ impl<'a> SemanticAnalysis<'a> {
             has_undefined_variables: false,
             has_type_errors: false,
             in_constraint_comprehension: false,
+            current_function: None,
         }
     }
 
@@ -338,6 +342,14 @@ impl VisitMut<SemanticAnalysisError> for SemanticAnalysis<'_> {
         &mut self,
         function: &mut EvaluatorFunction,
     ) -> ControlFlow<SemanticAnalysisError> {
+        // Set the current function context for dependency tracking
+        let prev_function = self.current_function.clone();
+        let current_item = QualifiedIdentifier::new(
+            self.current_module.clone().unwrap(),
+            NamespacedIdentifier::Function(function.name),
+        );
+        self.current_function = Some(current_item.clone());
+
         // Only allow integrity constraints in this context
         self.constraint_mode = ConstraintMode::Integrity;
         // Start a new lexical scope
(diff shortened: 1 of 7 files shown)


huitseeker (commented):
`binary_and` is in #476 and #484, but `compute_limb_and` seems to have been deleted in #466, am I tracking where the examples are properly?

It would be great to add an air file in this PR that uses the x-module call syntax to ensure non-regression?

Al-Kindi-0:
> `binary_and` is in #476 and #484, but `compute_limb_and` seems to have been deleted in #466, am I tracking where the examples are properly?
> 

You are probably right, I think I might have messed up rebasing.

> It would be great to add an air file in this PR that uses the x-module call syntax to ensure non-regression?

Good point, will do

parser/src/sema/semantic_analysis.rs
```diff
@@ -338,6 +342,14 @@ impl VisitMut<SemanticAnalysisError> for SemanticAnalysis<'_> {
...
         function: &mut EvaluatorFunction,
     ) -> ControlFlow<SemanticAnalysisError> {
+        // Set the current function context for dependency tracking
+        let prev_function = self.current_function.clone();
+        let current_item = QualifiedIdentifier::new(
+            self.current_module.clone().unwrap(),
+            NamespacedIdentifier::Function(function.name),
+        );
+        self.current_function = Some(current_item.clone());
+
```
Soulthym, line 352: 346-351 should probably be using Option::replace instead:

```rust
let prev_function = self.current_function.replace(QualifiedIdentifier::new(
    self.current_module.clone().expect("current module should have been set"),
    NamespacedIdentifier::Function(function.name),
));
```

Also swapped the unwrap for an expect in my suggestion.

parser/src/sema/semantic_analysis.rs
```diff
@@ -398,6 +412,14 @@ impl VisitMut<SemanticAnalysisError> for SemanticAnalysis<'_> {
         &mut self,
         function: &mut Function,
     ) -> ControlFlow<SemanticAnalysisError> {
+        // Set the current function context for dependency tracking
+        let prev_function = self.current_function.clone();
+        let current_item = QualifiedIdentifier::new(
+            self.current_module.clone().unwrap(),
+            NamespacedIdentifier::Function(function.name),
+        );
+        self.current_function = Some(current_item.clone());
```
Soulthym, line 421: Same remark as for evaluators, Option::replace seems more appropriate.

Soulthym (commented):
Hey @Al-Kindi-0 

Overall the fix looks good to me. I left a few suggestions below to improve code clarity a little.

Al-Kindi-0:
Superseded by #524

PR #515, opened 2025-11-19
Use Plonky3 super trait

Leo-Besancon:
## Describe your changes

This is a draft PR for now to test out `MidenAir` and `MidenAirBuilder` super traits (from https://github.com/0xMiden/Plonky3/pull/12, targetting https://github.com/0xMiden/Plonky3/pull/9).

The aux trace is generated from the Air.


air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -300,25 +122,22 @@ pub(crate) fn check_constraints_with_airscript_traits<F, EF, A>(
...
     EF: ExtensionField<F>,
-    A: for<'a> Air<DebugConstraintBuilderWithAirScriptTraits<'a, F, EF>>,
-    A: for<'a> AirScriptAir<F, DebugConstraintBuilderWithAirScriptTraits<'a, F, EF>>,
+    A: MidenAir<F, EF>,
 {
     let height = main.height();
 
     let aux_bus_boundary_values: Vec<_> = (0..air.aux_width()).map(|_| EF::GENERATOR).collect();
-    let alpha_f: Vec<F> = (0..5).map(|i| F::from_u64(123456789 * i)).collect(); // Dummy alpha in F
+    let alpha_f: Vec<F> = (0..2).map(|i| F::from_u64(123456789 * i)).collect(); // Dummy alpha in F
```
Soulthym: Should we keep that dummy alpha here?

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -300,25 +122,22 @@ pub(crate) fn check_constraints_with_airscript_traits<F, EF, A>(
...
     let alpha = EF::from_basis_coefficients_iter(alpha_f.iter().cloned()).unwrap();
     let beta = EF::from_u64(987654321);
```
Soulthym, line 135: Same remark for beta's dummy value

air-script/src/tests/buses/buses_simple_plonky3.rs
```diff
@@ -12,41 +13,75 @@ pub const NUM_BETA_CHALLENGES: usize = 2;
...
-impl<F: Field, AB: AirScriptBuilder<F = F>> AirScriptAir<F, AB> for BusesAir {
     fn aux_width(&self) -> usize {
         AUX_WIDTH
     }
 
-    fn num_beta_challenges(&self) -> usize {
-        NUM_BETA_CHALLENGES
-    }
+    fn build_aux_trace(&self, _main: &RowMajorMatrix<F>, _challenges: &[EF]) -> Option<RowMajorMatrix<EF>> {
+        // Note: consider using Some(build_aux_trace_with_miden_vm::<F, EF>(_main, _challenges))
```
Soulthym: Is this comment still valid? Why dont we call to build_aux_trace_with_miden_vm?

air-script/src/tests/buses/buses_varlen_boundary_both_plonky3.rs
```diff
@@ -12,41 +13,75 @@ pub const NUM_BETA_CHALLENGES: usize = 2;
...
-impl<F: Field, AB: AirScriptBuilder<F = F>> AirScriptAir<F, AB> for BusesAir {
     fn aux_width(&self) -> usize {
         AUX_WIDTH
     }
 
-    fn num_beta_challenges(&self) -> usize {
-        NUM_BETA_CHALLENGES
-    }
+    fn build_aux_trace(&self, _main: &RowMajorMatrix<F>, _challenges: &[EF]) -> Option<RowMajorMatrix<EF>> {
+        // Note: consider using Some(build_aux_trace_with_miden_vm::<F, EF>(_main, _challenges))
```
Soulthym: Same remark, why dont we call the ready made helper?

air/src/passes/expand_buses.rs
```diff
@@ -139,6 +141,19 @@ impl<'a> BusOpExpand<'a> {
         };
         // Store the generated constraint
         ir.constraints.insert_constraint(TraceSegmentId::Aux, root, domain);
+
+        // Also store the initial value for auxiliary trace generation
+        if boundary == Boundary::First {
+            // TODO: May be invalid? For now, we put its value to zero
```
Soulthym, line 147: I dont think thats valid here. The bus checksum starts at 0 (or 1), but it doesnt mean 0 or 1 were inserted as they would get multiplied by the random values (alpha and betas) + an offset.

Soulthym (commented):
Hey!
I left a few comments from a first review pass

air-script/src/tests/buses/buses_simple_plonky3.rs
Leo-Besancon: `build_aux_trace_with_miden_vm` corresponds to the generation for the aux trace of specific Airs (for the MidenVM chiplets constraints), so we insert the comment everytime but using this helper is only valid for a specific Air.

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
```diff
@@ -0,0 +1,81 @@
...
+    F: Field + PrimeField64,
+    EF: ExtensionField<F>,
+{
+    const AUX_WIDTH: usize = 3;
+
+    // Create an AuxTraceBuilder
+    let kernel = Kernel::new(&[]).unwrap();
+    let ace_hints = AceHints::new(0, vec![]);
+    let final_transcript_state = PrecompileTranscriptState::default();
+    let aux_trace_builder = AuxTraceBuilder::new(kernel, ace_hints, final_transcript_state);
```
adr1anh: Nice!

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -56,8 +16,10 @@ pub struct DebugConstraintBuilderWithAirScriptTraits<'a, F: Field, EF: Extension
     row_index: usize,
     /// A view of the current and next main row as a vertical pair.
     main: VerticalPair<RowMajorMatrixView<'a, F>, RowMajorMatrixView<'a, F>>,
+    /// A view of the current and next preprocessed row as a vertical pair.
+    preprocessed: VerticalPair<RowMajorMatrixView<'a, F>, RowMajorMatrixView<'a, F>>,
```
adr1anh, line 20: We currently don't have, or plan to, support preprocessed traces, but we can leave it in and assume it's an empty matrix.

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -56,8 +16,10 @@ pub struct DebugConstraintBuilderWithAirScriptTraits<'a, F: Field, EF: Extension
...
     /// A view of the current and next aux row as a vertical pair.
-    aux: VerticalPair<RowMajorMatrixView<'a, EF>, RowMajorMatrixView<'a, EF>>,
+    aux: Option<VerticalPair<RowMajorMatrixView<'a, EF>, RowMajorMatrixView<'a, EF>>>,
```
adr1anh: Ideally, we remove the `Option` since we always have one, but this may interfere with P3. We should try to make P3 assume that there is always an aux trace.

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -116,35 +77,17 @@ where
...
-impl<'a, F, EF> ExtensionBuilder for DebugConstraintBuilderWithAirScriptTraits<'a, F, EF>
-where
-    F: Field,
-    EF: ExtensionField<F>,
-{
-    type EF = EF;
 
-    type ExprEF = EF;
+    fn periodic_evals(&self) -> &[<Self as MidenAirBuilder>::PeriodicVal] {
+        self.periodic_columns.as_slice()
```
adr1anh, line 85: Should this depend on the current row? I guess for debugging this is fine, but the values returned will depend on the row modulo the period.

air-script/src/test_utils/plonky3_traits.rs
adr1anh: Yeah it seems fine for now since this is for debugging.

air/src/passes/expand_buses.rs
Leo-Besancon: This comment is related to this: https://hackmd.io/@adr1anh/H1uC5ckJ-e#Bus-boundary-constraints-unified-bus
For now, AirScript allows first row bus boundaries to be constrained with variable length public inputs.
In that case, we don't have the initial value we should put in the aux trace first row at that point., so we put 0 instead (but it will build an aux trace that does not correspond to the constraint).

I think we should either:
- Disallow buses to start with variable length public inputs values (and we need to catch it here)
- Or feed the initial aux trace with the corresponding value, and we need to update this code.

air-script/src/test_utils/plonky3_traits.rs
```diff
@@ -329,39 +149,41 @@ pub(crate) fn check_constraints_with_airscript_traits<F, EF, A>(
...
             periodic_columns,
-            alpha,
-            beta,
-            beta_powers: beta_powers.clone(),
             permutation_randomness: permutation_randomness.clone(),
             aux_bus_boundary_values: aux_bus_boundary_values.clone(),
         };
 
-        AirScriptAir::eval(air, &mut builder);
+        if let Some(aux_trace) = &aux_trace {
```
adr1anh: Same comment as below about being able to assume we always have an auxiliary trace..

air/src/passes/expand_buses.rs
Soulthym: > I think we should either:
> 
> - Disallow buses to start with variable length public inputs values (and we need to catch it here)
> - Or feed the initial aux trace with the corresponding value, and we need to update this code.

I think the first option is acceptable for now, but we probably want the second one down the line

codegen/plonky3/src/air/graph.rs
```diff
@@ -78,7 +82,15 @@ impl Codegen for Value {
             Value::PeriodicColumn(pc) => {
                 let index =
                     ir.periodic_columns.iter().position(|(qid, _)| qid == &pc.name).unwrap();
-                format!("periodic_values[{index}].into()")
+                match elem_type {
```
adr1anh, line 85: These are always EF elements (not expressions), no?

adr1anh (commented):
I realize it's still in flux, but it's looking great already. Left a few comments in the mean time

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
```diff
@@ -0,0 +1,81 @@
...
+    // Convert main trace to Miden format
+    let mut main_trace_vec_vec = Vec::new();
+    for row_index in 0..main.height() {
+        let row_f: Vec<F> = main.row_slice(row_index).unwrap().to_vec();
+        let row_felt =
+            row_f.into_iter().map(|x| Felt::new(x.as_canonical_u64())).collect::<Vec<_>>();
+        main_trace_vec_vec.push(row_felt);
+    }
+    let transposed_main_trace_vec_vec: Vec<Vec<_>> = (0..main_trace_vec_vec[0].len())
+        .map(|i| main_trace_vec_vec.iter().map(|row| row[i].clone()).collect())
```
Soulthym: nit: the .clone() isnt necessary here

codegen/plonky3/src/air/graph.rs
Leo-Besancon: If I recall correctly, the issue was that for instance, `AB::F::from_u64(12) * main_current[7].clone().into()`
This fails as the type for main_current elements, `AB::Var`, cannot be multiplied with AB::F.

By digging a bit more if we inverse the order of the elements it works (the op traits are not commutative.

The only way I found to satisfy the bounds was to convert everything to Expr or ExprEF. But maybe we could restrict the types to allow more things.

air-script/src/test_utils/plonky3_traits.rs
Leo-Besancon: Removed the Option

air-script/src/test_utils/plonky3_traits.rs
Leo-Besancon: The modulo is done when building the DebugConstraintBuilderWithAirScriptTraits struct

air-script/src/test_utils/plonky3_traits.rs
Leo-Besancon: Removed the Option

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
```diff
@@ -0,0 +1,81 @@
...
+    const AUX_WIDTH: usize = 3;
+
+    // Create an AuxTraceBuilder
+    let kernel = Kernel::new(&[]).unwrap();
+    let ace_hints = AceHints::new(0, vec![]);
+    let final_transcript_state = PrecompileTranscriptState::default();
+    let aux_trace_builder = AuxTraceBuilder::new(kernel, ace_hints, final_transcript_state);
+
+    // Convert main trace to Miden format
+    let mut main_trace_vec_vec = Vec::new();
```
adr1anh: This whole block is not performance sensitive right? Plonky3 has a transpose primitive optimized for different architectures we could use if we need.

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
```diff
@@ -0,0 +1,81 @@
...
+    let trace_length = num_rows * AUX_WIDTH;
+    let long_trace = EF::zero_vec(trace_length);
+    let mut aux_trace = RowMajorMatrix::new(long_trace, AUX_WIDTH);
+    let (prefix, rows, suffix) = unsafe { aux_trace.values.align_to_mut::<[EF; AUX_WIDTH]>() };
+    assert!(prefix.is_empty(), "Alignment should match");
+    assert!(suffix.is_empty(), "Alignment should match");
+    assert_eq!(rows.len(), num_rows);
+
+    for j in 0..AUX_WIDTH {
+        let col = aux_trace_miden.get(j).unwrap();
```
adr1anh, line 102: If we create a vector of concatenated columns, we should be able to use some methods from plonky3 like [`flatten_to_base`](https://github.com/Plonky3/Plonky3/blob/main/util/src/lib.rs) and `RowMajorMatrix::transpose` to simplify the code below. Mainly marking it for info, as it seems this is used for tests.

air-script/src/tests/binary/binary_plonky3.rs
```diff
@@ -12,34 +13,20 @@ pub const NUM_BETA_CHALLENGES: usize = 0;
 
 pub struct BinaryAir;
 
-impl<F> BaseAir<F> for BinaryAir {
+impl<F, EF> MidenAir<F, EF> for BinaryAir
+where F: Field,
+      EF: ExtensionField<F>,
```
adr1anh: We don't need this trait bound IIRC

air/src/ir/mod.rs
```diff
@@ -195,6 +195,12 @@ pub struct Air {
     ///
     /// Only their name, type, and the first and last boundary constraints are stored here.
     pub buses: BTreeMap<Identifier, Bus>,
+    /// Buses initial values used for auxiliary trace generation (indexed by bus index)
+    pub buses_initial_values: BTreeMap<usize, NodeIndex>,
+    /// Buses transition expressions used for auxiliary trace generation (indexed by bus index)
+    /// The tuple contains the numerator and an optional denominator operation (p_prime = numerator
+    /// if None or numerator / denominator)
+    pub buses_transitions: BTreeMap<usize, (NodeIndex, Option<NodeIndex>)>,
```
adr1anh, line 209: This only works for multi-set rights? Maybe a simple math-y comment might be helpful here.

codegen/plonky3/src/air/graph.rs
adr1anh: Converting everything to Expr should be fine. In cases where it matters, they are the same type so the compiler should inline. But having the extra variant here doesn't seem to be a problem, though we just want to avoid converting base elements to extension field elements unnecessarily.

codegen/plonky3/src/air/mod.rs
```diff
@@ -61,52 +62,40 @@ fn add_constants(scope: &mut Scope, ir: &Air) {
...
-    // add the custom BaseAir implementation block
-    let base_air_impl = scope.new_impl(name).generic("F").impl_trait("BaseAir<F>");
-    base_air_impl.new_fn("width").arg_ref_self().ret("usize").line("MAIN_WIDTH");
+    // add the custom MidenAir implementation block
+    let miden_air_impl = scope
+        .new_impl(name)
+        .generic("F")
+        .generic("EF")
+        .bound("F", "Field")
+        .bound("EF", "ExtensionField<F>")
```
adr1anh: Maybe we don't need this.

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
```diff
@@ -0,0 +1,81 @@
+use miden_air::{Felt, FieldElement, trace::main_trace::MainTrace};
+use miden_processor::{
+    AceHints, AuxTraceBuilder, ColMatrix, Kernel, PrecompileTranscriptState, QuadExtension,
+};
+use p3_field::{ExtensionField, Field, PrimeField64};
+use p3_matrix::Matrix;
+use p3_miden_air::RowMajorMatrix;
+
+/// Builds the Miden VM auxiliary trace using the provided main trace and challenges.
+pub fn build_aux_trace_with_miden_vm<F, EF>(
```
Al-Kindi-0, line 58: Just trying to better understand the plan with this function. I am assuming that once Miden VM will migrate to using Plonky3 as its backend, [this](https://github.com/0xMiden/miden-vm/blob/926a0c6b714cc69a52596aefad3c77b294750ddc/processor/src/trace/mod.rs#L267) will be doing what `build_aux_trace_with_miden_vm` is doing now, right?

Al-Kindi-0: But maybe this is just for tests and then my comment is void?

air-script/src/tests/buses/buses_complex_plonky3.rs
```diff
@@ -12,41 +13,75 @@ pub const NUM_BETA_CHALLENGES: usize = 3;
...
+                &main,
+                _challenges,
+                &periodic_values,
+                prev_row,
+            );
+            for j in 0..AUX_WIDTH {
+                rows[i+1][j] = next_row[j];
+            }
+        }
+        Some(trace)
```
Al-Kindi-0, line 69: Seems to me that `build_aux_trace` is pretty much the same for all examples, and the same goes for `buses_transitions` with the exception of maybe the constants.
Should we refactor this to remove code duplication? Maybe even just a macro?

air/src/ir/mod.rs
Al-Kindi-0: We should also probably create an issue to track adding support for LogUp based buses

air/src/passes/expand_buses.rs
Al-Kindi-0: I wouldn't spend to much time on this but we should create an issue to track it

Al-Kindi-0 (commented):
Looks good!
Did an initial review and will hopefully do a more thorough one later on

air/src/passes/expand_buses.rs
Soulthym: https://github.com/0xMiden/air-script/issues/518

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
Soulthym: I think thats just for tests but Im not certain, Ill check with Leo

air/src/ir/mod.rs
Leo-Besancon: It is implemented for both multiset and logup buses (see here in expand_buses for the logup implementation: https://github.com/0xMiden/air-script/pull/515/files#diff-4c56c872b147b67a3091ba3cca1d2de8b66d8bcd39219e2fa83271f04142c32eR402). The E2E bus complex test has both bus types and the aux trace generation works well (the aux constraints validate against them) so I think the math works out, for this simple example at least.

The only potential limitation is that for logup we need to add new nodes in the algebraic graph to represent the numerator / denominators (as the formula does not give it directly), so it potentially as a (very small IMO) impact on the CSE pass: it should always work, but the resulting algebraic graph may be less optimized than if we did not add these nodes).

I've added a comment with the math in dc8043fdda4ee85885d4d6a6c4d6d4b9ff1bf39d

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
Leo-Besancon: > Just trying to better understand the plan with this function. I am assuming that once Miden VM will migrate to using Plonky3 as its backend, [this](https://github.com/0xMiden/miden-vm/blob/926a0c6b714cc69a52596aefad3c77b294750ddc/processor/src/trace/mod.rs#L267) will be doing what `build_aux_trace_with_miden_vm` is doing now, right?

The goal of this function is to be able to delegate the aux trace generation by using the miden vm code.
Then, we will be able to compare that the aux trace generation made with AirScript (using the default implementation that calls the buses_transitions function) yields the same values as the one in `build_aux_trace_with_miden_vm` that calls the miden vm code. Then we can choose which one we use depending on our needs.

For now, the `build_aux_trace_with_miden_vm` function only handles the aux trace for chiplets, I plan to add the other MidenVM aux traces today. I planned to have a way to select which aux trace to generate, but seeing the snippet you linked, maybe it would be easier to do that same and handle all the modules at once. What do you think, is it useful to build the Air for each module individually and validate e.g. the chiplet's aux trace independantly, or would we build all the modules at once?

air-script/src/tests/buses/buses_complex_plonky3.rs
Leo-Besancon: I'm not sure it is possible to prevent code duplication here in the general case, as the generated file should not depend on imports from other files of the crate (otherwise, we generate code that cannot be used outside the airscript repo).

To avoid too much code duplication for tests we maybe could do in codegen:
```rs
#[cfg(test)]
{
  let trace = crate::test_utils::build_aux_trace_from_airscript::<F, EF>(_main, _challenges, AUX_WIDTH, &periodic_table, initial_values, BusesAir::buses_transitions::<F, EF>);
}

#[cfg(not(test))]
{
  // Current code
}
```

This way, the generated code only import a helper function in tests. What do you think?

codegen/plonky3/src/air/mod.rs
Leo-Besancon: It is just needed when there are either buses or periodic columns, I filter it otherwise in the latest commit.

adr1anh (approved):
This looks great!

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
Al-Kindi-0: Makes sense! Should we create an issue to track the future TODOs? Also, can you resolve the comments raised by Adrian on this function a few lines below?

air/src/ir/mod.rs
Al-Kindi-0: Thank you!
Would it make sense to create an issue to investigate the impact on the CSE pass in the future?

air/src/passes/expand_buses.rs
Al-Kindi-0: Thanks

Al-Kindi-0 (approved):
LGTM!
Left some comments and question about adding some issues to track some points.

air/src/ir/mod.rs
Leo-Besancon: Sure! The issue for the impact on the CSE pass: https://github.com/0xMiden/air-script/issues/521
Also on the same subject, the issue for optimizing the logup bus aux trace generation (for now they do not use batch inverse computation): https://github.com/0xMiden/air-script/issues/519

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
Leo-Besancon: done in the latest commit

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
Leo-Besancon: I did not find a clean way to do it (unless `flatten_to_base` could be used on the winterfell types? I have not checked). Keeping it as is for now, but I've mentionned this potential optimization in https://github.com/0xMiden/air-script/issues/522

air-script/src/test_utils/miden_vm_aux_trace_generator.rs
Leo-Besancon: Sure, here is the issue: https://github.com/0xMiden/air-script/issues/522
Handled the first comment directly in the latest commit, kept the other one as a future TODO in the issue.

PR #517, opened 2025-11-24
fix: Import::Partial equality to be symmetric

MozirDmitriy:
## Describe your changes

The previous PartialEq implementation for Import::Partial used a subset check, which is asymmetric and violates the Eq contract, since Eq requires symmetry. This could lead to incorrect behavior when comparing Module.imports and during AST equality checks in tests. The implementation now requires exact equality of both the module and the item set (l == r && ls == rs), relying on HashSet’s correct Eq semantics. Added unit tests to assert that {a} and {a,b} are not equal in either direction, and that identical sets compare equal.

### DIFF
--- a/parser/src/ast/declarations.rs
+++ b/parser/src/ast/declarations.rs
@@ -236,10 +236,8 @@ impl PartialEq for Import {
     fn eq(&self, other: &Self) -> bool {
         match (self, other) {
             (Self::All { module: l }, Self::All { module: r }) => l == r,
-            (Self::Partial { module: l, items: ls }, Self::Partial { module: r, items: rs })
-                if l == r =>
-            {
-                ls.difference(rs).next().is_none()
+            (Self::Partial { module: l, items: ls }, Self::Partial { module: r, items: rs }) => {
+                l == r && ls == rs
             },
             _ => false,
         }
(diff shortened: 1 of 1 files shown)


parser/src/ast/declarations.rs
```diff
@@ -438,3 +436,54 @@ impl PartialEq for Function {
...
+    fn import_partial_subset_is_not_equal_either_direction() {
+        let module = module_id(&["m"]);
+        let mut set_a: HashSet<Identifier> = HashSet::default();
+        set_a.insert(ident("a"));
+        let import_a = Import::Partial { module: module.clone(), items: set_a };
+
+        let mut set_ab: HashSet<Identifier> = HashSet::default();
+        set_ab.insert(ident("a"));
+        set_ab.insert(ident("b"));
+        let import_ab = Import::Partial { module: module.clone(), items: set_ab };
```
Soulthym: nit: we can probably remove this module.clone()

parser/src/ast/declarations.rs
```diff
@@ -438,3 +436,54 @@ impl PartialEq for Function {
...
+        let module = module_id(&["m"]);
+        let mut set1: HashSet<Identifier> = HashSet::default();
+        set1.insert(ident("a"));
+        set1.insert(ident("b"));
+        let mut set2: HashSet<Identifier> = HashSet::default();
+        set2.insert(ident("b"));
+        set2.insert(ident("a"));
+
+        let import1 = Import::Partial { module: module.clone(), items: set1 };
+        let import2 = Import::Partial { module: module.clone(), items: set2 };
```
Soulthym: likewise for this .clone()

Soulthym (commented):
Hey, thanks for the PR!

The code looks good and should indeed fix issues should we rely on Eq's symmetry.

I left a minor comment regarding .clone(), otherwise seems ready to merge.

parser/src/ast/declarations.rs
MozirDmitriy: > likewise for this .clone()

Did it

Soulthym (approved):
Thank you @MozirDmitriy, I think this is ready to merge @bobbinth

PR #520, opened 2025-12-03
fix: pin mdbook version to 0.4.48

Leo-Besancon:
Non-updated mdbook katex plugin caused issue with the CI in latest PRs.


bobbinth (approved):
Looks good! Thank you!

Issue #512, opened 2025-11-12
[Plonky3] Use `prove` / `verify` in E2E tests for full workflow


Leo-Besancon:
Following up on https://github.com/0xMiden/air-script/pull/508 for the Plonky3 Codegen, we currently only implement the `AirScriptAir` and `AirScriptBuilder` traits for the debug structure asserting the constraints.
As a result, in tests, we can only call `check_constraints_with_airscript_traits` to assert that the trace validates the given constraints, but not prove or verify a given proof.

Once all the structs in https://github.com/0xMiden/Plonky3 support `AirScriptAir` and `AirScriptBuilder`, we should add back the code that proves / verifies a trace, e.g.:

```rust
let proof = prove(&config, &BitwiseAir {}, trace, &inputs_goldilocks);
verify(&config, &BitwiseAir {}, &proof, &inputs_goldilocks).expect("Verification failed");
```

PR #19, opened 2022-10-23
Add tests for codegen

tohrnii:
WIP
Addressing #9.

PR #523, opened 2025-12-05
feat: prove/verify and sync plonky3

Leo-Besancon:
## Describe your changes

This PR, based on #461, aims to:
- Resolve https://github.com/0xMiden/air-script/issues/512
- Sync with the latest Plonky3 fork work (containing the prover, handling of periodic columns, etc.)

It may be used in a full E2E workflow.


air-script/src/test_utils/air_tester_macros.rs
```diff
@@ -62,7 +62,8 @@ macro_rules! generate_air_plonky3_test_with_airscript_traits {
             let challenge_mmcs = ChallengeMmcs::new(val_mmcs.clone());
             let challenger = Challenger::from_hasher(vec![], byte_hash);
             let dft = Dft::default();
-            let fri_params = p3_fri::create_benchmark_fri_params(challenge_mmcs);
+            let mut fri_params = p3_fri::create_benchmark_fri_params(challenge_mmcs);
+            fri_params.log_blowup = 4; // Use a higher blowup than default for degree 9 constraints
```
Al-Kindi-0: Should we fix this now or is there something blocking it?

Al-Kindi-0 (approved):
LGTM

Al-Kindi-0:
@Leo-Besancon , should we create an issue in the P3 repo regarding missing support for aux boundary constraints?

air-script/src/test_utils/air_tester_macros.rs
Leo-Besancon: > Should we fix this now or is there something blocking it?

I don't think anything is blocking it. There are multiple functions in p3_fri to create params (`create_test_fri_params`, `create_test_fri_params_zk`, `create_benchmark_fri_params`..). Maybe we could add a `create_miden_fri_params` with the values we want if we know them all.

Here it generates:
```rust
    FriParameters {
        log_blowup: 4,
        log_final_poly_len: 0,
        num_queries: 100,
        proof_of_work_bits: 16,
        mmcs,
        log_folding_factor: 1,
    }
```

Al-Kindi-0: Makes sense, and also a configuration with blow up factor 8 for Miden VM constraints.

Leo-Besancon:
> @Leo-Besancon , should we create an issue in the P3 repo regarding missing support for aux boundary constraints?

I've added the issue there: https://github.com/0xMiden/Plonky3/issues/17, would be great if you (or @adr1anh) could check it and add comments if needed!

air-script/src/test_utils/air_tester_macros.rs
Leo-Besancon: > Makes sense, and also a configuration with blow up factor 8 for Miden VM constraints.

Did you mean like so? https://github.com/0xMiden/Plonky3/pull/19

adr1anh: Do we need `log_blowup=4`? I thought we only went up until 3 for our constraints, but maybe we're doing 4 for smaller proofs?

Al-Kindi-0: Agreed, `log_blowup=3` should be enough for us in this context

Leo-Besancon: I've updated #19 to only have a log_blow_up of 3 and one set of parameters for miden, when targetting it here I have some issues but I think just increasing the number of rows for the proof will be enough, I'll test it.

Leo-Besancon: Also will reduce the constraints degrees for a few of the tests.

Al-Kindi-0 (approved):
Looks great!

PR #514, opened 2025-11-17
Allow pure functions import

Al-Kindi-0:
## Describe your changes
Extends the import system to support pure functions, enabling them to be imported across hierarchical modules alongside existing constant and evaluator imports.

## Checklist before requesting a review
- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](./CONTRIBUTING.md).
- Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.
- Updated `CHANGELOG.md`

PR #524, opened 2025-12-12
 fix inlining loop on cross imports

Soulthym:
## Describe your changes
Merges several pending PRs #513 #507 #514  into the branch al-constraints-air-only
Fixes an issue when inlining nested function calls accross modules.

- [x] Repo forked and branch created from `next` according to naming convention.
created from `al-constraints-air-only`
- [x] Commit messages and codestyle follow [conventions](./CONTRIBUTING.md).
- [x] Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- [x] Relevant issues are linked in the PR description.
- [x] Tests added for new functionality.
- [x] Documentation/comments updated according to changes.
- [x] Updated `CHANGELOG.md`

### DIFF
--- a/mir/src/passes/inlining.rs
+++ b/mir/src/passes/inlining.rs
@@ -532,7 +532,10 @@ fn extract_accessor(op: Link<Op>) -> Link<Op> {
     };
     let mut indexable = accessor.indexable.clone();
     match &accessor.access_type {
-        MirAccessType::Default => accessor.indexable.clone(),
+        MirAccessType::Default => {
+            // Recursively extract accessors from the indexable
+            extract_accessor(accessor.indexable.clone())
+        },
         MirAccessType::Index(idx_op) => {
             let idx = get_inner_const(idx_op)
                 .unwrap_or_else(|| panic!("expected constant index, got {:#?}", idx_op))
--- a/mir/src/passes/inlining.rs
+++ b/mir/src/passes/inlining.rs
@@ -476,7 +476,6 @@ impl Visitor for InliningSecondPass<'_> {
                 self.visit_call(graph, call_op.clone())?;
                 // NOTE: Once the call has been processed, we return early to avoid
                 // an infinite loop.
-                // modules.
                 // If there are nested calls, they will be handled via fixed-point
                 // compilation (see [<Inlining<'_> as Pass>::run]).
                 return Ok(());
--- a/mir/src/passes/inlining.rs
+++ b/mir/src/passes/inlining.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, ops::Deref};
+use std::{
+    collections::HashMap,
+    ops::Deref,
+};
 
 use air_pass::Pass;
 use miden_diagnostics::{DiagnosticsHandler, Severity, SourceSpan, Spanned};
@@ -141,6 +144,7 @@ pub struct InliningSecondPass<'a> {
     // HashMap<CaleePtr, (Callee, Vec<Call nodes where called>)>
     func_eval_nodes_where_called: HashMap<usize, (Link<Root>, Vec<Link<Op>>)>, // Op is a Call here
     had_calls: bool,
+    seen_root_call: bool,
 }
 
 impl<'a> InliningSecondPass<'a> {
@@ -158,6 +162,7 @@ impl<'a> InliningSecondPass<'a> {
             func_eval_nodes_where_called,
             func_eval_inlining_order,
             had_calls: false,
+            seen_root_call: false,
         }
     }
 }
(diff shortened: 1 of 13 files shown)


mir/src/passes/inlining.rs
```diff
@@ -470,54 +472,57 @@ impl Visitor for InliningSecondPass<'_> {
             // First, check if it's a known `Call` to inline,
             // if so, set the context and scan its body
```
Al-Kindi-0, line 473: Should we update the comment and explain the early return ?

mir/src/passes/inlining.rs
```diff
@@ -470,54 +472,57 @@ impl Visitor for InliningSecondPass<'_> {
...
-                    );
+                if !self.seen_root_call {
+                    self.seen_root_call = true;
+                    self.visit_call(graph, call_op.clone())?;
+                    return Ok(());
                 }
             }
+
+            // Else, we are currently visiting the body of a function or an evaluator of a call
+            // we want to inline We use our helper duplicate_node_or_replace to
```
Al-Kindi-0: typo: "..inline. We..."

parser/src/sema/semantic_analysis.rs
```diff
@@ -695,10 +719,21 @@ impl VisitMut<SemanticAnalysisError> for SemanticAnalysis<'_> {
                             FunctionType::Evaluator(_) => DependencyType::Evaluator,
                             _ => DependencyType::Function,
                         };
+
+                        // If we're currently analyzing a function, add a direct dependency edge
+                        // from the current function to the called function. This ensures that
+                        // transitive dependencies across module boundaries are properly tracked.
+                        if let Some(caller) = self.current_function.clone() {
```
Al-Kindi-0, line 726: Could we remove the clone ?

Al-Kindi-0 (approved):
Looks great, thank you!
Could we add tests that cover the fixes implemented in this PR? Otherwise, this looks ready to merge.

parser/src/sema/semantic_analysis.rs
Soulthym: It seems not, we cannot copy because the ModuleId is now a Vec due to #507.

And due to https://github.com/0xMiden/air-script/blob/efcd89196e97cf3b64dd0284ec3d43637348c64f/parser/src/sema/semantic_analysis.rs#L727 with the signature:
```rust
get_node_index_or_add(&mut self, ...)
```
we otherwise end up having an immutable reference to self, that we need to borrow mutably at the same time.

mir/src/passes/inlining.rs
Soulthym: Addressed in efc46d5

mir/src/passes/inlining.rs
Soulthym: Fixed in efc46d5

Soulthym:
Thank you @Al-Kindi-0!

> Could we add tests that cover the fixes implemented in this PR?

Working on it, I should have that in about an hour. Other comments have been addressed!

Soulthym:
Everything should have been addressed now!

I combined entries #507 #513 and #514 in the changelog since they are fixes on 507.

The added test is under mir/tests/ir, Ive tested it on earlier versions and it does indeed timeout, this should cover regressions.

As for the failing docs, I believe the fix is part of another PR so I didnt want to introduce an unnecessary conflict here.

mir/src/passes/inlining.rs
```diff
@@ -469,54 +471,60 @@ impl Visitor for InliningSecondPass<'_> {
 
             // First, check if it's a known `Call` to inline,
             // if so, set the context and scan its body
-            if call_op.clone().as_call().is_some() {
+            if call_op.clone().as_call().is_some() && !self.seen_root_call {
+                self.seen_root_call = true;
                 self.visit_call(graph, call_op.clone())?;
+                // NOTE: Once the call has been processed, we return early to avoid
+                // an infinite loop.
+                // modules.
```
Al-Kindi-0: This should go as well, right?

Al-Kindi-0 (approved):
Looks great, thanks @Soulthym !
Left one minor comment.
Also I am supposing that you tested the failure of the test without the fix.

mir/src/passes/inlining.rs
Al-Kindi-0: I meant only line 479

Soulthym: This was added to comment the early return as you requested here: https://github.com/0xMiden/air-script/pull/524#discussion_r2631501207

Soulthym:
> Also I am supposing that you tested the failure of the test without the fix.

From my previous message, yes tested on 4354d71efe's parent:
> The added test is under mir/tests/ir, Ive tested it on earlier versions and it does indeed timeout, this should cover regressions.

mir/src/passes/inlining.rs
Soulthym: Ah yes ok mb, removing it

Soulthym: Addressed on fabad207

Soulthym:
Rebased on next and removed constraints related commits from this PR.

Should be ready to merge @Al-Kindi-0 !

Al-Kindi-0:
Thank you @Soulthym !

PR #526, opened 2025-12-22
Fix cross module constants and comprehension scoping

Al-Kindi-0:
## Describe your changes

  This PR fixes two bugs related to cross-module imports and comprehension scoping:

  1. Cross-module constant dependencies:
   Constants used in comprehension iterables fail when imported across module boundaries.

  2. Comprehension periodic binding scoping:
   Iterating over a vector containing periodic column references causes the binding to be mistyped as PeriodicColumn, leading to "no entry found for key" panic.

- Updated `CHANGELOG.md`


parser/src/sema/semantic_analysis.rs
```diff
@@ -444,9 +444,9 @@ impl VisitMut<SemanticAnalysisError> for SemanticAnalysis<'_> {
             self.current_module.clone().unwrap(),
             NamespacedIdentifier::Function(function.name),
         );
-        let current_item_node_index = self.deps_graph.add_node(current_item);
-        for (referenced_item, ref_type) in self.referenced.iter() {
-            let referenced_item_node_index = self.deps_graph.add_node(referenced_item.clone());
+        let current_item_node_index = self.get_node_index_or_add(&current_item);
+        for (referenced_item, ref_type) in self.referenced.clone().iter() {
+            let referenced_item_node_index = self.get_node_index_or_add(referenced_item);
             self.deps_graph.add_edge(
```
Soulthym, line 450: I'm assuming this is the part of the issue you encountered, we were overriding values without checking for their existence first?

This edit looks correct, just asking to confirm I understand the problem properly

parser/src/ast/expression.rs
```diff
@@ -865,7 +865,7 @@ pub enum AccessType {
     #[default]
     Default,
     /// Access binds a sub-slice of a vector
-    Slice(RangeExpr),
+    Slice(Box<RangeExpr>),
```
Soulthym, line 868: Is there a reason to switching to a Box here instead of an owned RangeExpr?

Soulthym (commented):
Hey @Al-Kindi-0, I left a few comments below about clarifications and a minor nit regarding Boxing RangeExpr. The code itself looks good to me otherwise, great job!

parser/src/ast/expression.rs
Al-Kindi-0: `Slice(Box<RangeExpr>)` is there mainly to avoid the `clippy::large_enum_variant` warning.

parser/src/sema/semantic_analysis.rs
Al-Kindi-0: Indeed, we were calling `add_node` unconditionally, which could create duplicate nodes. `get_node_index_or_add` reuses the existing node index.

Leo-Besancon (approved):
Nice catches, thank you!
LGTM, I was wondering if the fix for the bindings from PeriodicColumn to local variable could have any impact but I don't see how, and if the tests don't catch regressions it's probably good!

PR #531, opened 2026-01-05
fix(mir): make BackLink parent removal precise

phrwlk:
## Describe your changes

BackLink implements PartialEq as always true so that backlink fields are ignored in derived equality and hashing. However, 
remove_parent in MIR ops compared BackLink<Owner> values directly:
```rust
self.parents.retain(|p| *p != parent.clone().into());*
```

Because of the always-true PartialEq, this predicate was always false and effectively cleared the entire parents vector instead of removing only the requested parent.

This change updates all remove_parent implementations for ops that use parents: Vec<BackLink<Owner>> to compare against the upgraded Link<Owner> via to_link(), keeping BackLink’s structural Eq/Hash semantics intact while making parent removal semantically correct.

### DIFF
--- a/mir/src/ir/nodes/ops/accessor.rs
+++ b/mir/src/ir/nodes/ops/accessor.rs
@@ -70,6 +70,9 @@ impl Child for Accessor {
         self.parents.push(parent.into());
     }
     fn remove_parent(&mut self, parent: Link<Self::Parent>) {
-        self.parents.retain(|p| *p != parent.clone().into());
+        self.parents.retain(|p| match p.to_link() {
+            Some(link) => link != parent,
+            None => true,
+        });
     }
 }
--- a/mir/src/ir/nodes/ops/add.rs
+++ b/mir/src/ir/nodes/ops/add.rs
@@ -33,10 +33,51 @@ impl Child for Add {
     fn get_parents(&self) -> Vec<BackLink<Self::Parent>> {
         self.parents.clone()
     }
+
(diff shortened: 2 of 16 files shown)


AlgofootPrint:
Review: Fix is incomplete. Link::PartialEq compares by value, not pointer identity @phrwlk 

  I tested this PR locally and the included test add_remove_parent_removes_only_target fails:

 thread 'ir::nodes::ops::add::tests::add_remove_parent_removes_only_target' panicked
  assertion `left == right` failed
  left: 0
  right: 1

  The root cause diagnosis is correct — BackLink::PartialEq always returning true breaks retain. However, the proposed
  fix still has a problem.

  Switching to p.to_link() and then link != parent uses Link<T>::PartialEq, which compares by value (self.link ==
  other.link → Rc<RefCell<T>> → inner value equality). So if two different owners happen to hold the same value (as in
  the test: both are Owner::None(span)), they compare as equal and both get removed, leaving parents.len() == 0.

 The fix needs pointer identity, using the already-available get_ptr() method on Link<T>:

 fn remove_parent(&mut self, parent: Link<Self::Parent>) {
       self.parents.retain(|p| match p.to_link() {
           Some(link) => link.get_ptr() != parent.get_ptr(),
           None => true,
       });
   }

 This uses Rc::as_ptr under the hood, which correctly distinguishes different allocations regardless of value
 equality. The same fix needs to be applied to all 16 op files.

 The test itself is well-structured and proves the regression correctly but it just needs the implementation to match.

PR #539, opened 2026-02-16
Tagging infrastructure 

Al-Kindi-0:
## Describe your changes

This PR introduces the tagging infrastructure discussed offline for matching Miden VM constraints implemented in `miden-vm` using `AirBuilder` pattern and the ones in `air-script`.

In addition to expanding the syntax with the `tag` keywords and adding needed plumbing through parsing, semantic analysis, and MIR for it, we needed to add a linear (sum‑form) bus expansion path for multiset reductions in the case of parallel interactions gated by mutually exclusive flags. 
Currently, we need to annotate the bus to choose the expansion but I think that it shouldn't be difficult to automate this, or at least add some checks to make sure linear expansion is possible.

- Updated `CHANGELOG.md`


CHANGELOG.md
```diff
@@ -13,10 +13,10 @@
 - Fix regressions on MIR and list_comprehensions (#449).
 - Fixed a vector unrolling issue in nested match evaluations (#491).
 - Fix evaluator argument vector slice expansion (#495).
-- Fixed MIR's constant propagation to fold 0^0 to 1 (#509)
+- Fixed MIR's constant propagation to fold 0^0 to 1 (#509).
 - Support importing hierarchical modules (#507, #513, #514).
 - Fix MIR inlining loop on deeply nested calls (#524).
-- Fix cross-module constant resolution and comprehension scoping (#526).
```
Leo-Besancon, line 19: Isn't this line still relevant?

parser/src/parser/grammar.lalrpop
```diff
@@ -351,6 +416,31 @@ ConstraintStatements: Vec<Statement> = {
 }
 
 ConstraintStatement: Vec<Statement> = {
+    <l:@L> <tag:TagAttr> "enf" <expr: ScalarConstraintExpr> <comprehension: ConstraintComprehension<ScalarExpr>?> <selector: WithSelector?> ";" <r:@R> => {
```
Leo-Besancon, line 446: Is it not possible to unify this branch and the third branch with an optional tag instead to avoid code duplication?

Leo-Besancon (commented):
Left two nits, otherwise it looks good to me, but I did not go other the changes in the bus codegen constraints.
Also, maybe we want an additionnal `air-script/tests/buses` test for `@sum_form` tag behaviour?

CHANGELOG.md
Al-Kindi-0: Lost when changing the base.
Restored now

parser/src/parser/grammar.lalrpop
Al-Kindi-0: Done

air/src/passes/expand_buses.rs
```diff
@@ -138,17 +148,45 @@ impl<'a> BusOpExpand<'a> {
             Boundary::Last => ConstraintDomain::LastRow,
         };
         // Store the generated constraint
-        ir.constraints.insert_constraint(TraceSegmentId::Aux, root, domain);
+        ir.constraints.insert_constraint(TraceSegmentId::Aux, root, domain, tag);
+
+        // Also store the initial value for auxiliary trace generation
+        if boundary == Boundary::First {
+            // TODO: May be invalid? For now, we put its value to zero
```
huitseeker: That seems like it can desync the aux trace from the boundary constraint when `p.first` is a public input table. Would it make sense to store the same `PublicInputTable`-derived value (or computed reduction) here instead of zero?

air/src/passes/expand_buses.rs
```diff
@@ -234,8 +272,103 @@ impl<'a> BusOpExpand<'a> {
...
+            ConstraintDomain::EveryFrame(2),
+            tag,
+        );
+
+        // Also store the expression to computed p_prime for auxiliary trace generation
+        ir.buses_transitions.insert(bus_index, (p_prod, p_prime_factor));
+    }
+
+    /// Helper function to expand multiset constraints using the sum form
+    /// (assumes mutually-exclusive latches).
```
huitseeker, line 282: I don’t see a validation or an explicit opt-in of the distinct latches. If latches overlap, the `1 - sum(latch)` term can under‑constrain the bus. Is there a check/diagnostic (or a negative test) we could use to guard this?

mir/src/passes/translate.rs
```diff
@@ -447,9 +577,20 @@ impl<'a> MirBuilder<'a> {
         }
     }
 
-    fn translate_enforce(&mut self, enf: &'a ast::ScalarExpr) -> Result<Link<Op>, CompileError> {
-        let node = self.translate_scalar_expr(enf)?;
-        self.insert_enforce(node)
+    fn translate_enforce(&mut self, enf: &'a ast::Enforce) -> Result<Link<Op>, CompileError> {
+        let tag_spec = enf.tag.clone();
+        let tag = tag_spec.as_ref().and_then(|spec| spec.as_single());
```
huitseeker: Here we reduce tag specs to a single tag. Does that mean `@tag_range` / `@tag([..])` won’t propagate into bus boundary extraction via `current_constraint_tag`, and that `p.first` / `p.last` can lose tags?

Can we plumb tag specs through or expand before setting bus boundary tags?

air/src/passes/expand_buses.rs
Al-Kindi-0: The reason I wasn't tight here was because we should be moving to the uniformized way of treating buses we have adopted in `p3-miden` and in `miden-vm`, namely we start with identity and check against `aux_finals`. 
This implies some simplifications, most notably the possibility to deprecate variable length public inputs (input tables) in the language.

I fixed it for now to work as you expected and add a short comment with the above caveat.

air/src/passes/expand_buses.rs
Al-Kindi-0: I have thought about this and it is a hard problem. With the structure we have on the flags, we might be able to come up with acceptable solutions or at least heuristics to check for mutual-exclusivity of latches.
This will have to be done in a followup work as it is not trivial.
For now, I changed the syntax to make explicit the assumption we are making.

mir/src/passes/translate.rs
Al-Kindi-0: Fixed it so that boundary constraints require tags to be single and emit an error otherwise.

adr1anh (approved):
Mainly checked the bus-related code which looks good to me. This tagging infrastructure will make it easier to manually review whether the codegen is working as expected.

Leo-Besancon (approved):
LGTM, thank you for the additional test!

mir/src/passes/translate.rs
```diff
@@ -137,8 +142,133 @@ impl<'a> MirBuilder<'a> {
         Ok(())
     }
 
+    fn validate_constraint_tags(&self) -> Result<Option<u64>, CompileError> {
+        let mut tags = Vec::new();
+        Self::collect_tags_from_statements(&self.program.boundary_constraints, &mut tags)?;
+        Self::collect_tags_from_statements(&self.program.integrity_constraints, &mut tags)?;
```
Soulthym, line 150: I think there is a validation gap for tagged constraints defined only inside evaluator bodies.

Right now `CURRENT_MAX_ID` enforcement in MIR translation appears to collect tags only from root `boundary_constraints` / `integrity_constraints` statements (plus bus transition tags), so evaluator-local `@tag(...)` may not trigger the requirement. Then AIR `TagValidation` is skipped when `expected_max_constraint_id` is `None`, which lets these tags pass without the global check.

I reproduced this locally with two minimal cases:

Case A (Evaluator-only tag, no `CURRENT_MAX_ID` -> Compiles successfully)
```air
def test

ev tagged_ev([a]) {
    @tag(0) enf a' = a + 1;
}

trace_columns {
    main: [a],
}

public_inputs {
    stack_inputs: [16],
}

boundary_constraints {
    enf a.first = 0;
}

integrity_constraints {
    enf tagged_ev([a]);
}
```

Compiler Output (Case A):
```
Success! Transpiled to ...
```

Case B (Evaluator tag + Root tag, no `CURRENT_MAX_ID` -> Fails correctly)
```air
def test

ev tagged_ev([a]) {
    @tag(0) enf a' = a + 1;
}

trace_columns {
    main: [a],
}

public_inputs {
    stack_inputs: [16],
}

boundary_constraints {
    @tag(0) enf a.first = 0;
}

integrity_constraints {
    enf tagged_ev([a]);
}
```

Compiler Output (Case B):
```
error: missing CURRENT_MAX_ID constant for tagged constraints
 = define `const CURRENT_MAX_ID = <max id>;` in the root module

error: compilation failed, see diagnostics for more information
```

A few issues I can think of with this:
- Validation depends on where the tag is written (root section vs evaluator body), not whether tagged constraints exist in final AIR.
- It allows constraints emitted through evaluator expansion to bypass the global `CURRENT_MAX_ID` contract.

Suggestions:
Include evaluator-body tags in the pre-validation tag collection used to gate `CURRENT_MAX_ID`.

Or  require `CURRENT_MAX_ID` whenever any final AIR constraint has a tag (post-expansion).

Add a regression test for the evaluator-only tagged case (expected failure without `CURRENT_MAX_ID`).

I may have missed something, might be worth documenting it for now and doing this in a follow-up, up to you.

air-script/tests/buses/buses_sum_form.air
```diff
@@ -0,0 +1,28 @@
+def BusesSumFormAir
+
+trace_columns {
+    main: [s],
+}
+
+buses {
+    multiset p @sum_form,
```
Soulthym: This still uses the @sum_form syntax, but the parser now requires @sum_form(assume_exclusive).

I verified this .air currently fails to transpile with: `invalid bus form attribute (expected @sum_form(assume_exclusive))`.

This doesnt fail because this .air file is not currently covered by the Winterfell transpile snapshot list, so CI still passes.

Suggestions:
- Update this .air to @sum_form(assume_exclusive).
- Optionally add buses_sum_form.air in air-script/tests/codegen/winterfell.rs tests so syntax regressions on this file are caught automatically.

Soulthym (approved):
Hey @Al-Kindi-0, I left 2 comments below regarding CURRENT_MAX_ID and a .air file that didnt get updated to the @sum_form(assume_exclusive) syntax.

Those arent really blocking so I marked this review as Approved, the rest looks good to me!

mir/src/passes/translate.rs
Al-Kindi-0: Good point, and I have gone ahead and fixed it so that it covers tags inside of evaluators.
Also added a regression test.
Thanks!

air-script/tests/buses/buses_sum_form.air
Al-Kindi-0: Yeah, I applied your suggestion as I have missed updating this test to the new explicit syntax

PR #540, opened 2026-02-16
Tagged system + range checker constraints

Al-Kindi-0:
- Updated `CHANGELOG.md`


constraints/tests/fixtures/ood_expected.rs
```diff
@@ -0,0 +1,130 @@
...
+];
+
+/// Expected OOD values.
+pub fn expected_ood_evals() -> Vec<EvalRecord> {
+    vec![
+        EvalRecord {
+            id: 0,
+            name: "system.clk.first_row",
+            value: QuadFelt::new(Felt::new(1065013626484053923), Felt::new(0)),
+        },
```
Leo-Besancon, line 48: I assume for now you copy-pasted the output of prints in the miden-vm pipeline? If it does not cause circular dependency issues I think it would be nice to create an issue to output them in a file in miden-vm that could be imported here (in the future), to make it easier to check that the evaluations are in sync between the two repos.

constraints/tests/helpers/ood.rs
```diff
@@ -0,0 +1,640 @@
...
+        out[0..8].copy_from_slice(&self.seed.to_le_bytes());
+        out[8..16].copy_from_slice(&counter.to_le_bytes());
+        out[16..24].copy_from_slice(&mix.to_le_bytes());
+        out[24..32].copy_from_slice(&sum.to_le_bytes());
+        out
+    }
+}
+
+/// OOD eval uses Goldilocks quadratic extension (x^2 - 7) to match Plonky3/Miden VM.
+/// This is a test-only workaround until we depend on Plonky3 types.
```
Leo-Besancon, line 242: Aside from this aspect, how realistic would it be to unify these OOD eval helpers with the ones in `air/src/ir/random_inputs.rs`?

Leo-Besancon (approved):
LGTM! Left two questions but not for this PR.

constraints/tests/fixtures/ood_expected.rs
Al-Kindi-0: Agreed, the manual solution was chosen just for expediency, in the future we can come back and improve on things.

constraints/tests/helpers/ood.rs
Al-Kindi-0: Yes, the main issue if I remember correctly, is that the extension field used in Plonky3 is different than the one you referenced and used in Winterfell. This is why there is this work-around. We might think about a solution once we have some spare capacity.

PR #542, opened 2026-02-19
General improvements and fixes 

Al-Kindi-0:
## Describe your changes

 This PR makes the compiler faster (more precisely usable) and more predictable when converting MIR into AIR. It adds a cache so we don’t translate the same MIR structure multiple times, and it cleans up the helper code used during translation. It also adds simple stats so we can see cache usage.

  Why:

Large AIRs often contain repeated MIR patterns, especially after inlining. Without caching, the compiler re‑translates the same structure many times. For example, in the general stack constraints we used a helper that returns a vector of op flags. Each access to that vector triggered another inlined expansion of the same helper, which blew up the IR size and exhausted memory. The cache prevents this repeated work and keeps translation both faster and more stable.

- Updated `CHANGELOG.md`


mir/src/ir/quad_eval.rs
```diff
@@ -65,6 +68,32 @@ pub struct RandomInputs {
...
+        if let Op::Value(value) = node.borrow().deref()
+            && let MirValue::Constant(ConstantValue::Felt(c)) = value.value.value
+        {
+            return Ok(c as usize);
+        }
+        println!("Non-constant index in RandomInputs::eval for {context}");
+        Err(CompileError::Failed)
+    }
+
+    fn eval_trace_access(&mut self, trace_access: TraceAccess) -> Result<QuadFelt, CompileError> {
```
Leo-Besancon, line 81: I think there is still some code duplication with L. 156

mir/src/lib.rs
```diff
@@ -26,11 +26,30 @@ impl Pass for MirPasses<'_> {
...
+        let mir = ast_to_mir.run(input)?;
+
+        let mut projection = passes::IndexProjection::new(self.diagnostics);
+        let mir = projection.run(mir)?;
+
+        let mut inlining = passes::Inlining::new(self.diagnostics);
+        let mir = inlining.run(mir)?;
+
+        // AIR_DISABLE_CSE disables the common subexpression elimination pass.
+        let mir = if std::env::var("AIR_DISABLE_CSE").is_ok() {
```
Leo-Besancon: In the env var name, we should differentiate between the new CSE in MIR passes here and the CSE in AIR passes (I assume the AIR CSE is still relevant).

Leo-Besancon: Also, it seems that the checks are a bit inconsistent, for IndexProjection we always instantiate the pass but it does nothing if the env var is set.

air/src/passes/translate_from_mir.rs
```diff
@@ -701,6 +1065,69 @@ impl AirBuilder<'_> {
...
+                .sum(),
+            Op::Enf(enf) => {
+                let child = accessor_to_scalar(&enf.expr);
+                if child.as_vector().is_some() || child.as_matrix().is_some() {
+                    Self::expanded_boundary_len(&child)
+                } else {
+                    1
+                }
+            },
+            _ => 1,
```
Leo-Besancon, line 1096: Here (and it's the same in `expanded_integrity_len`), we return 1 for all the other Op types. I assume this is fine because at this stage of the compilation the unrolling / constant_prop would have removed the Op types that could pose problem here, but I'd expect some comments / checks on these assumptions here. For instance, what accessors we can have and why we are sure they expand to a scalar.

Leo-Besancon (commented):
The optimizations and fixes in this PR seem really good, in particular the caches and the new MIR node ownership approach!

Going through the diff, it's hard to make sure that no edge cases have been missed, but the code seems clean and the tests would have caught most of the potential regressions. I've left a few comments but nothing major.

mir/src/ir/quad_eval.rs
Al-Kindi-0: Good catch, deduplicated it

mir/src/lib.rs
Al-Kindi-0: Fixed both points, thanks for catching them

air/src/passes/translate_from_mir.rs
Al-Kindi-0: Added clarifying comments about that

Leo-Besancon (approved):
Thanks for the small cleanups, LGTM now!

parser/src/sema/semantic_analysis.rs
```diff
@@ -643,13 +643,78 @@ impl VisitMut<SemanticAnalysisError> for SemanticAnalysis<'_> {
...
+                    .with_secondary_label(prev.span(), "previously bound here")
+                    .emit();
+                return ControlFlow::Break(SemanticAnalysisError::NameConflict(name.span()));
+            }
+            bound.insert(name);
+        }
+
+        let binding_ty = self.expr_binding_type(&expr.value).unwrap();
+        let expected_len = binding_names.len();
+        if expected_len > 1 {
```
huitseeker: The `expected_len > 1` guard means `let [a] = scalar_expr;` passes even when the value isn’t a vector. Can this later create invalid accessors and crash in MIR→AIR? Can we validate vector-ness for any `LetBinding::Vector`, even length 1?

air/src/passes/translate_from_mir.rs
```diff
@@ -218,6 +393,236 @@ fn enf_to_scalar(mir_node: &Link<Op>) -> Link<Op> {
...
+            },
+            MirValue::PeriodicColumn(periodic_column_access) => MirValueKey::PeriodicColumn {
+                name: Self::qualified_identifier_key(&periodic_column_access.name),
+                cycle: periodic_column_access.cycle,
+            },
+            MirValue::PublicInput(public_input_access) => MirValueKey::PublicInput {
+                name: public_input_access.name.name(),
+                index: public_input_access.index,
+            },
+            MirValue::PublicInputTable(public_input_table_access) => {
```
huitseeker, line 511: `MirValueKey::PublicInputTable` ignores `bus_type`. With MIR→AIR caching, could two tables with the same name/cols but different bus types be deduped to the wrong AIR value? Should the key include `bus_type`?

parser/src/transforms/constant_propagation.rs
```diff
@@ -133,14 +133,39 @@ impl<'a> ConstantPropagation<'a> {
...
         if is_constant {
             match expr.value {
-                Expr::Const(ref value) => {
-                    self.local.insert(expr.name, value.clone());
+                Expr::Const(ref value) => match expr.binding {
+                    LetBinding::Single(name) => {
+                        self.local.insert(name, value.clone());
+                    },
+                    LetBinding::Vector(ref names) => {
+                        let ConstantExpr::Vector(ref values) = value.item else {
```
huitseeker: This will panic if the constant isn’t a vector or the vector is shorter than `names`. Since this runs before semantic analysis, inputs like `let [a, b] = [1];` or `let [a, b] = 1;` will crash instead of producing a diagnostic. Should we validate shape/length here (or skip folding and let sema handle it)?

huitseeker (commented):
Very cursory pass from me.

parser/src/sema/semantic_analysis.rs
Al-Kindi-0: Agreed, changed it so that the validation covers the ` = 1` case as well.

air/src/passes/translate_from_mir.rs
Al-Kindi-0: Indeed, that is a real possibility even though as I mentioned we should deprecate `PublicInputTable`.
I have fixed it in the meantime as you suggested

parser/src/transforms/constant_propagation.rs
Al-Kindi-0: Good point, done!

PR #543, opened 2026-02-19
Stack general constraints

Al-Kindi-0:
## Describe your changes

Implements the general constraints for the operand stack.

- Updated `CHANGELOG.md`

### DIFF
--- a/constraints/flags.air
+++ b/constraints/flags.air
@@ -1,25 +1,747 @@
 ##########################################################################################
 # FLAG HELPERS MODULE
 ##########################################################################################
+#
+# This module defines the per-op opcode flag polynomials derived from decoder op bits.
+# These helpers are shared across system, stack, and decoder constraints.
+# Keep these in sync with the Rust `op_flags` module.
+##########################################################################################
 
 mod flags
 
+# SYSTEMATIC OPERATION FLAG FUNCTIONS - INDIVIDUAL FLAGS PER OP
+##########################################################################################
+
+##########################################################################################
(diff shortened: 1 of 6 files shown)


Leo-Besancon:
Haven't checked the flags computations in depth, but I think the logic of the stack constraints based on these and the tagging make sence.

constraints/flags.air
```diff
@@ -1,25 +1,1037 @@
...
+    let mov2_prefix = is_movup2 + is_movdn2;
+    let mov3_prefix = is_movup3 + is_movdn3;
+    let mov4_prefix = is_movup4 + is_movdn4;
+    let mov5_prefix = is_movup5 + is_movdn5;
+    let mov6_prefix = is_movup6 + is_movdn6;
+    let mov7_prefix = is_movup7 + is_movdn7;
+    let mov8_prefix = is_movup8 + is_movdn8;
+
+    # Directional MOV sums by depth.
+    let movdn_ge2 = is_movdn2 + is_movdn3 + is_movdn4 + is_movdn5 + is_movdn6 + is_movdn7 + is_movdn8;
```
adr1anh, line 114: If we wrote these in reverse, we could avoid summing the same terms.

adr1anh (approved):
Looks good! I didn't meticulously go through all the constraints as I think it'll make more sense to focus on one set (either airbuilder or airscript).

constraints/flags.air
Al-Kindi-0: Personally, the current way is more instructional and informative, though it is not efficient.
I propose we keep as is unless there are strong preferences

PR #545, opened 2026-02-24
Stack shuffling ops constraints

Al-Kindi-0:
- Updated `CHANGELOG.md`


constraints/stack_ops.air
```diff
@@ -0,0 +1,257 @@
...
+    let is_swapw = flag_swapw(op_bits);
+    let is_swapw2 = flag_swapw2(op_bits);
+    let is_swapw3 = flag_swapw3(op_bits);
+    let is_swapdw = flag_swapdw(op_bits);
+
+    let is_cswap = flag_cswap(op_bits);
+    let is_cswapw = flag_cswapw(op_bits);
+
+    # Dummy term to keep binary constraints in the transition domain.
+    let zero = s0' - s0';
```
Leo-Besancon: Could you add a bit more details there?
I would expect this to be only used when we know that the last row will break a constraint, which may not be the case here?

Al-Kindi-0: Same point as in the other PR, this constraint doesn't need to be forced to be transition only. Fixed it now

Leo-Besancon (approved):
LGTM

PR #547, opened 2026-02-28
Decoder core constraints

Al-Kindi-0:
- Updated `CHANGELOG.md`


constraints/decoder.air
```diff
@@ -0,0 +1,325 @@
...
+    let is_join = flag_join(op_bits, e0);
+    let is_split = flag_split(op_bits, e0);
+    let is_loop = flag_loop(op_bits, e0);
+    let is_dyn = flag_dyn(op_bits, e0);
+    let is_dyncall = flag_dyncall(op_bits, e0);
+    let is_call = flag_call(op_bits, e1);
+    let is_syscall = flag_syscall(op_bits, e1);
+    let is_end = flag_end(op_bits, e1);
+    let is_repeat = flag_repeat(op_bits, e1);
+    let is_halt = flag_halt(op_bits, e1);
```
Soulthym, line 54: Shouldn't these be mutually exclusive? I didn't find this being enforced

Soulthym (commented):
I'm not super familiar with the decoder constraints yet so I may have missed something. Left a nit below, not sure if relevant. The rest looks good to me.

constraints/decoder.air
```diff
@@ -0,0 +1,325 @@
...
+    # If batch size is <= 4 groups, unused hasher lanes h4..h7 must be zero.
+    let small_batch = batch_g1 + batch_g2 + batch_g4;
+    let small_batch_tail = [h4, h5, h6, h7];
+    enf small_batch_tail[i] = 0 for i in 0..4 when small_batch;
+
+    # If batch size is <= 2 groups, unused hasher lanes h2..h3 must be zero.
+    let tiny_batch = batch_g1 + batch_g2;
+    let tiny_batch_tail = [h2, h3];
+    enf tiny_batch_tail[i] = 0 for i in 0..2 when tiny_batch;
+    enf h1 = 0 when batch_g1;
```
Leo-Besancon, line 202: Add comment for consistency (If batch size is 1 group, unused hasher lanes h1 must be zero.)

constraints/decoder.air
```diff
@@ -0,0 +1,325 @@
...
+    # Tracks block nesting. Inserts push (block_id, parent_id, is_loop, ctx, depth, overflow, fn_hash)
+    # on block entry; removals pop the same tuple on END/RESPAN.
+    # For JOIN/SPLIT/SPAN/DYN/RESPAN, the context fields (ctx, depth, overflow, fn_hash) are 0
+    # and is_loop = 0 (except LOOP which uses s0). CALL/SYSCALL/DYNCALL fill:
+    #   ctx = system_ctx, depth = stack_b0 (or h4 for DYNCALL), overflow = stack_b1 (or h5),
+    #   fn_hash = fn_hash_0..fn_hash_3.
+    bus_0_decoder_p1.insert(addr', addr, 0, 0, 0, 0, 0, 0, 0, 0) when fjoin;
+    bus_0_decoder_p1.insert(addr', addr, 0, 0, 0, 0, 0, 0, 0, 0) when fsplit;
+    bus_0_decoder_p1.insert(addr', addr, 0, 0, 0, 0, 0, 0, 0, 0) when fspan;
+    bus_0_decoder_p1.insert(addr', addr, 0, 0, 0, 0, 0, 0, 0, 0) when fdyn;
```
Leo-Besancon, line 354: Q: With `@sum_form(assume_exclusive)` being defined, does this lead to the same bus constraint as if we do 
`bus_0_decoder_p1.insert(addr', addr, 0, 0, 0, 0, 0, 0, 0, 0) when fjoin + fsplit + fspan + fdyn`?
If not, would this be more optimized?

The same comment applies below when we insert the same expression on different flags.

constraints/decoder.air
```diff
@@ -0,0 +1,325 @@
...
+    bus_0_decoder_p1.remove(addr, addr', is_loop_flag, 0, 0, 0, 0, 0, 0, 0) when is_end & is_simple_end;
+
+    # ---------------------------------------------------------------------------
+    # BUS 1: block hash table (p2)
+    # ---------------------------------------------------------------------------
+    # Tracks blocks awaiting execution. Inserts child hashes on JOIN/SPLIT/LOOP/REPEAT/etc.,
+    # and removes the parent hash on END. The `is_first_child`/`is_loop_body` flags help
+    # disambiguate left/right children and loop bodies in the multiset.
+    bus_1_decoder_p2.insert(addr', h0, h1, h2, h3, 1, 0) when fjoin;
+    bus_1_decoder_p2.insert(addr', h4, h5, h6, h7, 0, 0) when fjoin;
```
Leo-Besancon, line 380: Q: Does this work with the `sum_form(assume_exclusive)` flag? Is it ok because the latch is exactly the same?

Leo-Besancon (commented):
Left a nit and two questions on bus constraints, otherwise looks good to me!

constraints/decoder.air
Leo-Besancon: These should come from the op_bits bit decomposition, so I believe they are mutually exclusive by design without needing to enforce it through constraints.

Soulthym: Ah, that makes sense then.

constraints/decoder.air
Al-Kindi-0: The two forms are mathematically equivalent. With separate when clauses, the compiler creates four entries in the `insert_groups` map keyed by `latch`, then sums them as `x * fjoin + x * fsplit + x *  fspan + x * fdyn + (1 - (fjoin + fsplit + fspan + fdyn))`.
With a combined flag it would create one entry  `x * (fjoin + fsplit + fspan + fdyn) + (1 - (fjoin + fsplit + fspan + fdyn))`, which is basically the same polynomial.

The combined form would produce a slightly more compact IR (fewer intermediate nodes), but the final result is identical. I kept them separate for readability and to match the `miden-vm` Rust side which also lists them individually.

constraints/decoder.air
Al-Kindi-0: Indeed, this works correctly with `assume_exclusive`. When two inserts share the same latch, the compiler groups them and multiplies their encoded messages together as `(msg1 * msg2) * fjoin`, which is the encoding we want when tuples are inserted on the same row.

The `assume_exclusive` assumption is only about different flags being mutually exclusive (i.e., at most one active per row). Multiple inserts on the same flag are handled by the grouping logic described above and implemented in `expand_multiset_constraint_sum_form`, meaning that they get folded into a single product term for that flag's group.

Leo-Besancon: Perfect, that's what I thought but wanted to be sure, thank you!

